PR #371, opened 2016-08-12
Fix typos

Djiko:
A quick fix on typos I noticed in the fr language file. Thanks for making it possible :) !


roundhill:
We actually use GlotPress for the translations, so you can fix it here: https://translate.wordpress.com/projects/simplenote/android/fr/default/

I will update the README with translation contribution info. Thanks for contributing!

PR #372, opened 2016-08-12
Update welcome note text, as there is not actually a pencil icon

irontoby:
Wasn't sure how to best describe the button in text, but it's certainly not a pencil :)


roundhill:
Oh yeah, it _used_ to be a pencil! The description works for me. :shipit: Thanks for contributing.

Issue #360, opened 2016-05-24
Date translation "quoted" on the Note info bottom sheet


maxme:
`24 mai 'à' 10:55` should be `24 mai à 10:55`

![screenshot_20160524-105744](https://cloud.githubusercontent.com/assets/40213/15498103/5b651352-219e-11e6-946f-e4111c390f93.png)

theck13:
Closed by https://github.com/Automattic/simplenote-android/pull/932.  See the screenshot below for illustration.

![360](https://user-images.githubusercontent.com/3827611/75552354-5a972000-59f3-11ea-8f7c-ccddcd0952c0.png)

PR #375, opened 2016-08-21
  Improving French translation

marwendoukh:
Improving French translation


maxme:
`Modifié le %s` would be better, and this doesn't fix #360 (which is related to date formatting)

Issue #376, opened 2016-08-21
Add possibility to set reminders for notes


smonsays:
It would be a nice feature to be able to set time controled reminders to notes, which then pop up in the notification area at the defined time.

philgood:
I would like to see this feature incorporated into the app.

marwendoukh:
Hi @roundhill , I'm currently developing this new feature and I have some questions . How can I contact the Simplenote team ?

roundhill:
@marwendoukh Since we've gone open source, perhaps we should create a Slack group to use for Simplenote dev? You can ask questions here, too.

marwendoukh:
Well , how can I get the "id" of a given Note ? because every Reminder should have its "NoteID"

roundhill:
> Well , how can I get the "id" of a given Note ? because every Reminder should have its "NoteID"

Every note is a Simperium object with its own key, so use `note.getSimperiumKey()`.

Jonatino:
Yes please!! This is a feature I think a lot of people want. In Google Keep I have a note to send a reminder every day at 6:45PM for stuff I think of during the work day but will forget by the time I get home.

BrixV2:
Please!

unai-ndz:
Please!

Wizardling:
I would also love this feature to be added, along with the ability to set a reoccurring notification. Then Simplenote could totally replace Google Keep on my Android devices.

theck13:
Thanks for the suggestion!  I have added it to our feature requests.

PR #380, opened 2016-08-27
Add reminder feature

marwendoukh:
Adding reminder feature ( issue #376 )


roundhill:
Thanks a lot for the contribution! I think you've got a great idea going here, and I think we can tweak some things to get it feeling at home as a Simplenote feature:

*  Instead of adding a button to set a reminder, what if when a date is typed into a note and it is picked up by `Linkify`, we prompt the user if they want to set a reminder for that date? This way we get UX out of the way. If that's not possible, I would recommend we move the button into the note info panel so that we aren't cluttering up the toolbar.
*  It'd be cool if we saved the reminder date as a UTC timestamp to the `note` object and then called `save()` so that it syncs. That way other Simplenote clients could show the reminder as well.
*  The reminder ui seems really in your face, wouldn't a standard Android notification fit in a bit better? And you could have the notification open the note when tapped.

One more thing, if there's any way to split up the PR in smaller chunks, that would help the code review.

marwendoukh:
@roundhill  Is it better now ?

PR #381, opened 2016-08-27
Add note reminder

jegumi:
Based on the issue https://github.com/Automattic/simplenote-android/issues/376, I've implemented a similar feature.

Clicking on a note, you can add a reminder. Currently is just a one time reminder, that can be snooze, but it'd be quite easy to add a repetition feature.

At the date indicated on the reminder, the app will show a notification. There are two actions on the notification:

- Snooze. When clicking on this action, this reminder will be delayed 10 minutes. The notification is cleared, but the reminder will appear again in 10 minutes. You can check the state of the snooze reminder on the reminder sheet of the note

- Stop. It disables the reminder. Right now it's not a big deal, as the reminder is just one shot, but if repetition is added it'd be interesting.

After creating this PR, I saw that this feature was implemented some hours ago by other contributor. Sorry, I didn't notice it before, and anyway I enjoyed working on this feature :D


roundhill:
Hmm, maybe you and @marwendoukh could coordinate efforts?

PR #383, opened 2016-08-29
Editor: Add Back `lineSpacingExtra` for better line spacing

roundhill:
Looks like Goole finally fixed the `lineSpacingExtra` bug in 7.0 that they added in 5.0 :)

Before:
screen shot 2016-08-29 at 4 12 43 pm

After:
screen shot 2016-08-29 at 4 12 21 pm

This makes the text in the editor more readable.

cc @0nko


0nko:
Sweet! :)

Issue #255, opened 2014-12-23
Clean up compiler warnings


roundhill:
Inspecting the code in Android Studio reveals quite a few little warnings that can be easily fixed.

PR #395, opened 2016-10-05
Cleanup

Tunous:
Just an automatic Android Studio cleanup and few manual fixes :dash:
Might be related to #255


roundhill:
Looks like there's some conflicts now, happy to have a look once resolved.

Tunous:
Fixed

Simplenote/src/main/java/com/automattic/simplenote/utils/HtmlCompat.java
```diff
@@ -0,0 +1,16 @@
+package com.automattic.simplenote.utils;
+
+import android.os.Build;
+import android.text.Html;
+import android.text.Spanned;
+
+public class HtmlCompat {
+    @SuppressWarnings("deprecation")
+    public static Spanned fromHtml(String source) {
```
roundhill: What's this for? I'm not familiar with the recent changes to `Html`

Tunous: This change isn't really required. I just added it to avoid using now deprecated method and eliminate all warnings. 

In N they added this new fromHtml method which takes flags parameter that allows for changing how text rendering works. 

I linked few references in description of this commit https://github.com/Tunous/simplenote-android/commit/c7ae2fac26d3a35a54d8ca38c271c6a39378eea9. 

I can revert if that's not wanted.

roundhill:
:shipit:

PR #402, opened 2016-11-28
Fill Viewport with Input Text Box in Edit Tab

theck13:
### Fix
The input field in the Edit tab only fills area with text.  If no text is added, you must tap the top of the input field just below the Edit tab in order to select the field and input text.  The green area show in the screenshot below represents the ***old*** selectable/tappable area.

simplenote_layout_bounds

This fix expands the input field to fill the viewport (i.e. screen) allowing you to tap anywhere in the input field to select it.  The green area show in the screenshot below represents the ***new*** selectable/tappable area.

simplenote_layout_bounds_new

### Test
1. Create new note.
2. Select ***Tags*** field.
3. Tap anywhere in ***Edit*** tab field.
4. Notice input field is selected.

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -37,7 +37,7 @@ dependencies {
     wearApp project(':Wear')
 }
 
-version "1.5.1"
+version "1.5.2"
 
 android {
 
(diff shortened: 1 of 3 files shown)


roundhill:
Works as described :shipit: 

Thanks for the quick fix @theck13

PR #403, opened 2016-11-29
Travis config + Fix release build by using an old version of the gradle android plugin

maxme:
New Travis config: build debug and release versions, also run lint checks.

There is a bug with NDK processing in 2.1+ https://code.google.com/p/android/issues/detail?id=223162 that prevents any release build, I had to downgrade to 2.0.0. Hopefully this will be fixed in 2.2.3+ or 2.3+.

### DIFF
--- a/Simplenote/gradle.properties-example
+++ b/Simplenote/gradle.properties-example
@@ -1,5 +1,3 @@
 simperiumAppId=FIXME
 simperiumAppKey=FIXME
-hockeyAppId=FIXME
-googleAnalyticsId=FIXME
-facebookAppId=FIXME
+googleAnalyticsId=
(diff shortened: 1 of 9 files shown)


Simplenote/gradle.properties-example
```diff
@@ -0,0 +1,5 @@
+simperiumAppId=FIXME
+simperiumAppKey=FIXME
+hockeyAppId=FIXME
+googleAnalyticsId=FIXME
+facebookAppId=FIXME
```
roundhill: We should update the `README` to rename this file maybe? Doesn't need to be in this PR

roundhill: Also, we can kill `facebookAppId` and `hockeyAppId`

maxme:
@roundhill I updated the gradle.properties and README files.

roundhill:
:shipit:!

PR #404, opened 2016-11-29
Markdown Preview Fixes

roundhill:
Fixes the markdown preview from not loading utf-8 characters properly and from not updating after a change was made in the editor. (https://github.com/Automattic/simplenote-android/issues/405)

cc @theck13


theck13:
:shipit:

PR #408, opened 2016-12-01
Update Library to Passcode Lock v1.5.1

theck13:
### Fix
Update Passcode Lock library to [v1.5.1](https://github.com/wordpress-mobile/PasscodeLock-Android/releases/tag/1.5.1) as described in https://github.com/Automattic/simplenote-android/issues/407.


roundhill:
Confirmed that the fix works and lock screen working as expected in other scenarios. :shipit:

PR #415, opened 2016-12-22
Fix shadow on editor toolbar.

roundhill:
Somewhere along the line the shadow for the editor toolbar stopped displaying. It appears to be related to the `app:layout_scrollFlags` property that we apply to the toolbar.

I think the issue is having too many options in the field there. If I limit it to two it works fine and the shadow appears. @0nko you added the `snap|exitUntilCollapsed` options to it back in your redesign PR. Do you remember why?

This PR reverts it to how @theck13 had it for his markdown PR.

Before:
screen shot 2016-12-22 at 3 30 54 pm

After:
screen shot 2016-12-22 at 3 31 07 pm

_Note: This was not an issue on pre-lollipop devices._


0nko:
I'm not sure exactly but I know there was a problem with the missing shadow on some older version of Android and I remember fixing it. So perhaps this was it..

PR #416, opened 2016-12-22
Added `windowLightStatusBar` settings for light and dark theme.

### LINKED ISSUE #413: Dark Status Bar Icons
Currently the color for the status bar doesn't provide good contrast for the status bar icons.
This can be fixed on Android 6.0 and higher by switching to the dark status bar icons.
This [link](http://stackoverflow.com/questions/33316668/how-to-change-the-status-bar-notification-icons-color-tint-in-android-marshmal) provides guidance on how to go about doing this.

roundhill:
Fixes #413. You should see dark icons when using the light theme, and light icons when using the dark theme if on Android 6.0 or higher:

screen shot 2016-12-22 at 3 40 51 pm

Thanks for the idea @Lokesh-Krishna!

### DIFF
--- a/Simplenote/src/main/res/values-v23/styles.xml
+++ b/Simplenote/src/main/res/values-v23/styles.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="Theme.Simplestyle" parent="Base.Theme.Simplestyle">
+        <item name="android:windowLightStatusBar">true</item>
+    </style>
+
+    <style name="Theme.Simplestyle.Dark" parent="Base.Theme.Simplestyle.Dark">
+        <item name="android:windowLightStatusBar">false</item>
+    </style>
+</resources>


roundhill:
@drw158 Any thoughts on how this looks with the light theme? Maybe we should lighten up the grey color even more? Or were the white icons ok in the first place?

Lokesh-Krishna:
If I may, the white icons definitely suffer from really low contrast.
These ones look good but yes, lightening the status bar color further would make them look even better. I really hope this is a change that makes it.

theck13:
Just for reference, here are some other popular apps with a light theme.

Lokesh-Krishna:
It is Google Calendar that I learned this being a possibility from. 
@theck13 Are you suggesting this not be done then?

theck13:
I do like the Calendar status bar since the status icons seem to match the action icons, but part of me likes the low contrast better for Simplenote.  It helps keep focus on the notes and removes distraction in a way.  Some reading apps remove the network icons and dim the battery icon in the status bar to put the user in a quasi-immersive mode.  I think Simplenote coincidentally has a similar effect with the low contrast status bar icons.  I looking forward to hear what @drw158 thinks.

Lokesh-Krishna:
>  Some reading apps remove the network icons and dim the battery icon in the status bar to put the user in a quasi-immersive mode.

With apps that do that, I think that they usually make it possible to restore them to their normal appearance by tapping on the status bar or some other action, thereby letting users check on battery or time. 

I think that's what the status bar is for, to keep that particular information always easily accessible. The status bar icons were against a fixed dark background until KitKat came along and made transparent status bars possible and then Lollipop made it possible to colour the status bar with the colour of your app and suddenly you couldn't guarantee the visibility of the status bar icons. 

I think that the hit that the visibility took was finally fixed in Marshmallow by making dark status bar icons possible, so that apps that had trouble displaying them properly while staying true to their colours were no longer limited to the white icons. 

I do get what you're saying and like the "not a bug but a feature" angle but I feel that just because Google itself hasn't gotten around to fixing it in their apps doesn't mean we shouldn't either. Google hasn't really been all that respectful of their own design guidelines lately.

drw158:
Because I view Simplenote as utilitarian, I think the status icons should be readable. I'm thinking something like this:

screen shot 2017-01-02 at 2 07 50 pm

Maybe a little darker. Maybe just follow what Calendar does, but lighten the icons a bit.

screen shot 2017-01-02 at 2 22 50 pm


I don't have an Android device on me right now so I can't test out the dark and light themes. The dark theme should be the same as the light theme, but reversed, to be consistent.

I do love @theck13's idea of dimming the icons, but I would only do that in the editor if possible. That'd be really slick.

roundhill:
Here it is with a lighter color in the system bar background as well as a completely white option:

screen shot 2017-01-03 at 3 10 30 pm

screen shot 2017-01-03 at 3 11 13 pm

Lokesh-Krishna:
> The dark theme should be the same as the light theme, but reversed, to be consistent.

The second image, with the completely white status bar does this best, seeing how the dark theme has a completely black status bar.

roundhill:
Looks like we didn't get to a consensus on this, going to close for now.

PR #418, opened 2017-01-09
Add crashlytics to track eventual crashes in the app

maxme:
Add crashlytics to track eventual crashes in the app

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ gradle.properties
 
 # When developing against Simperium-android source directly
 Simperium
+
+# Crashlytics API Key
+Simplenote/crashlytics.properties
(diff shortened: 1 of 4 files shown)


roundhill:
I totally missed this. We do get crash reports via Google Analytics, is that good enough or do we need the detail that crashlytics provides?

maxme:
> We do get crash reports via Google Analytics, is that good enough or do we need the detail that crashlytics provides?

That's good enough if someone is looking at them :) - IMO it's better to track crashes from all our apps in the same place (Crashlytics).

roundhill:
> IMO it's better to track crashes from all our apps in the same place (Crashlytics).

Good point, I've tested this and confirmed that it is working fine! :shipit:

Issue #394, opened 2016-10-05
Disappearing BottomSheetDialog


Tunous:
When closing any of the note bottom sheet dialogs via swipe down gesture they no longer appear but only show dimmed screen.

This appears to be support library bug and I didn't find a solution for that other than recreating bottom sheets every time.

Relevant android bugs issue: https://code.google.com/p/android/issues/detail?id=220706

theck13:
I cannot reproduce that behavior.  Feel free to reopen this issue if you can.

PR #420, opened 2017-01-22
Fix: bottomsheet not showing up

maheshgaya:
Similar issues: #394
- **Issue:** If a user drags a bottom sheet down to dismiss it, the bottom sheet
does not show up when the user taps on the share button again
- **Fix:** remove check for null state when creating the bottom sheet.
Instead moved that null check to when the bottom sheet is being dismissed. Now, the bottom sheet will reappear.

**Issue in screenshots:**
Step 1: User taps on Share button for the first time

image

Step 2: User drags the bottomsheet down, which should dismiss it [in theory]

image

Step 3: User taps the share button again. The bottom sheet does not show up!

image


maheshgaya:
Similar issues: #394

theck13:
Thanks for the fix, @maheshgaya!

PR #432, opened 2017-03-03
Enforce Sign in but allow previous offline users to still use the app

roundhill:
Unfortunately we weren't able to figure out why some users would lose their notes after signing in to the app while using it offline for a while. This PR removes the capability for offline use for new users, but allows users to continue using it offline when they upgrade.

This also removes the 'WelcomeView' that we showed in the sidebar to entice offline users to sign in. The only place you can sign in/out now is from the Settings view.

There's a few test scenarios:
**Fresh Install**
* Delete the app and install it fresh from this branch. You should immediately see the login screen after launching the app, and pressing the back button should quit the app.

**Upgrade with Offline Use**
* Reinstall the app fresh from the develop branch and don't sign in. Create a few notes.
* Switch to this branch and run it.  You should be able to continue using the app offline normally.
* Sign in from the settings menu. The offline notes should sync up to the account.

**Upgrade with Online Use**
* Reinstall the app from the develop branch, and sign in to the app.
* Update to this branch. You should still be signed in and able to use the app normally.


Simplenote/src/main/java/com/automattic/simplenote/utils/PrefUtils.java
```diff
@@ -32,6 +32,9 @@
     // boolean, set on first launch
     public static final String PREF_FIRST_LAUNCH = "pref_key_first_launch";
 
+    // boolean, set on first launch
```
theck13: Is this a copy/paste artifact?  Maybe it should be something like `// boolean, require account login` instead.

roundhill: Yeah, I probably should have made it more specific. Fixed in 1ac193a

PR #433, opened 2017-03-03
Swap the position of the pin and trash note icons in multi select.

roundhill:
I think we discussed this somewhere before, but it makes more sense to have the destructive action at the right side. This PR switches the positioning of the pin and trash icons so the trash is on the right:

screen shot 2017-03-03 at 2 33 00 pm


mzorz:
:shipit:

PR #441, opened 2017-04-01
feature: encryption

dogear42:
Hello!

I have added an encryption feature to simplenote-android (and also to simperium-android, which I will submit a separate pull request for), and would like to discuss whether it is something that would be desirable in the base product.

The encryption happens only on the wire to/from the server.   When encrypted, the data is kept in ASCII-ARMOR mode, so that it is text-diffable.  Consequently, the client/server communications still work as always.  The server just deals with the data as though it were not encrypted, as though it were simple text, even though the client secretly encrypts it.  There is a performance hit, I am sure, as when changing a single byte (like the first character), all subsequent bytes will likely change.  So, most of the time, full-note transmission will probably occur on changes.  A cost for having encryption, I suppose.

Since the ghost is unencrypted, searching still works.

I have had this working on my local for a while, but have not dealt with adding it to other implementations yet (e.g. electron/ios/mac/etc.).  I'm not sure how much work that would be.

Note that the current implementation uses spongycastle passphrase symmetric encryption, with the passphrase stored in preferences.

I look forward to your comments and ideas!

Thanks!

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -44,6 +44,9 @@ dependencies {
     }
 
     wearApp project(':Wear')
+    compile 'com.scottyab:secure-preferences-lib:0.1.4'
+    compile 'com.madgag.spongycastle:pg:1.54.0.0'
+    compile 'com.madgag.spongycastle:prov:1.54.0.0'
 }
 
 version "1.5.5"
--- a/Simplenote/src/main/java/com/automattic/simplenote/BouncyCastleOpenPgpCryptographyAgent.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/BouncyCastleOpenPgpCryptographyAgent.java
@@ -1,0 +1,81 @@
+package com.automattic.simplenote;
+
+import com.automattic.simplenote.utils.PassphraseUtils;
+import com.simperium.client.CryptographyAgent;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.json.JSONTokener;
+import org.spongycastle.bcpg.SymmetricKeyAlgorithmTags;
+import org.spongycastle.jce.provider.BouncyCastleProvider;
+import org.spongycastle.openpgp.PGPException;
+import org.spongycastle.openpgp.examples.ByteArrayHandler;
+
+import java.io.IOException;
+import java.security.NoSuchProviderException;
+import java.security.Security;
+
+// ATTRIBUTION:
+// parts of this file were extracted from
+// https://github.com/rtyley/spongycastle/blob/spongy-master/pg/src/main/java/org/spongycastle/openpgp/examples/ByteArrayHandler.java
+// which uses this license from the original BouncyCastle:
+//
+// Except where otherwise stated, this software is distributed under a license
+// based on the MIT X Consortium license. To view the license, see here:
+// http://www.bouncycastle.org/licence.html
+//
+// Resources:
+// https://www.codeproject.com/Articles/549119/Encryption-Wrapper-for-Android-SharedPreferences
+// https://github.com/scottyab/secure-preferences
+// https://github.com/tozny/java-aes-crypto
+
+public class BouncyCastleOpenPgpCryptographyAgent extends CryptographyAgent {
+
+    private static final String CIPHERTEXT_PROPERTY_NAME = "cipherText";
+
+    public BouncyCastleOpenPgpCryptographyAgent() {
+        Security.addProvider(new BouncyCastleProvider());
+    }
+
+    // Encrypts the rawObject into a string as property "cipherText" of a JSONObject.
+    @Override
+    public final JSONObject encryptJson(JSONObject rawObject) {
+        String rawText = rawObject.toString();
+        if (rawText == null || rawText.length() < 1) {
+            return null;
+        }
+        try {
+            JSONObject encryptedObject = new JSONObject();
+            String cipherText;
+            byte[] encryptedBytes = ByteArrayHandler.encrypt(
+                    rawText.getBytes("UTF-8"),
+                    PassphraseUtils.getPassphrase(),
+                    "bcopca",
+                    SymmetricKeyAlgorithmTags.AES_256,
+                    true
+            );
+            cipherText = new String(encryptedBytes);
+            encryptedObject.put(CIPHERTEXT_PROPERTY_NAME, cipherText);
+            return encryptedObject;
+        } catch (IOException | NoSuchProviderException | PGPException | JSONException e) {
+            throw new RuntimeException("Exception encrypting text", e); // todo how to handle this?  We really want to stop syncing at this point and throw up an error screen. No possible recovery.
+        }
+    }
+
+    // Decrypts the "cipherText" property of the JSONObject into a new JSONObject.
+    @Override
+    public final JSONObject decryptJson(JSONObject encryptedObject) {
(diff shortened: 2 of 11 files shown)


dogear42:
PS.  This will not compile until the encryption feature for simperium has been merged.

dmsnell:
Thanks for the PR @dogear42!

Encryption is an important feature we'd love to have in Simplenote.

Here is an important question for the work, which I think is a primary driver for delay in encrypting notes: what happens when two devices make changes to the same note? Simperium depends on the ability to rebase changes to a document and send those out to the client devices.

Let's say we store the encrypted text "The fox jumps over the dog" and in client A change it to "The _quick_ fox jumps over the dog" while in client B we change it to "The fox jumps over the _lazy_ dog". Both of the clients should end up with "The quick fox jumps over the lazy dog" but I'm afraid that this method of encrypting will prevent that since the server cannot perform that plaintext diff.

Further, how would passphrase changes work? What happens if I change the key on client A? Will the old passphrase still unlock prior versions of the note? Will my notes on client B be indecipherable until I enter the new passphrase? Or what if client B changes the passphrase without first synchronizing with the change from client A?

Something you are probably aware of with this design is also the fact that were the passphrase to be forgotten there would be nothing we could do to help with that. The content would be gone forever.

Please note the security limitations here:
 - The relative size and frequency of changes to any given note would still be decipherable
 - Note history would need to be solved to prevent decoding/reading older versions of a note
 - Since every change would create an entirely duplicate copy of the data on the server, this could increase the data available for statistical attacks. That is to say, if I encrypt a note of a single letter and change that letter to each of A-Z then there are now twenty-seven copies of that data encrypted on the server, all with the same key.

Personally I welcome thinking on how to pull in encryption to Simplenote and Simperium. It's a difficult challenge to work through though, and at some point we may discover that it's simply not an adequate fit with the Simperium model.

dogear42:
The co-PR for simperium is here: https://github.com/Simperium/simperium-android/pull/195

dogear42:
@dmsnell 

Sorry in advance for the "novel".

Yeah, interesting points.

Let me start by saying that I, too, am nervous about how this would work.  I already have a mechanism in place where I can apply crypto to _parts_ of notes using GPG/OpenKeychain, but it is manual and tedious.  If a client knew how to encrypt "parts" of notes, that would solve most of the issue for me.  However, it just seems better to have privacy solutions that are holistic and built into the architecture/framework of the product.  I mean, privacy and security by design has got to be better than bandaids, right?  That's why I tried to see if I could do this in the existing codebase/framework.

"What happens when two devices make changes to the same note? Simperium depends on the ability to rebase changes to a document and send those out to the client devices."
Well, I tested this (or tried), and it seemed to work.  I used a breakpoint in one of the clients to prevent it from completing the transmission its change until the other one had completed.  I don't recall the details of the transmissions, though.  Could the server have detected that too much of the (encrypted) content changed when the second transmitter sent its change, and so it just sent the whole note back to the second transmitter instead of a delta, and the second one then re-diffed and resolved the conflict?  (The client does its diffs on raw data without encryption, BTW.)  Perhaps if you can describe the complete process that you are concerned about, I can understand what it would look like, and I can test it again and ensure I have tested the exact scenario.

"I'm afraid that this method of encrypting will prevent that since the server cannot perform that plaintext diff."  
Probably so.  Interestingly, I have issues all the time between the Simplenote Android client (from the google play store) and the web client where the server does not diff correctly.  I have to fix it myself.  So, the current algorithm certainly isn't perfect, either.  But, I can play with this more to try to see if I can break this encrypted version, I suppose.  So far, it hasn't behaved for me any worse than the google play store version.

"How would passphrase changes work?"  
Well, that's a problem.  You can't change the passphrase on one client and not the other.  They have to be in sync.  Today, if you completely sync withOUT a passphrase, so that all notes are in the client, then you set the passphrase, it will encrypt the data when you save it next (after a change).  I haven't coded anything yet that would walk the notes and force a save to the server, though.  It would probably be a good idea.  

And, although I haven't tried it, I think if you already are synced WITH a passphrase, and you change the passphrase, we could have the client walk all notes and force-save them under this scenario, since they will have changed due to the encryption.  [Also, while I can understand reasons for needing to do this (device security known to have been compromised), I have to say that, in general, I'm not a fan of regular passphrase changes.  If you have a sufficiently difficult passphrase to break, changing it doesn't really accomplish much.  (I read some articles on this, but can't recall where, so I have no backing to provide for this argument right now.)]

But, yeah, the real problem comes when you have to change the passphrase on more than 1 client. 
 Not sure if it would even work.  My suggestion would be to uninstall all clients except 1, change the passphrase on it, then reinstall all of the clients, setting the passphrase on them when you do.  I agree that there are all sorts of potential problems with that, though!  And, I haven't coded the client to prevent syncing until a passphrase is set.  So, if you login, it will start syncing, even if the passphrase isn't set.   I wasn't really expecting this PR to be merged yet.  I was just hoping we could get the code under development.

"Something you are probably aware of with this design is also the fact that were the passphrase to be forgotten there would be nothing we could do to help with that. The content would be gone forever."   Yeah, but that is true with any well-designed security system.  I have that problem already with encrypted email (protonmail) and encrypted cloud storage (via rclone).  If I lose my keys/passphrases, well ... I'm done.  But, just because encryption is available doesn't mean everyone has to use it.  And, security isn't free of all costs.  So, I am careful not to lose important things like keys.

"The relative size and frequency of changes to any given note would still be decipherable".  Sure, but I'm not sure what this tells anyone.  At any rate, it is much less of an issue than having the plain text of my notes on a server somewhere.  To me there is a big difference between a) preventing the success of a targeted attack on an individual and b) preventing someone who cracks a server from getting gigabytes of plain-text data from millions of users (each already associated with an identity item like an email address).  If I'm seeking data illegally, I'm not going to waste energy trying to crack a passphrase for some users if there are still thousands more that have potentially valuable plain-text data just sitting there for the taking.

"Note history would need to be solved to prevent decoding/reading older versions of a note".  True.  For myself, I was planning on using another one of my email accounts to create a new simplenote account, and after setting the passphrase, import my old notes.

"Since every change would create an entirely duplicate copy of the data on the server, this could increase the data available for statistical attacks. That is to say, if I encrypt a note of a single letter and change that letter to each of A-Z then there are now twenty-seven copies of that data encrypted on the server, all with the same key."  Sure.  But, by the same token, all 27 copies are actually different, so I'm not sure if you could attack based on that.  I need to double-check the algorithm, but I'm pretty sure it uses padding/salts, so the kind of attacks you are thinking about might not work very well.  To be honest, this is beyond my educational level, though.  I'm not a security expert.  However, again, anything seems better than plain text.

Some other things we have not discussed yet, but which I thought about while researching this:

1. What about having some notes encrypted and some not?  We could set a flag per-note to say which should be encrypted.

dmsnell:
that's quite a novel indeed @dogear42! :wink:

the recommendation here might simply to be to release your code as a fork of Simplenote. we definitely need to have any given solution available in all the apps if it's going to involve encryption. we also want to make sure that any encryption system we build is very rigorous, not giving just the appearance of security.

some of the changes here are workable but don't necessarily fit within the goals of Simplenote, such as recreating an account and cloning all of the notes when turning on encryption, or when dealing with merge conflicts from two different devices when password changes haven't propagated.

finally, this encryption model just isn't well-suited for Simperium. it ultimately places the burden of diff resolution back on the clients where plaintext inspection is possible and it's going to be hard to maintain that in a distributed environment where some devices won't even have network access.

thanks for the PR for sure! we have this goal in mind (encrypting notes) but I think we're simply a long way away from there at this point.

PR #452, opened 2017-06-21
Update build config to use dev app and key

roundhill:
This PR adds a hard coded dev app key to the app so that developers will not need to create a Simperium account and app in order to build and test the app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ sdk.dir=/Applications/Android Studio.app/sdk
 ./gradlew installDebug
 ```
 
-Sign up for a new account within the app. Use the account for **testing purposes only** as all note data will be periodically cleared out on the server.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -20,18 +16,13 @@ Sample `local.properties`
...
-simperiumAppKey=SIMPERIUM_KEY
-googleAnalyticsId=
-```
-
 * Install debug build with Android Studio or:
 ```shell
 ./gradlew installDebug
 ```
 
+Sign up for a new account within the app. Use the account for **testing purposes only** as all note data will be periodically cleared out on the server.
```
nbradbury: Should there be a note here **not** to sign in using an existing account?

roundhill: Thanks for the copy idea in Slack, updated in 93d9e43

nbradbury:
Looks good! :shipit:

roundhill:
Thanks for the review @nbradbury!

Issue #444, opened 2017-05-09
Crash loading markdown in NoteEditorFragment


roundhill:
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.webkit.WebView.loadDataWithBaseURL(java.lang.String, java.lang.String, java.lang.String, java.lang.String, java.lang.String)' on a null object reference
       at com.automattic.simplenote.NoteEditorFragment.loadMarkdownData(Unknown Source)
       at com.automattic.simplenote.NoteEditorFragment.access$2400(Unknown Source)
       at com.automattic.simplenote.NoteEditorFragment$saveNoteTask.onPostExecute(Unknown Source)
       at com.automattic.simplenote.NoteEditorFragment$saveNoteTask.onPostExecute(Unknown Source)
       at android.os.AsyncTask.finish(AsyncTask.java:651)
       at android.os.AsyncTask.access$500(AsyncTask.java:180)
       at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:668)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:148)
       at android.app.ActivityThread.main(ActivityThread.java:5417)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

nthcircle:
PR for fix here: https://github.com/Automattic/simplenote-android/pull/453

Issue #423, opened 2017-02-03
Android tablet: App crashes in landscape mode


fditrapani:
Related to to issue #3034317-t. A customer wrote in with the following issue:

> Happened on a Lenovo tablet wich Android Kitkat 4.4. When I opened a note in Portrait view and then turned the tablet in landscape mode, it automatically crashed.

theck13:
I wasn't able to reproduce the crash on two devices (Samsung Galaxy S III on Android 4.4.4 and Samsung Galaxy Tab 3 on Android 4.4.2) or an emulator (Nexus 5 on Android 4.4.2) running Simplenote 1.5.5.  I tested opening a note in portrait orientation then rotating into landscape orientation without Markdown enabled, with Markdown enabled while viewing the Edit tab, and with Markdown enabled while viewing the Preview tab.

We might need to find out exactly which device model, what flavor of Android (i.e. stock, AOSP, AOKP, CyanogenMod, etc.), and what version of Simplenote is being used.  A stack trace would be extremely helpful too.

fditrapani:
Thanks for the reply @theck13. I'll follow up with the customer to see if they can provide us with more info.

nthcircle:
I just noticed this bug. It's related to stack trace logged here:
https://github.com/Automattic/simplenote-android/issues/444

I've submitted a PR with a fix, along with step to reproduce, here:
https://github.com/Automattic/simplenote-android/pull/453

PR #453, opened 2017-06-24
Fix NPE when callbacks attempt to load markdown in the wrong view.

nthcircle:
Fixes https://github.com/Automattic/simplenote-android/issues/444 and https://github.com/Automattic/simplenote-android/issues/423

Bug is reproducible with these steps:

1. Device is a tablet (large screen)
2. Markdown is enabled
3. User starts NotesActivity in _portrait_ orientation
4. User selects a note, which starts NoteEditorActivity (instead of remaining in the NotesActivity's master-detail flow, which occurs if device is in landscape)
5. NoteEditorFragment is loaded into the NoteEditorActivity's ViewPager, along with a sibling NoteMarkdownFragment, which is now the correct target for the markdown data.

Previous code was assuming large screen size + landscape orientation (`DisplayUtils.isLargeScreenLandscape(getActivity())`) meant markdown should display in the NoteEditorFragment's embedded WebView. Callbacks were attempting to load the markdown data into this null WebView, causing NPEs.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -1162,14 +1162,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
             // Show tabs if markdown is enabled globally, for current note, and not tablet landscape
             if (mIsMarkdownEnabled) {
                 // Get markdown view and update content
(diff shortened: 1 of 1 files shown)


roundhill:
Sorry for the huge delay in reviewing this, I was on sabbatical. Will have a look this week and thanks for the contribution.

roundhill:
I had a hard time reproducing the NPE, but this looks good 👍 

Thanks!

PR #459, opened 2017-08-20
New Feature: Add App shortcut for api 25 and above

maheshgaya:
Beginning from api 25, apps can have shortcuts on their launch icon. It would be nice to have some shortcuts for Simplenote. I have added a shortcut for creating a new note. Let me know what you think.

Note: if you are compiling the debug version, change android:targetPackage in res/xml/shortcuts.xml to "com.automattic.simplenote.debug"

Here is an example:

screenshot_20170819-220435


roundhill:
Cool! Sorry for the delay, will review this soon.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -198,6 +199,11 @@ public void onViewCreated(View view, Bundle savedInstanceState) {
 
         NotesActivity notesActivity = (NotesActivity) getActivity();
 
+        if (ACTION_NEW_NOTE.equals(notesActivity.getIntent().getAction())){
+            //if user tap on "app shortcut", create a new note
```
roundhill: We should bail out here if the user isn't signed in so a new note isn't created in that case. You can use `NotesActivity.userIsUnauthorized()` for that. It's ok to make that private method public, too.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -221,21 +227,25 @@ public void onClick(View v) {
...
     }
 
+    private void createNewNote(){
+        if (!isAdded()) return;
+
+        addNote();
+        AnalyticsTracker.track(
+                AnalyticsTracker.Stat.LIST_NOTE_CREATED,
+                AnalyticsTracker.CATEGORY_NOTE,
+                "action_bar_button"
```
maheshgaya: @roundhill Do you want this to be parameterized? For instance this will become `new_note_shortcut` for this new feature and `action_bar_button` for when the user taps on the fab.

Simplenote/src/main/res/xml/shortcuts.xml
```diff
@@ -0,0 +1,15 @@
...
+    xmlns:tools="http://schemas.android.com/tools"
+    tools:targetApi="25">
+    <shortcut
+        android:shortcutId="shortcut_new_note"
+        android:enabled="true"
+        android:icon="@drawable/ic_create_24dp"
+        android:shortcutShortLabel="@string/new_note">
+        <intent
+            android:action="com.automattic.simplenote.NEW_NOTE"
+            android:targetPackage="com.automattic.simplenote"
```
maheshgaya: @roundhill This will be different in `debug` version. Should I add a copy of the `shortcuts.xml` in the debug folder? The targetPackage will be `com.automattic.simplenote.debug` for debug version.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
maheshgaya: Should we show an error message or exit the app if the user is unauthorized or both?

maheshgaya: Never mind, it actually redirects to LoginActivity

Simplenote/src/main/res/xml/shortcuts.xml
roundhill: Yeah good idea 👍

PR #468, opened 2017-10-09
Update Dependency Versions

roundhill:
The app has fallen quite behind in build tools/ support library versions.

**To Test**
Build and launch the app, and give it a smoke test to verify that areas the support libraries touch still work as expected.

_Note: I did test a release build of this branch with the wear app, and it worked fine._


maheshgaya:
Are we updating Gradle as well?

Issue #469, opened 2017-10-10
Trash Action Shows Disabled Icon for Enabled State


theck13:
### Observed
The trash icon in the navigation drawer and the toolbar show a disabled state when the action is actually enabled.  

### Expected
The trash icon in the navigation drawer and the toolbar show a disabled state when the action is disabled and an enabled state when the action is enabled.

### Reproduced
1. Go to ***Trash***.
2. Empty trash if trashed notes exist.
3. Notice disabled ***Trash*** icon and disabled action.
4. Go to ***All Notes***.
5. Open any note.
6. Notice disabled ***Trash*** icon and enabled action.

See the screenshots below for illustration.
![trash](https://user-images.githubusercontent.com/3827611/31414586-e0ca7206-addb-11e7-9324-ff65ed82487c.png)

#### Tested
Google Pixel on Android 7.1.1 with Simplenote 1.5.6

PR #470, opened 2017-10-10
Issue/469 trash icon state

theck13:
### Fix
Update empty trash icon state logic as described in https://github.com/Automattic/simplenote-android/issues/469.

### Test
1. Go to ***Trash***.
2. Empty trash if trashed notes exist.
3. Notice disabled ***Trash*** icon and disabled action.
4. Go to ***All Notes***.
5. Open any note.
6. Notice disabled ***Trash*** icon and enabled action.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -38,7 +38,6 @@ import android.widget.ProgressBar;
 import com.automattic.simplenote.analytics.AnalyticsTracker;
 import com.automattic.simplenote.models.Note;
 import com.automattic.simplenote.models.Tag;
-import com.automattic.simplenote.utils.AniUtils;
 import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.DrawableUtils;
 import com.automattic.simplenote.utils.HtmlCompat;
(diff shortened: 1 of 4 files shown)


dmsnell:
Can you help a dope understand this? How does the icon turn lighter in color? is that a default behavior of the disabled state?

theck13:
Sure!

The [`ic_trash_selector.xml`](https://github.com/Automattic/simplenote-android/pull/470/files#diff-922ce2028ac2e7ab2096ee83a017ee1c) file tells the system which drawable (i.e. icon) to use based on the states listed in that file.  The [`ic_trash_disabled_24dp`](https://github.com/Automattic/simplenote-android/pull/470/files#diff-5b93f2745f51ad7cd06f93344b46af2e) is used when `android:state_enabled="false"` (i.e. disabled) and the `ic_trash_24dp` is used when the other states aren't met (i.e. default).  In this case, the only states listed in that selector are not enabled and default.

The [`ic_trash_disabled_24dp`](https://github.com/Automattic/simplenote-android/pull/470/files#diff-5b93f2745f51ad7cd06f93344b46af2e) is lighter because the fill color is [#80000000](https://github.com/Automattic/simplenote-android/pull/470/files#diff-5b93f2745f51ad7cd06f93344b46af2eR11).  The first two hexadecimal digits represent the alpha and the last six hexadecimal digits are the color.  So that drawable is 50% black.  That's why it appears lighter.

In the world of Android, a disabled action is represented with this lighter color _and_ shows no touch feedback when the icon is long-pressed.  In the screenshots below, I'm touching the ***Trash*** icon.  The title of each screenshot describes whether the right or wrong behavior is shown.

state

dmsnell:
thanks for the great explanation @theck13 - this PR makes so much more sense now!

roundhill:
Looks good and tested well, I also ran it on a 4.4 emulator to make sure the vector was working properly.

PR #475, opened 2017-11-22
Fixing build for latest version of gradle

roundhill:
I was getting this error in the console when building:

https://developer.android.com/studio/build/gradle-plugin-3-0-0.html?utm_source=android-studio#known_issues

I moved the gradle plugin dependency to the main `build.gradle` and deleted it from the sub modules, and everything is happy again.

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -4,7 +4,6 @@ buildscript {
         maven { url 'https://maven.fabric.io/public' }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.3'
         classpath 'io.fabric.tools:gradle:1.+'
     }
 }
--- a/Simplenote/src/main/res/layout/auth_preference.xml
+++ b/Simplenote/src/main/res/layout/auth_preference.xml
@@ -28,7 +28,7 @@
         android:layout_weight="1"
         android:ellipsize="marquee"
         android:gravity="right"
-        android:maxLines="1"
+        android:singleLine="true"
         android:textAlignment="gravity"
         android:textAppearance="?android:attr/textAppearanceMedium"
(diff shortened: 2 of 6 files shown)


0nko:
When I try to build the project I get the following error: 

```Due to a limitation of Gradle’s new variant-aware dependency management, loading the Android Gradle plugin in different class loaders leads to a build error.
This can occur when the buildscript classpaths that contain the Android Gradle plugin in sub-projects, or included projects in the case of composite builds, are set differently.
To resolve this issue, add the Android Gradle plugin to only the buildscript classpath of the top-level build.gradle file.
In the case of composite builds, also make sure the build script classpaths that contain the Android Gradle plugin are identical across the main and included projects.
If you are using a version of Gradle that has fixed the issue, you can disable this check by setting android.enableBuildScriptClasspathCheck=false in the gradle.properties file.
```

It can be fixed by removing the `buildscript` and `repositories` from individual submodules and putting everything into the top-level `build.gradle`:

```
buildscript {
    repositories {
        jcenter()
        maven { url 'https://maven.fabric.io/public' }
    }
    dependencies {
        classpath 'com.android.tools.build:gradle:3.0.1'
        classpath 'io.fabric.tools:gradle:1.+'
    }
}

allprojects {
    repositories {
        jcenter()
        maven { url "https://s3.amazonaws.com/repo.commonsware.com" }
        maven { url 'https://maven.fabric.io/public' }
        maven { url "https://maven.google.com" }
    }
}
```

roundhill:
Hmm, I thought that's what I did here:

https://github.com/Automattic/simplenote-android/pull/475/files#diff-c197962302397baf3a4cc36463dce5eaR2

Does cleaning help it at all?

0nko:
Sure did :). :shipit:

PR #478, opened 2017-11-29
Fix line height squashing in editor

### LINKED ISSUE #450: Line Spacing Changes When Editing Note
On the latest Android version (7.1.2) with the latest Simplenote app, it's okay when I open a note without altering it. But as soon as I change anything (add something, delete something), the spacing around that particular line contracts, so that it's uneven across the note, and so that the edited lines are harder to read and edit. The more edits I make, the more scrambled the note becomes.

Screenshot:
screenshot_20170616-161538

3258766-t

roundhill:
To fix #450 I had to remove the `lineSpacingExtra` value in the EditText. This unfortunately makes the line heights a bit more compact in the editor now:

screen shot 2017-11-29 at 9 41 30 am

I also tried using `lineSpacingMultiplier` instead but got the same results.

**To Test**
Add a note with multiple lines, like:

`Line 1
Line 2
Line 3
Line 4`

Using the soft keyboard, delete the `3` in `Line 3`. The line heights of all text should remain the same.

This PR depends on #475

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -4,7 +4,6 @@ buildscript {
         maven { url 'https://maven.fabric.io/public' }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.3'
         classpath 'io.fabric.tools:gradle:1.+'
     }
 }
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -57,7 +57,7 @@ android {
         applicationId "com.automattic.simplenote"
         versionName project.version
         versionCode 59
-        minSdkVersion 15
+        minSdkVersion 16
         targetSdkVersion 26
 
         vectorDrawables.useSupportLibrary = true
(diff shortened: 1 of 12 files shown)


0nko:
This could probably be fixed with a workaround mentioned here: https://issuetracker.google.com/issues/37009353 (at the bottom). 

We have the same problem in Aztec but I didn't know what was causing it. The code above seems to fix this.

roundhill:
Nice fine. Hmm the workaround seems pretty intense! We can try it out on a low end device with a really long note to see how it performs?

0nko:
I was trying to run it on my old HTC phone but it keeps rebooting so no luck there. Unless you have a device to test it we can just merge your fix now and leave this for later.

roundhill:
> Unless you have a device to test it we can just merge your fix now and leave this for later.

I tried it on a Galaxy S6 (I know, not that old) but it was able to edit the entire text of 'Romeo and Juliet' without any lag. I do like the workaround because it looks so much better with the increased line height.

I bumped minSdk to 16 to avoid an SDK check for `getLineSpacingExtra` access. I also did the check in `afterTextChanged()` because we need to calculate the title text size first before making the adjustment.

0nko:
Cool. Let's keep it then! 👍 
:shipit:

Cor-13:
Thank you for taking care of this issue! It's fixed as of the new app version 1,5.7. Much appreciated!

Issue #430, opened 2017-02-15
Search by tag


memeplex:
If this is already possible I'm not able to figure out how to do it. Neither tag:<tag> nor plain <tag> match anything.

ericbrandwein:
For the same reason, there is no way of finding notes by multple tags, at least not one that I'm aware of.

robertoissc:
I believe this should be closed. :)

theck13:
Yep, this can be closed.  That feature was added in https://github.com/Automattic/simplenote-android/pull/479.

PR #479, opened 2017-11-30
Adds multiple tag search

ericbrandwein:
You can now search for notes with tag "tag1" AND tag "tag2", and with text "text", by writing "tag:yourtag1 tag:yourtag2 text" in the searchbar. This solves #430. 
Maybe it isn't the best way of implementing it, but I did it to show that it is not so hard to implement, and it was the only thing stopping me from migrating from Google Keep. Hope it helps!

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -7,7 +7,6 @@ import android.database.sqlite.SQLiteException;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Handler;
-import android.os.PatternMatcher;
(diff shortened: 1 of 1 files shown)


roundhill:
Thanks @ericbrandwein! I'll give this a spin soon.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -7,10 +7,12 @@
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Handler;
+import android.os.PatternMatcher;
```
roundhill: Looks like this import isn't needed?

roundhill:
Thanks again @ericbrandwein! :shipit:

PR #480, opened 2017-12-14
New release checks script

maxme:
To make sure we have the correct keystore and properties before we ship the app.

cc @roundhill

### DIFF
--- a/tools/release-checks.sh
+++ b/tools/release-checks.sh
@@ -32,9 +32,9 @@ function checkGradleProperties() {
 }
 
 function checkKeystore() {
-  keystore=`cat gradle.properties | grep storeFile | cut -d= -f 2 | sed -e 's/^[ \t]*//'`
+  keystore=`cat gradle.properties | grep storeFile | cut -d= -f 2 | sed -e 's/^[ ]*//'`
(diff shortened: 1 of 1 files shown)


maxme:
Note, to run it:

```shell
$ ./tools/release-checks.sh
```

tools/release-checks.sh
```diff
@@ -0,0 +1,42 @@
...
+  pOk
+}
+
+function checkGradleProperties() {
+  /bin/echo -n "Check gradle.properties..."
+  checkFileAgainstHash $APPDIR/gradle.properties 2e36d6696bc71af3bc976b838966ac40cf06e4fe
+}
+
+function checkKeystore() {
+  keystore=`cat gradle.properties | grep storeFile | cut -d= -f 2 | sed -e 's/^[ ]*//'`
```
roundhill: Are we meant to run this from the root app directory? If so I needed to change this to: `cat $APPDIR/gradle.properties`

roundhill:
Once I had the files in the right place, it worked well!

```
Check gradle.properties...[OK]
Check Keystore...[OK]
Simplenote/build.gradle version 1.5.7
```

:shipit:

PR #481, opened 2017-12-14
Markdown Preview Updates

roundhill:
We received a hackerone report that the app doesn't sanitize/escape html in the markdown view. Because of the shared note system, a markdown note with an html form in it could be used to gather sensitive data from a user.

The latest update to `AndDown` allows for escaping html output 👍

I also added a few extension so that we can get strikethrough, 

**To Test**
* Create a note with markdown.
* Add HTML content and markdown syntax.
* When previewing the note, the markdown should render and the html tags should be escaped.

Here's some example markdown content to test with:

```
# Cool

<h1>Cool</h1>

[click](https://wp.com)

![image](http://cdn.abclocal.go.com/content/wpvi/images/cms/2163296_1280x720.jpg)

<script>alert('hi')</script>

Naughty html:
<input type="text" />
<form><input type="submit" /></form>

~~strike~~

> Quoted

```

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -38,7 +38,7 @@ dependencies {
     // Third Party
     compile 'com.takisoft.fix:preference-v7:24.2.1.0'
     compile 'com.loopj.android:android-async-http:1.4.9'
-    compile 'com.commonsware.cwac:anddown:0.2.4'
+    compile 'com.commonsware.cwac:anddown:0.4.0'
     compile('com.crashlytics.sdk.android:crashlytics:2.6.5@aar') {
         transitive = true;
     }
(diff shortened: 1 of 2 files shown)


roundhill:
> Is there any reason why we shouldn't enable all the extensions in AndDown?

I was trying to have it match up with the syntax that we support in the other apps. I don't think we support things like math elsewhere. We do support tables, but it didn't seem to work when I added the flag. Maybe we just need to add some CSS for it?

theck13:
> I was trying to have it match up with the syntax that we support in the other apps.

That makes sense to me.

> We do support tables, but it didn't seem to work when I added the flag. Maybe we just need to add some CSS for it?

That sounds like a rabbit hole.  If you're okay with leaving out tables for now, so am I.  The code looks good as is and it's ready to merge.

PR #482, opened 2017-12-15
Issue/preview padding clip

theck13:
### Fix
Add the `android:clipToPadding="false"` attribute and value to the `NestedScrollView` of the layout for the ***Preview*** tab to avoid clipping the `WebView` to the parent view's padding.  See the screenshots below for illustration.

clip


roundhill:
:shipit:

PR #487, opened 2018-01-12
Add Monospace Font Editor Option

roundhill:
This PR adds support for using a monospace font in the editor. @drw158 did some research and we decided up on using iAWriter Duospace:

screenshot_1515717970

I also updated the settings to include a new `Text Appearance` section which includes size and the new toggle for the monospace font:

screenshot_1515718004

We were using an old 3rd party lib for preferences, I booted that out and used the support library version.

**To Test**
* Open the settings and toggle the monospace option.
* Edit a note, and the font should be monospace and editing should work normally, including linking of urls.
* Go back to the settings and disable the monospace option and return to the editor. It should be back to using the Roboto font.


nbradbury:
I'm not sure what the cause is, but when I run this on an API 16 emulator, the section headings in preferences lose their left margins. This doesn't happen with `develop`. Other than that, this looks good.

screenshot_1515874636

nbradbury:
:shipit:

PR #490, opened 2018-01-25
Added a warning when signing out with unsynced notes.

roundhill:
Some users have attempted to sign out of the app to sync fixing issues, not realizing that all notes get deleted when doing so. This PR adds a prompt to show the user if they sign out and have notes that have never been synced. An unsynced note can be identified as having a `version` of `0`.

screen shot 2018-01-31 at 1 09 02 pm

We give the user the option to view the web app to reconcile which notes haven't synced. The performance was really fast for me for an account with 1,000 notes. But just in case I've put the process of checking for the unsynced note in an `AsyncTask`.

**To Test**
* Create a note while offline.
* Go to settings, tap `Sign Out`.
* Verify you received the prompt and that all of the buttons work as expected.
* Verify that signing out with all notes synced does not show the prompt, and just signs you out.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
@@ -154,41 +154,6 @@ public class PreferencesFragment extends PreferenceFragmentCompat implements Use
         });
     }
 
-    private static class SignOutAsyncTask extends AsyncTask<Void, Void, Boolean> {
-        private WeakReference<PreferencesFragment> fragmentWeakReference;
-
-        SignOutAsyncTask(PreferencesFragment fragment) {
-            fragmentWeakReference = new WeakReference<>(fragment);
-        }
-
-        @Override
-        protected Boolean doInBackground(Void... voids) {
-            PreferencesFragment fragment = fragmentWeakReference.get();
-            return fragment == null || fragment.hasUnsyncedNotes();
-        }
-
-        @Override
-        protected void onPostExecute(Boolean hasUnsyncedNotes) {
-            PreferencesFragment fragment = fragmentWeakReference.get();
-            if (fragment == null) {
-                return;
-            }
-
-            // Safety first! Check if any notes are unsynced and warn the user if so.
-            if (hasUnsyncedNotes) {
-                AlertDialog.Builder builder = new AlertDialog.Builder(fragment.getContext());
-                builder.setTitle(R.string.unsynced_notes);
-                builder.setMessage(R.string.unsynced_notes_message);
-                builder.setPositiveButton(R.string.delete_notes, fragment.signOutClickListener);
-                builder.setNeutralButton(R.string.visit_web_app, fragment.loadWebAppClickListener);
-                builder.setNegativeButton(R.string.cancel, null);
-                builder.show();
-            } else {
-                fragment.signOut();
-            }
-        }
-    }
-
     private DialogInterface.OnClickListener signOutClickListener = new DialogInterface.OnClickListener() {
         @Override
         public void onClick(DialogInterface dialogInterface, int i) {
@@ -203,7 +168,7 @@ public class PreferencesFragment extends PreferenceFragmentCompat implements Use
         }
     };
 
-    public boolean hasUnsyncedNotes() {
+    private boolean hasUnsyncedNotes() {
         Simplenote application = (Simplenote) getActivity().getApplication();
         Bucket<Note> notesBucket = application.getNotesBucket();
         Bucket.ObjectCursor<Note> notesCursor = notesBucket.allObjects();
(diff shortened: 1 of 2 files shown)


roundhill:
Refs: https://github.com/Automattic/Simplenote-United/issues/7

maheshgaya:
There is an issue with this, maybe a performance issue (because I have only have a total of 5 notes with 1 unsynced not). Steps:
1. Go offline
2. Create a new note. Save and directly go to the settings page
3. Now, go online. Try to sign out immediately.

It takes a while before it finally syncs. (approx. 5 to 10 secs)

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -158,6 +154,88 @@ public boolean onPreferenceChange(Preference preference, Object o) {
...
+    };
+
+    private DialogInterface.OnClickListener loadWebAppClickListener = new DialogInterface.OnClickListener() {
+        @Override
+        public void onClick(DialogInterface dialogInterface, int i) {
+            startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse(WEB_APP_URL)));
+        }
+    };
+
+    public boolean hasUnsyncedNotes() {
```
maheshgaya: can be `private`. It will be passed to the asyncTask since it's a subclass of this class

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -158,6 +154,88 @@ public boolean onPreferenceChange(Preference preference, Object o) {
...
+            Note note = notesCursor.getObject();
+            if (note.getVersion() == 0) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    void signOut() {
```
maheshgaya: can be `private`. It will be passed to the asyncTask since it's a subclass of this class

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -158,6 +154,88 @@ public boolean onPreferenceChange(Preference preference, Object o) {
...
+        @Override
+        protected void onPostExecute(Boolean hasUnsyncedNotes) {
+            PreferencesFragment fragment = fragmentWeakReference.get();
+            if (fragment == null) {
+                return;
+            }
+
+            // Safety first! Check if any notes are unsynced and warn the user if so.
+            if (hasUnsyncedNotes) {
+                AlertDialog.Builder builder = new AlertDialog.Builder(fragment.getContext());
```
maheshgaya: There maybe no need for this variable here. And this may be more readable since it's a `builder`:
```java
new AlertDialog.Builder(fragment.getContext())
        .setTitle(R.string.unsynced_notes)
        .setMessage(R.string.unsynced_notes_message)
        .setPositiveButton(R.string.delete_notes, fragment.signOutClickListener)
        .setNeutralButton(R.string.visit_web_app, fragment.loadWebAppClickListener)
        .setNegativeButton(R.string.cancel, null)
        .show();
```

or 

```java
AlertDialog.Builder builder = new AlertDialog.Builder(fragment.getContext())
        .setTitle(R.string.unsynced_notes)
        .setMessage(R.string.unsynced_notes_message)
        .setPositiveButton(R.string.delete_notes, fragment.signOutClickListener)
        .setNeutralButton(R.string.visit_web_app, fragment.loadWebAppClickListener)
        .setNegativeButton(R.string.cancel, null);
builder.show();
```

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -10,19 +13,25 @@
...
 import com.takisoft.fix.support.v7.preference.SwitchPreferenceCompat;
 
+import java.lang.ref.WeakReference;
+
 /**
  * A simple {@link Fragment} subclass.
  */
 public class PreferencesFragment extends PreferenceFragmentCompat implements User.StatusChangeListener, Simperium.OnUserCreatedListener {
 
+    private static final String WEB_APP_URL = "https://app.simplenote.com";
```
maheshgaya: This can be in `string.xml`

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
maheshgaya: Also, move the subclass `SignOutAsyncTask` to be the last thing in the file (after `onUserCreated` method)

daniloercoli:
@maheshgaya made some good comments. I agree with him, we can try to synch on logout if the device is online and there are unsynched notes.

Also note that `hasUnsyncedNotes` only checks if there are notes with `version` = `0`, so if you modify a note locally, when the device is offline, that routine doesnt return properly.

roundhill:
> Also note that hasUnsyncedNotes only checks if there are notes with version = 0, so if you modify a note locally, when the device is offline, that routine doesnt return properly.

Yes, this only checks for _never_ synced notes. I'm not quite sure how to check for local changes, but it  should be possible! I'll poke around.

roundhill:
> but it should be possible!

Found some methods in Simperium's `syncable` object that made this possible. It now checks for local modifications in addition to new notes. Added in fbcbadb

daniloercoli:
LGTM! :shipit:

PR #492, opened 2018-02-01
Updated Simplenote logo drawable

roundhill:
We were using an older 'skinny s' logo in a few places, including the login screen:

screenshot_1517504853

After updating, it now looks like this:

screenshot_1517507400

I also tweaked the sizing of the logo that is used for the tablet placeholder to match the DPI:

screenshot_1517507856


0nko:
:shipit:

PR #493, opened 2018-02-01
Use dynamic copyright date in about fragment

roundhill:
The copyright is now two years old in the about fragment:

screen shot 2018-02-01 at 10 28 33 am

Since it doesn't require translation I've taken it out of `strings.xml` and formatted it directly in the fragment. It was also requested to remove `, inc` and add a space between the copyright symbol and the year:

screen shot 2018-02-01 at 10 28 41 am

We will have some leftover `automattic_copyright` strings in the translated files, but those will go away when we update the translations for the next release.

@0nko care to take a look?


0nko:
:shipit:

PR #497, opened 2018-02-26
Disable monospace option for next release

roundhill:
We're not quite ready with the other clients to add the monospace font option yet, so I'm reverting #487 for this next release.

**To Test**
* Load the app settings, you shouldn't see the monospace option.
* The post editor should function normally.


0nko:
Looks good. :shipit:

PR #504, opened 2018-03-13
Use Higher Font Size Values for Large and High DPI Screens

roundhill:
On a Chrome OS device with a high DPI screen, the default font size of 16pt is too small. This PR bumps up the font sizes for those devices. The new default 'Normal' font size is 18pt.

Before:
screenshot 2018-03-13 at 12 49 25 pm

After:
screenshot 2018-03-13 at 12 49 52 pm

**To Test**
- Ensure font sizes are normal/readable on a device like a phone.
- Check font size readability on a large/high dpi screen device (I have a Samsung Chromebook Pro)

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -335,7 +335,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
         mTagView.setOnTagAddedListener(this);
 
         if (mContentEditText != null) {
-            mContentEditText.setTextSize(TypedValue.COMPLEX_UNIT_SP, PrefUtils.getIntPref(getActivity(), PrefUtils.PREF_FONT_SIZE, 14));
+            mContentEditText.setTextSize(TypedValue.COMPLEX_UNIT_SP, PrefUtils.getFontSize(getActivity()));
         }
     }
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -189,7 +189,7 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
     protected void getPrefs() {
         boolean condensedList = PrefUtils.getBoolPref(getActivity(), PrefUtils.PREF_CONDENSED_LIST, false);
         mNumPreviewLines = (condensedList) ? 0 : 2;
-        mPreviewFontSize = PrefUtils.getIntPref(getActivity(), PrefUtils.PREF_FONT_SIZE, 14);
+        mPreviewFontSize = PrefUtils.getFontSize(getActivity());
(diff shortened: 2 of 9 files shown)


kwonye:
@roundhill just an FYI I don't have a chromebook yet (and won't til next Friday at the earliest)

maxme:
Changeset looks good - I'll merge after @iamthomasbishop approval.

iamthomasbishop:
Apologies, but I also don't have a Chromebook on hand (should have one soon) so I'm not able to test first-hand. I don't have a strong opinion on this either way, so feel free to ignore me. 

Some initial questions - mostly coming from a place of cluelessness in terms of Chrome OS:

* How do other similar apps handle this on Chrome OS?
* Is there anything we let the OS handle in terms of accessibility/font-size settings?
* If not, would it make more sense to have an app-wide setting to adjust font size? For example: a setting that lets you choose a specific value _or_ "Small", "Medium", "Large", "Extra Large" for text/font-size and then adjusts the relative font sizes across the app (specific to a single device)

As mentioned previously, not a strong opinion either way. I am generally okay with bumping up font sizes, just want to understand the context and options :)

roundhill:
> How do other similar apps handle this on Chrome OS?

It's pretty hard to tell, but most apps appear to have their own font size settings,

> Is there anything we let the OS handle in terms of accessibility/font-size settings?

There is an option on my device to change the font size, but setting it to 'Medium' seems way too small to my eyes, which is why I made this PR.

> If not, would it make more sense to have an app-wide setting to adjust font size? For example: a setting that lets you choose a specific value or "Small", "Medium", "Large", "Extra Large" for text/font-size and then adjusts the relative font sizes across the app (specific to a single device)

The app has this already, but the 'Normal' setting was too small which is why I've adjusted it for these larger screen/dpi devices.

iamthomasbishop:
> The app has this already, but the 'Normal' setting was too small which is why I've adjusted it for these larger screen/dpi devices.

Oh fair enough, I hadn't seen this (looked on iOS - hadn't looked on Android).

One thought - and I'm not sure what font-size each are of each level - but if "Large" is ~18pt, you could just have it set to "Large" by default. That would mean you don't really need to adjust font-sizes, just what the setting defaults to. But if you're looking to shift _all_ of the values relatively up a couple pt, then this probably isn't the right way to go.

roundhill:
> One thought - and I'm not sure what font-size each are of each level - but if "Large" is ~18pt, you could just have it set to "Large" by default. That would mean you don't really need to adjust font-sizes, just what the setting defaults to. But if you're looking to shift all of the values relatively up a couple pt, then this probably isn't the right way to go.

That would be an easier solution for sure. There's only one 'Extra large' option available after that, but maybe not a big deal?

roundhill:
> That would be an easier solution for sure.

Actually, the only difference would be to remove the extra `arrays.xml` entries for the larger font sizes. So I'm voting to leave this as-is.

roundhill:
@maxme @iamthomasbishop and I agreed that we should use this solution for now in Slack, so please merge at your convenience :)

maxme:
:shipit:

PR #505, opened 2018-03-14
Fix key navigation from note editor to tags field.

roundhill:
Pressing the down arrow key should now focus the tags field when the note editor is active. Similarly, pressing up while the tags editor is active should focus the note editor again.


maxme:
Looks good - :shipit:

PR #506, opened 2018-03-15
Right-click support

roundhill:
Adds support for right-clicking on the list view to start multi select note mode.

**To Test**
* Fire up the app with a mouse connected.
* Right-click on an item in the notes list.
* Multi-select should now be active.
* Right-clicking again should close multi-select mode.
* Regular clicks should work normally.


maxme:
Simple but great feature - makes the app more like a normal chrome app - :shipit:

PR #508, opened 2018-03-17
Update Simperium to 0.6.9

roundhill:
Updates Simperium to 0.6.9, which contains some nice Chrome OS/large screen fixes for the `LoginActivity`.

Details: https://github.com/Simperium/simperium-android/pull/197

The main change you should see when running this is that the login form will be centered:

screenshot 2018-03-13 at 3 36 18 pm

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -38,7 +38,7 @@ dependencies {
     }
 
     // Automattic and WordPress dependencies
-    compile "com.simperium.android:simperium:0.6.8"
+    compile "com.simperium.android:simperium:0.6.9"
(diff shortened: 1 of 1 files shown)


maxme:
I think that we should move the UI parts from Simperium to Simplenote to setup Smart Lock for Passwords (there is a helper class in wpandroid that will make the implementation quite easy).

Looks good, :shipit:

roundhill:
> I think that we should move the UI parts from Simperium to Simplenote to setup Smart Lock for Passwords (there is a helper class in wpandroid that will make the implementation quite easy).

+1000

PR #514, opened 2018-03-26
Revert right click to start multi select

roundhill:
Reverting #506 based on feedback from Google.

Revert "Adding support for right-clicking on the list view to start multi select note mode."

This reverts commit c4aa0fb826925b3e77526030906de3fe44dbf0e2.

**To Test**
* Right clicking the notes list should no longer start multi select mode.
* Long press should still start multi select mode.


maxme:
😭 :shipit:

roundhill:
things

PR #519, opened 2018-04-13
Add right-click popup menu to the notes list

roundhill:
I'm back with another right-click PR :) Right-clicking an item in the notes list will now show a popup menu with the pin and trash options:

screen shot 2018-04-13 at 12 39 05 pm

**To Test**
* Right-click on a note. You should be able to trash and pin/unpin it successfully. If the note is pinned the menu title should change to `Unpin from top`.
* View the trash, and right-click on a note. There should be one menu option, `Restore` that will bring the note out of the trash.
* If you right-click on a note that is selected and delete or restore it, the editor placeholder should become visible.
* If you right-click on a note that is not selected and delete or restore it, the note list should update but the currently selected note should remain active.
* When trashing a note, the 'undo bar' should function properly.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -1,5 +1,6 @@
 package com.automattic.simplenote;
 
+import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.Intent;
 import android.database.Cursor;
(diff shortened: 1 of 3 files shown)


kwonye:
> When trashing a note, the 'undo bar' should function properly.

What does this mean, sorry?

kwonye:
Got clarification that it's the snackbar with the undo action, thanks @roundhill!

PR #523, opened 2018-04-23
Fix notes list content preview line length in on Oreo

### LINKED ISSUE #522: Simplenote Condensed view not Working on Android 8
The user updated phone software to Android 8, Samsung Experience 9 and simplenote app condensed note view is no longer condensed.

roundhill:
Fixes #522 by setting visibility of content preview textview instead of using `maxLines` set to zero.

Apparently in Oreo they've change what `maxLines` does if set to zero on a TextView, it seems to remove the limit:

screen shot 2018-04-23 at 9 41 37 am

We were previously setting the `maxLines` property to zero to essentially hide the note content preview. We can instead simply set the `visibility` of the textview to `gone` when the condensed note list is set.

I also fixed a font size issue with the `new note` placeholder text. It was set to a hard coded size in the styles. This adds an `AbsoluteSizeSpan` so that the font size can be dynamic and look correct no matter what font size was selected in the settings:

screen shot 2018-04-23 at 9 45 03 am

**To Test**
* Open the app in an Oreo emulator.
* In the settings, toggle 'condensed note list' to on.
* Return to the notes list. You should see only one line of text.
* Turn condensed notes list off, there should be up to two lines of preview text.
* The `New note...` placeholder text for an empty note should be the same font size as other notes in the list.
* Repeat the above on an emulator older than Oreo.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -12,6 +12,7 @@ import android.support.design.widget.FloatingActionButton;
 import android.support.v4.app.ListFragment;
 import android.text.SpannableString;
 import android.text.TextUtils;
+import android.text.style.AbsoluteSizeSpan;
 import android.text.style.TextAppearanceSpan;
 import android.util.SparseBooleanArray;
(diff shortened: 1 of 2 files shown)


0nko:
👍:shipit:

Issue #500, opened 2018-02-28
Keep screen on while editing


gregmac:
I use my phone for meeting notes, and as such, often sit it on the table in front of me during a meeting. It would be nice in this situation if the screen would stay on so I didn't have to wake/unlock the device to quickly add something.

Specifically, while the editor is open, it should prevent the screen from sleeping ([Android reference:  FLAG_KEEP_SCREEN_ON](https://developer.android.com/training/scheduling/wakelock.html)). This maybe needs to be a setting as well (I don't really think so, but others might).

theck13:
Thanks for the suggestion!  I have added it to our feature requests.

PR #524, opened 2018-04-24
Added keep screen on while editing in setting preference.

SuhanKoh:
It allows user to decide to keep the screen on while editing notes.
Also moved one of the hardcoded string for condense note to use the one defined in PrefUtils.
Added #500 

keepscreenon


SuhanKoh:
Perhaps it could be on the tooltips on top of the editor, would that be too cluttered? May need some icons for this.

roundhill:
Hmm it'd probably not work to do it in the editor unless you restarted the activity after enabling it, right? Maybe better to keep the setting like you've done.

`NotesActivity` can contain the editor when on a large device in landscape. So we'd probably need the same checks added to that activity to keep the screen on.

SuhanKoh:
The checks is in ```NoteEditorFragment```, ```NotesActivity``` will load the fragment which would trigger the keep screen on flag, not sure if you meant move the check to ```NoteEditorActivity``` and ```NotesActivity```?

roundhill:
> The checks is in NoteEditorFragment, NotesActivity will load the fragment which would trigger the keep screen on flag, not sure if you meant move the check to NoteEditorActivity and NotesActivity?

You're right, I didn't realize you could enable it from a fragment 👍

roundhill:
@SuhanKoh I guess we let this die :) Did you want to still explore this? If you're able to enable from within the fragment then I think it'd be good to add a setting to the note info panel or something.

SuhanKoh:
@roundhill Hi Dan, I don't think I have the capacity to do this at the moment, maybe someone else can pick it up, otherwise close it for now.

Issue #464, opened 2017-09-30
Show notes without tag - find "orphans"


gubrist:
**Steps to reproduce/understand**
I write a lot of notes and any one is supposed to have one single tag in order to assign it to a specific topic. I organize my notes by these topics.

**What I expected**
Show notes without tag in order to find "orphans"

**What happened instead**
This functionality is not yet available.

**OS version**
Android 6

MorgothSauron:
+1. Showing orphans would be helpful to find notes for which I forgot to add a tag when I created it.

bummytime:
Looks like this was addressed in #527 — closing.

PR #527, opened 2018-04-27
Added a sidebar tab to view untagged notes.

SuhanKoh:
See: #464 
I have to modify setProperty(TAGS_PROPERTY, "") because when the note created, the tags are null, but when you remove a tag, the tags is actually empty string. Based on my limited understanding, I cannot do an or query.

screenshot_20180426-212506

### DIFF
--- a/Simplenote/src/main/res/values/strings.xml
+++ b/Simplenote/src/main/res/values/strings.xml
@@ -43,7 +43,7 @@
     <string name="search">Search notes</string>
     <string name="notes">All Notes</string>
     <string name="trash">Trash</string>
-    <string name="untaggedNotes">Untagged Notes</string>
+    <string name="untagged_notes">Untagged Notes</string>
     <string name="today">Today</string>
     <string name="yesterday">Yesterday</string>
     <string name="condensed_note_list">Condensed note list</string>
--- a/Simplenote/src/main/res/values/strings.xml
+++ b/Simplenote/src/main/res/values/strings.xml
@@ -43,6 +43,7 @@
     <string name="search">Search notes</string>
     <string name="notes">All Notes</string>
     <string name="trash">Trash</string>
+    <string name="untaggedNotes">Untagged Notes</string>
     <string name="today">Today</string>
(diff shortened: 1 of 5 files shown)


roundhill:
Since this will only be useful to a small amount of users, I think it'd be better to put this in a footer row of the tags list (if you have untagged notes) instead of a header item. Is that possible?

SuhanKoh:
Above the "Setting" text? Or before the divider line?

roundhill:
@drw158 do you have any thoughts on what the UX should be here? We'd need to apply this to all of the apps.

drw158:
Because of it's use, it should be within the tags list. I think it should be at the bottom of the tags list — not pinned or "sticky". So, in this case the button would sit directly below the "testing" tag.

Some design work needs to go into making it look better though. Because it would be in the tags list, it'll look strange with the icon. Although, you want it to stand out and not look like a regular tag.

SuhanKoh:
Can you provide a simple mock up?

drw158:
artboard

or if you want to add an icon

screen shot 2018-05-23 at 9 28 01 am

SuhanKoh:
Updated UI.

Simplenote/src/main/res/values/strings.xml
```diff
@@ -43,6 +43,7 @@
     <string name="search">Search notes</string>
     <string name="notes">All Notes</string>
     <string name="trash">Trash</string>
+    <string name="untaggedNotes">Untagged Notes</string>
```
roundhill: Can you please make this `untagged_notes` to match the style of the other string names?

SuhanKoh:
- [ ] Remove top and bottom margin
- [ ] Fix the string id to match existing style

SuhanKoh:
Screenshot on removed margin.
ss

roundhill:
Looks/works great! Thanks for the contribution @SuhanKoh!

alialwadie100:
make php function withe my email alialwadie@outlook.com

PR #530, opened 2018-05-22
Add WP.com Sign In

roundhill:
This PR adds support for signing to Simplenote via your WordPress.com account. It supports creating a new Simplenote account if one didn't exist yet, or alternatively it will sign the user to an existing account.

I utilized this library to handle the oauth work: https://github.com/openid/AppAuth-Android

wpcom-sign-in

**To Test**
* From the login screen tap 'Sign in with WordPress.com'
* Authorize a WordPress.com account
* 3 things should happen:
 ** If a Simplenote account already exists for the account, it should be signed in successfully.
 ** If a Simplenote account didn't exist for the account, a new one should be created and you'll see the welcome note.
 ** If the WordPress.com account hasn't been verified by email, you should get an error message.

We're storing the WP.com token in the keychain for future integrations. If a user signed out, it will be deleted.

_Note: The reviewer will need to add the client id to the config file. Ping me on slack to get the ID!_


Simplenote/src/main/java/com/automattic/simplenote/SignInActivity.java
```diff
@@ -0,0 +1,184 @@
...
+        super.onRestoreInstanceState(savedInstanceState);
+
+        if (savedInstanceState.containsKey(STATE_KEY_AUTH)) {
+            mAuthState = savedInstanceState.getString(STATE_KEY_AUTH);
+        }
+    }
+
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        if (requestCode == OAUTH_ACTIVITY_CODE) {
```
dmsnell: holy nesting batman. what if we performed an early-abort? this `if` doesn't even appear to have an `else`…

```java
if (requestCode != OAUTH_ACTIVITY_CODE) {
    return;
}

AuthorizationResponse authResponse…
```

and same going through this deeply-nested conditional tree

Simplenote/src/main/res/values/strings.xml
```diff
@@ -189,4 +191,6 @@
     <string name="restore">Restore</string>
     <string name="history">History</string>
     <string name="error_history">Couldn\'t retrieve history.</string>
+    <string name="or">or</string>
+    <string name="sign_in_with_wordpress_com">Sign in with WordPress.com</string>
```
theck13: Should the "i" in "in" be uppercase or lowercase in this case?  I noticed the other "Sign *" text on the screen uses uppercase (e.g. Sign Up, Sign In).

roundhill:
> Apart from those, I'm unable to sign in with my WordPress.com accounts.

Oops, I think you need a few more changes to gradle.properties, sent via Slack! It should work after that, let me know.

Simplenote/src/main/java/com/automattic/simplenote/SignInActivity.java
roundhill: I think I'll flatten it, I did similar early returns in other areas so best to match it up.

Simplenote/src/main/java/com/automattic/simplenote/utils/PrefUtils.java
```diff
@@ -48,6 +48,12 @@
     // boolean, determines if the theme was ever changed
     public static final String PREF_THEME_MODIFIED = "pref_theme_modified";
 
+    // string. WordPress.com access token
+    public static final String PREF_WP_TOKEN = "pref_key_wp_token";
+
+    // string. Private Simplenote preferences name
+    public static final String PREFS_PRIVATE_NAME = "prefs_simplenote";
```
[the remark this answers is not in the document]
roundhill: Oh! Didn't know it was private by default. I'll just add it to the default set then.

Simplenote/src/main/res/values/strings.xml
roundhill: I think I like it better in lowercase, but not quite sure which way is proper:

screen shot 2018-05-23 at 11 42 29 am

roundhill:
@theck13 I added a few more changes in 9d2096d. If those look good can we :shipit:?

PR #544, opened 2018-06-30
Add ukrainian translation file

uniss2209:


roundhill:
Thanks! We handle translations using GlotPress here:

https://translate.wordpress.com/projects/simplenote/android/uk/default/

Looks like they have been added and are waiting to be accepted by a validator.

Issue #350, opened 2016-04-09
Syncronization issue on Nexus 5


jerrysarcastic:
I had a user email us with the following bug report that affects synchronization on the Nexus 5 android version.  They left a pretty complete report so I am quoting it here with some edits

> There is a  difference in behaviour is that on the Nexus 5, going to the homescreen then re-opening Simplenote by tapping its icon in the launcher takes me to the list of notes.
> On the other Android device I tested on, when I go to the homescreen and re-open Simplenote from the launcher, it takes me directly to the note I was previously editing, skipping the list of notes.
> 
> This distinction is important, as to reproduce the sync issue on the Nexus 5, you need to re-open Simplenote from the recent apps list instead of tapping its icon in the launcher. The problem I describe only occurs if the note is already open when returning to the Simplenote app on Android.
> 
> Steps to reproduce on the Nexus 5 and stock Android 6.0.1:
> 1. In the Simplenote for Mac client (X), create a new note. I called it "Test note".
> 2. In the Simplenote for Android client (A), open the "Test note".
> 3. In X, append a new line with the text "X1". Observe that it syncs correctly to A.
> 4. In A, append a new line with the text "A1". Observe that it syncs correctly to X.
> 5. Press the Home button on A.
> 6. In X, append a new line with the text "X2".
> 7. Press the Recents button on A, then select Simplenote from the recent apps list.
> 8. Observe that the note now showing on A is missing the "X2" line.
> 9. In A, append a new line with the text "A2". Observe that in X, the "X2" line disappears, replaced with the "A2" line.
> 10. Press the Home button on A.
> 11. In X, append a new line with the text "X3".
> 12. Press the Recents button on A, and select Simplenote from the recent apps list.
> 13. Observe that the note now showing on A is missing the "X3" line.
> 14. In X, append a new line with the text "X4".
> 15. Observe that in X, both the "X3" and "X4" lines disappear after syncing. Observe that no changes show on A. 

Reported in 2616443

roundhill:
Dupe of #243

Issue #290, opened 2015-05-29
Synchronisation problems


rachelsquirrel:
A user reports that when they delete a note from the web app, it's still visible in Android when they open it and then gets re-synced with the web app and appears there again even though it was deleted. I can't reproduce this but wanted to record it somewhere anyway.

They don't use tags and only notice the problem when creating or editing notes in Android.

https://simplenote.desk.com/agent/case/62770

gracie:
Similar issue reported in #2908275-t.

Scenario: Web app is open in Windows and the Android app open on their phone (it does not happen when only one app is open).

What happens: They edit the note, then it syncs back to an old version of the note on both apps. They can see their edits in the history.

roundhill:
Dupe of #243

Issue #243, opened 2014-09-13
Syncing issues with simultaneous edits


beaucollins:
[Google Play Store review](https://play.google.com/store/apps/details?id=com.automattic.simplenote&reviewId=Z3A6QU9xcFRPRnY4NkRpUWY1VjR6clF6YmNHMm9qaHF5NW9wbUMta3c0NkpPTXpiaVlXOWVEcERHdjhkaDI2Y0wtVkpubUJmUnYwVko1SG5qaGtJdWhuczd3):

> Great app, but sync issues. When I edit the same note on Android and web, sometimes the sync is not up to date and I wind up editing an old version. When sync later occurs, I appear to lose text, and I need to go to the web and select an older note version to get my text back. To avoid having to do this, I try to make sure I am editing the latest version when I switch between platforms - based on content alone, as there is no sync timestamp or force sync button.

roundhill:
Fixed via #554 

props @whhone

PR #554, opened 2018-08-26
Stop note bucket on pause

whhone:
Issue: Note is stale when app is resumed. New change on app will
overwrite the server version. There are duplicate issues opened.

- https://github.com/Automattic/simplenote-android/issues/350
- https://github.com/Automattic/simplenote-android/issues/290
- https://github.com/Automattic/simplenote-android/issues/243

Step to reproduce:

1. Open a note and put it in the background.
2. Make change to the same note from web or other client
3. Bring the simplenote android app to the foreground.

Expected: New changes should appear.
Actual:   No change from the server. Editing note on Android will
          overwrite the server version.

This is because the app does not sync note from server on resume.

Changes: Stop bucket on pause so that bucket could be started correctly
onResume.

Also see https://simperium.com/docs/reference/android/#bucket-stop,
> It's recommended to "stop" (typo on the actual web page) buckets in
> activities or fragments in the onPause() method.

Tested: Perform the above repro steps and get the expected behavior.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -343,7 +343,11 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
 
     @Override
     public void onPause() {
+        super.onPause();  // Always call the superclass method first
+
         mNotesBucket.removeListener(this);
+        mNotesBucket.stop();
+
         // Hide soft keyboard if it is showing...
         if (getActivity() != null) {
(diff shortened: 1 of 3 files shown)


roundhill:
Thanks for the simple fix @whhone! I noticed that we also don't stop the buckets in `NotesActivity()` as well, would you mind adding it to `onPause()` there for both the tags and note buckets?

roundhill:
> It's recommended to "stop" (typo on the actual web page) buckets in
> activities or fragments in the onPause() method.

I fixed the typo 👍

whhone:
OK, I will make the update to `NotesActivity` and update this pull request later today or tomorrow.

whhone:
@roundhill, the buckets in `NotesActivity` are now stopped on pause now.

roundhill:
:shipit: Thanks again @whhone, this should prevent some user headaches. I'll get this packaged up and released this week.

PR #557, opened 2018-09-06
Update Simperium to Fix Case-Insensitive UTF Search

### LINKED ISSUE #278: Search is case sensitive for Cyrillic languages
Steps:
1. Create a note with the word Бабушка (Grandmother in Ukranian and Russian).
2. Search for the lower case version in the Android app - бабушка. The search doesn't return the note with the same word but an upper case first letter.

roundhill:
Fixes #278 

**To Test**
* Add some Capitalized UTF text to note, such as `Бабушка`.
* Search for the lower-case version of that word: `бабушка`.
* You should see the match in the search results. Other searches should work as usual.

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -37,7 +37,7 @@ dependencies {
         exclude module: 'recyclerview-v7'
     }
     // Automattic and WordPress dependencies
-    compile 'com.simperium.android:simperium:0.6.9'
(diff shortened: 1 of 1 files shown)


gtsiolis:
Thanks for fixing this @roundhill! I tried this locally and [this](https://github.com/Automattic/simplenote-android/issues/278#issuecomment-218152347) works like a charm now!

PR #559, opened 2018-09-06
Search: Fix within-word search using like query fallback

### LINKED ISSUE #463: Search doesn't support special characters
Steps to reproduce:

1. Create a note with $.
2. Try searching for $.

The Android app doesn't find it but the web app does.

roundhill:
The app uses Full Text Search, which works well for most use cases but doesn't work for searches within words. See #463.

I've fixed this by falling back to using a `LIKE` query if the full text search fails:

screen shot 2018-09-06 at 2 26 20 pm

**BUT**

This may cause some unexpected results for the user, because the full text search does nice stuff such as highlighting the search results and showing it in context, which we can't do for the LIKE query as far as I know:

screen shot 2018-09-06 at 2 26 57 pm

So it may not be worth adding this based on the different experience you get when searching in certain scenarios. But it does help match up the search results with the other Simplenote apps, which all do a similar `LIKE` query.

Fixes #463.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -403,6 +403,13 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
             query.include(new Query.FullTextSnippet(Note.MATCHED_TITLE_INDEX_NAME, Note.TITLE_INDEX_NAME));
             query.include(new Query.FullTextSnippet(Note.MATCHED_CONTENT_INDEX_NAME, Note.CONTENT_PROPERTY));
             query.include(Note.TITLE_INDEX_NAME, Note.CONTENT_PREVIEW_INDEX_NAME);
+
+            // Fall back to a 'LIKE' query if no full text results were found. See issue #463
+            if (query.execute().getCount() == 0) {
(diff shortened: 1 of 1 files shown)


roundhill:
Yeah, I think those unexpected results could be confusing. Do you think we should just leave this as-is? Another option could be to always do a `LIKE` search and figure out how to do the token matching with it.

theck13:
As a user, I would be confused if searching within words worked *sometimes*.  It threw me for a second and I know the code.

> Another option could be to always do a `LIKE` search and figure out how to do the token matching with it.

This would be the best option, but I'm not sure how difficult the token matching would be.

roundhill:
> This would be the best option, but I'm not sure how difficult the token matching would be.

I'll give it a try. If I can't pull it off I'll just close this :)

whhone:
Two questions:
1. Does the LIKE search work for multiple terms?
2. How does the electron app search work? Can we have unified experience on both platforms?

roundhill:
> Does the LIKE search work for multiple terms?

We'd need to split it up to multiple `OR` statements, which doesn't appear to be possible with the current version of Simperium.

> How does the electron app search work? Can we have unified experience on both platforms?

It does an in-memory regex of all of the notes... not really possible here.

I looked into the `LIKE` query and match locations, doesn't seem possible. Going to abandon this one.

Issue #562, opened 2018-09-07
Search highlight doesn't work for last word in note


roundhill:
I think we may have an 'off by 1' error in how we style a search result in the note editor. If the searched word is the last one in the note, it doesn't highlight.

**To Reproduce**
1) Add a note:
```
It's a note!
Wow
```
2) Search for `Wow` and tap on the note you just created . Observe that `Wow` is not highlighted in the note editor.

PR #563, opened 2018-09-07
Search: Scroll to first match in the note editor

roundhill:
When loading a note after searching, the note editor will now attempt to scroll to the first match in the text:

search-scroll

I was hoping this would be less code, but there's a few weird things with getting this to work:
1. You have to get the `layout` of the EditText, which could be null before the first layout pass. So I've added a `addOnGlobalLayoutListener()` which is where the scrolling code lives.
2. We use two layout files for `NoteEditorFragment`, one for small screens and another for large screens. The root scroll view is different in both of these files (`NestedScrollView` and `ScrollView`) so I had to do some goofy casting in order to use `smoothScrollTo`.

I also noticed #562 while testing this.

**To Test**
1. Add a note that has a lot of text, so the note editor needs to scroll.
2. Test searching for various words within the note. They should all be visible in the note editor when loaded.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/MatchOffsetHighlighter.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/MatchOffsetHighlighter.java
@@ -15,6 +15,8 @@ import java.util.Scanner;
 public class MatchOffsetHighlighter implements Runnable {
 
     static public final String CHARSET = "UTF-8";
+    static private final int FIRST_MATCH_LOCATION = 2;
+
     protected static OnMatchListener sListener = new DefaultMatcher();
     private static List<Object> mMatchedSpans = Collections.synchronizedList(new ArrayList<>());
     private SpanFactory mFactory;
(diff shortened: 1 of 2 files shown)


Simplenote/src/main/java/com/automattic/simplenote/utils/MatchOffsetHighlighter.java
```diff
@@ -87,6 +89,24 @@ public static void highlightMatches(Spannable content, String matches, int colum
         }
     }
 
+    // Returns the character location of the first match (3rd index)
```
theck13: Why is the first match the third index?

roundhill: I can see how that'd be confusing. There's 4 digits in a space-separated string for a `match`. The 3rd digit is the one that gets us the start of the match string in the text. I added a comment that _hopefully_ clears that up!

Issue #546, opened 2018-07-10
Searching notes brings up random results


mgalbritton:
User is on a Huawei P10 Lite mobile device using 1.6.1.

In the search bar, they search for a term, but it highlights seemingly irrelevant results. 

Reported in 1266338-zen with screenshots. 

I've tried to reproduce but was not able to.

lizswafford:
User reported that search does work on the desktop app, but still not on the mobile app.

1331012-zen.

roundhill:
If we could get the actual search terms, that would be helpful!

PR #566, opened 2018-09-12
Fix Search Highlighting in Note Editor

roundhill:
Beta testers who have been testing the search improvements have noticed that is quite easy to get #546 to occur. ~~I dug into this and discovered that the search match indices we get from sqlite were counting escape sequences such as `\n` as two characters. This would cause the highlighting to be off depending on how many escape sequences were in the note before the match.~~ The real bug was that we were note getting the correct substring in the `getByteOffset()` method, because we weren't getting the substring using the byte location of the match (which is what SQLite provides us in the FullTextSearch).

Also added fix for #562, which was simply a removal of a `length - 1` check when setting the span. It appears that you can set the span to the length of the string without a zero index and it will still work.

**To Test**
* Create a note that has line breaks and UTF-8 characters in it (like a cyrillic lorem ipsum).
* Return to the notes list and search for the term.
* Tapping the search result should be highlighted correctly.
* Now add a search term to the very end of a note, on its own line.
* Searching for this term should highlight at the very bottom of the note editor.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -340,7 +340,10 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
                     }
 
                     // Get the character location of the first search match
-                    int matchLocation = MatchOffsetHighlighter.getFirstMatchLocation(mMatchOffsets);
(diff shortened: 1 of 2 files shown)


roundhill:
This still isn't working quite right, I'll fix it up here in this branch.

roundhill:
OK, ready for review again. This new solution does do some array copying, so we should watch out for OOM issues. I created a huge note and didn't have any issues using search on a Redmi note 5.

Issue #564, opened 2018-09-07
Fix Inspection Errors/Warnings


roundhill:
Looks like the app has fallen quite behind in code structure/performance/reliability:

<img width="456" alt="screen shot 2018-09-07 at 9 31 39 am" src="https://user-images.githubusercontent.com/789137/45232299-c88ddb80-b283-11e8-95ee-151d0f3777b8.png">

We should clean all that up!

PR #569, opened 2018-09-18
Replace “…” with ellipsis character

jorgecasariego:
Update strings.xml to replace three dots with ellipsis character.

This is a partial code structure cleanup for #564


jorgecasariego:
Oops, I didn't know about this @roundhill 

 I will revert the changes I did in the other files!

PR #570, opened 2018-09-19
Add Tracks Opt-Out to Settings

### LINKED ISSUE #558: Tracks: Opt Out
### Details:
- [x] Add a new boolean field to the`Settings` entity (Simperium backed) to persist if the user opted out from Tracks
- [x] Add a new Switch in the Settings UI

Ref. [iOS Issue](https://github.com/Automattic/simplenote-ios/issues/222)

roundhill:
Adding tracks opt out setting using new `preferences` simperium bucket. 

**To Test**
* Flip the `Share analytics` switch to off in the settings.
* Verify that this method returns `false`: https://github.com/Automattic/simplenote-android/compare/add/558-tracks-opt-out?expand=1#diff-58aa35843c844af9f358fff48efda40bR97
* Switch it back to on, tracks events should resume sending.
* Tap the row with the `learn more` link in settings, it should load `https://automattic.com/cookies` in the browser.

Screenshot:
screen shot 2018-09-19 at 11 31 06 am

Fixes #558

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
@@ -1,14 +1,14 @@
 package com.automattic.simplenote;
 
 
-import android.content.DialogInterface;
-import android.content.SharedPreferences;
-import android.os.AsyncTask;
-import android.support.v7.app.AlertDialog;
 import android.app.Fragment;
+import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.net.Uri;
+import android.os.AsyncTask;
 import android.os.Bundle;
+import android.support.v7.app.AlertDialog;
 import android.support.v7.preference.ListPreference;
 import android.support.v7.preference.Preference;
 import android.support.v7.preference.PreferenceManager;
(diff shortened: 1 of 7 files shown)


Simplenote/src/main/java/com/automattic/simplenote/models/Preferences.java
```diff
@@ -0,0 +1,54 @@
...
+            return getProperties().getInt(ANALYTICS_ENABLED_KEY) > 0;
+        } catch (JSONException e) {
+            e.printStackTrace();
+            return true;
+        }
+    }
+
+    public void setAnalyticsEnabled(boolean enabled) {
+        try {
+            getProperties().put(ANALYTICS_ENABLED_KEY, enabled ? 1 : 0);
```
roundhill: This should be a normal boolean instead of the int > 0 check, but iOS seems to write a boolean to simperium as an integer: https://github.com/Automattic/simplenote-ios/pull/228#issuecomment-422970461

Depending on if that gets fixed, I'll do a quick follow-up PR to set this to a regular ol' boolean.

Simplenote/src/main/java/com/automattic/simplenote/models/Preferences.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final String BUCKET_NAME = "preferences";
+    private static final String ANALYTICS_ENABLED_KEY = "analytics_enabled";
+
+    private Preferences(String key, JSONObject properties) {
+        super(key, properties);
+    }
+
+    public boolean getAnalyticsEnabled() {
+        try {
+            return getProperties().getInt(ANALYTICS_ENABLED_KEY) > 0;
```
roundhill: Note: This should be a normal boolean instead of the int > 0 check, but iOS seems to write a boolean to simperium as an integer: Automattic/simplenote-ios#228 (comment)

Depending on if that gets fixed, I'll do a quick follow-up PR to set this to a regular ol' boolean.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -1,39 +1,49 @@
...
  * A simple {@link Fragment} subclass.
  */
-public class PreferencesFragment extends PreferenceFragmentCompat implements User.StatusChangeListener, Simperium.OnUserCreatedListener {
+public class PreferencesFragment extends PreferenceFragmentCompat implements User.StatusChangeListener,
+        Simperium.OnUserCreatedListener {
 
     private static final String WEB_APP_URL = "https://app.simplenote.com";
 
+    private Bucket<Preferences> mPreferencesBucket;
+    private SwitchPreferenceCompat mAnalyticsSwitch;
```
maheshgaya: `SwitchPreferenceCompat` will go away with the update to Android 27. We can switch it to the Android support version.

See SwitchPreferenceCompat (2017-08-04) in [Changelog](https://github.com/Gericop/Android-Support-Preference-V7-Fix/blob/b628c9cecf175ea24789a9877613dff14bfb40cd/CHANGELOG.md)

roundhill:
> I set breakpoints on the return statements in the `analyticsIsEnabled` method while testing. Regardless of whether the __Share analytics__ setting was enabled or disabled, a `BucketObjectMissingException` was caUght and `true` was returned.

🤪Well I guess I should test this a little more! Thanks.

PR #572, opened 2018-09-20
Show keyboard after returning to a note

### LINKED ISSUE #534: Keyboard doesn't stay open when leaving Simplenote and returning to a note
A user wrote in about this issue which I can reproduce. I am not sure if there is a way to keep the keyboard open in Android but here is how you can reproduce it:

1. Open Simplenote and tap into a note
2. Tap the home button and open any other app
3. Return to Simplenote

The cursor will still be in the text but the keyboard is not open automatically.

jorgecasariego:
Fixes #534 - Keyboard doesn't stay open when leaving Simplenote and returning to a note

To Test

**Note without focus**
- Open a note
- Notice that the keyboard is hidden
- Tap the home button and open any other app
- Return to Simplenote and notice that the keyboard is hidden

**Note with focus**
- Open a note
- Tap into the Note to show the keyboard
- Tap the home button and open any other app
- Return to Simplenote and notice that the keyboard is showing

p.s. 
I'm noticing that when I test with a **Nexus 5X (API 28)** the note opens with focus:

drawing

It's not the case with **Pixel (API 23)** or with **Huawei Y7 (API 26)**
drawing


roundhill:
Thanks @jorgecasariego! I noticed that I can't seem to open a note un-focused. It always has the cursor in the note editor for me. Do you not see the same on your Nexus 5?

jorgecasariego:
@roundhill In my case the first time I open a note is un-focused. Then if I tap on the note the keyboard show up. 

I made a video to show you how is working in a real device (Huawei Y7 API 26 )

jorgecasariego:
> The `showSoftKeyboard` changes look good to me.
> 
> There is a merge conflict though. Once that's resolved, we'll be ready to merge and close this pull request.

Thanks for the review @theck13 . After some issues with gradle.properties-example file and travis.yml file I could do the merge :)

Issue #581, opened 2018-10-24
Feature Proposal: Add an 'Auto' option to the theme selection options


iskandergaba:
I use Simplenote at different times of the day. While I prefer the Light theme during the daylight, I also like the dark theme during the night. A very common use case is the user, i.e. me, writing some notes while on the bed just before sleeping. In this case, the light theme kind of hurts the eyes and the dark one is much more eye friendly.

I propose adding a third option, `Auto`, to the theme selection menu. If the user selects this option, the application will decide, upon start, on what theme to use depending on the time of day. Just like Twitter's application auto setting of night mode. I believe this is better than me switching back and forth between the themes.

I have already started exploring the code and I am planning to implement this myself as it seems to be fairly straightforward. However, I wanted to post this so that I can have an idea about what others think about this feature and whether it is worth it or not.

robertoissc:
I think it's a good idea, I actually use that feature on Twitter...
I wonder how it decides to switch modes...
Twitter states that:
> night mode can be set to commence at sunset and turn off at sunrise

iskandergaba:
> I think it's a good idea, I actually use that feature on Twitter...
> I wonder how it decides to switch modes...
> Twitter states that:
> 
> > night mode can be set to commence at sunset and turn off at sunrise

That is the idea @robertoissc. I want dark theme to commence at sunset and turn off at sunrise if `Auto` option is selected.

The way I want to do this is by using Android's `DayNight` Theme that was added to `AppCompat` with the realease of Support Library v23.2.0. I want to refactor the existing `Light` and `Dark` themes into a single `DayNight` theme and use built-in functionality to let it decide if it has to display `Dark` or `Light` theme when `Auto` is selected. Of course, all colour and theme values will be preserved and will remain the same. This is just a refactor, not a visual theme change.

## Benefits:
- No longer need to maintain two separate themes and therefore better development experience.
- The Day/Night detection is taken care of by the framework.

## Resources:
- https://medium.com/androiddevelopers/appcompat-v23-2-daynight-d10f90c83e94
- https://blog.iamsuleiman.com/daynight-theme-android-tutorial-example
- http://www.androidtutorialshub.com/android-daynight-theme-example-using-appcompat-support-library

iskandergaba:
Implemented and merged. Please close this issue.

PR #584, opened 2018-10-25
Added the option `Auto` to the themes to use the Dark theme when it is night and the Light one when it is daytime

iskandergaba:
This is the implementation of feature #581 

Visual appearance of the themes are NOT tempered with. They are still the same as before. I only refactored their values for the `DayNight` theme to work out. No dramatic change of code is involved and the application can now determine when it is night and when it is daylight and according to that chooses the right theme (if `Auto` theme option is selected)

**P.S:** Although I've been developing for Android for a while, I am a little bit excited as this is my first *meaningful* open source contribution. I hope this little feature does not get shut down and that it actually makes it to release. Fingers crossed xD.

### Some References:
- https://medium.com/androiddevelopers/appcompat-v23-2-daynight-d10f90c83e94
- https://blog.iamsuleiman.com/daynight-theme-android-tutorial-example
- http://www.androidtutorialshub.com/android-daynight-theme-example-using-appcompat-support-library

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/ThemeUtils.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/ThemeUtils.java
@@ -30,7 +30,7 @@ public class ThemeUtils {
             if (data.getAuthority().equals(PREFERENCES_URI_AUTHORITY)) {
                 List<String> segments = data.getPathSegments();
 
-                // check if we have reched /preferences/theme
+                // check if we have reached /preferences/theme
                 if (segments.size() > 0 && segments.get(0).equals(URI_SEGMENT_THEME)) {
 
                     // activate the theme preference
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/ThemeUtils.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/ThemeUtils.java
@@ -7,8 +7,7 @@ import android.content.SharedPreferences;
 import android.graphics.Point;
 import android.net.Uri;
 import android.preference.PreferenceManager;
-
-import com.automattic.simplenote.R;
+import android.support.v7.app.AppCompatDelegate;
 
 import java.util.List;
 
@@ -16,8 +15,9 @@ public class ThemeUtils {
 
     // theme constants
     public static final int THEME_LIGHT = 0;
-    @SuppressWarnings("unused")
     public static final int THEME_DARK = 1;
+    @SuppressWarnings("unused")
+    public static final int THEME_AUTO = 2;
     static public final String PREFERENCES_URI_AUTHORITY = "preferences";
     static public final String URI_SEGMENT_THEME = "theme";
     public static String THEME_CHANGED_EXTRA = "themeChanged";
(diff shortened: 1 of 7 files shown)


iskandergaba:
Oh, one thing. I did not add translation for `Auto` word in other languages (because I only know three). While I could use Google Translate or something, since it is just one word, I thought that it'd be better to leave it to you guys to decide.

### Some Screenshots:

roundhill:
Cool! I'm wondering what we can put as the text instead of `Auto` so that it makes more sense to a user seeing it for the first time. Something like `Automatic Daylight Mode`?

roundhill:
> I did not add translation for `Auto` word in other languages

That's fine, you only need to add the English version. We have community translators that handle the rest.

iskandergaba:
> > I did not add translation for `Auto` word in other languages
> 
> That's fine, you only need to add the English version. We have community translators that handle the rest.

Okay, great then.

> Cool! I'm wondering what we can put as the text instead of `Auto` so that it makes more sense to a user seeing it for the first time. Something like `Automatic Daylight Mode`?

Yes we could name it like that. Or `Dark at Night Only`?

iskandergaba:
@roundhill  Which one looks better for you?

iskandergaba:
As expected, there was nothing wrong with the code. The build before the last one failed just because of failure downloading a package.

roundhill:
> Dark at Night Only

👍 let's go for that. I think We should have it be lower-case after the first work per Android guidelines: `Dark at night only`. 

Also I can't get this to work at all on my Xiamoi Note 5. It could just be their weird custom UI preventing it, but I do see this working well in the Simulator.

iskandergaba:
> +1 let's go for that. I think We should have it be lower-case after the first work per Android guidelines: `Dark at night only`.

@roundhill Change implemented.

> Also I can't get this to work at all on my Xiamoi Note 5. It could just be their weird custom UI preventing it, but I do see this working well in the Simulator.

While I do not know about Xiamoi Note 5, I am glad it works. Also, the application takes the sunset value as a fixed 10pm and the sunrise value as a fixed 6am. This is because it is not granted the location permissions to be able to determine those values accurately. This behaviour is also observed in Twitter application when Location permission is not granted (I made sure myself). Take a look [here](https://medium.com/androiddevelopers/appcompat-v23-2-daynight-d10f90c83e94):

> When using MODE_NIGHT_AUTO, we need to use the device location to be able to calculate sensible times to switch between day or night. This might be sound scary, but don’t fear. If your app already has location permissions granted, AppCompat will try and grab the last known location from LocationManager and use those to calculate the local sunrise and sunset.

> If your app has not been granted those permissions, or there simply isn’t a last known location, then we use some fixed values. These are currently 6 am for sunrise and 10 pm for sunset.

Twitter takes a silent approach and judges that those fixed values are good enough. I personally had the location service turned off for Twitter app and never noticed the difference until now as I am working on something similar. As for Simplenote, I see that you do not require any permissions from the users and probably would like to keep it that way.

Should I keep the 10pm-6am behaviour as it is? Or do you think it is better to prompt user for location services permissions when the `Dark at night only` theme is selected for the first time, where we can indicate why it is needed and the alternative behaviour if the permission is not granted?

roundhill:
> Should I keep the 10pm-6am behaviour as it is?

Aha! I had set it to 9pm which wasn't working, obviously. Setting it to 10 worked on my Xiaomi now.

> Or do you think it is better to prompt user for location services permissions when the `Dark at night only` theme is selected for the first time, where we can indicate why it is needed and the alternative behaviour if the permission is not granted?

We definitely like that we don't ask for too many permissions, but I think it'd be cool if we could have theme change right at the appropriate sunset time. Want to give it a try?

iskandergaba:
> We definitely like that we don't ask for too many permissions, but I think it'd be cool if we could have theme change right at the appropriate sunset time. Want to give it a try?

Alright! I will give it a go and see how it goes. Although the only way to know precise sunset and sunrise times is to know the location, I will make permissions request as minimal as possible.

iskandergaba:
Checking permissions when they are not granted.


Letting the user know about 10pm-6am behaviour if the user denies the permission

iskandergaba:
Works fine. Ready for merge.

PR #589, opened 2018-11-03
Fix resource validation errors

nthcircle:

- Refactor a xml resource name

Test Steps:

1. Build and run app. It should build as normal.


nthcircle:
Edit: I reverted the removal of passcode unlock manifest entries. These are indeed needed/used, even if inspection reports are unreferenced. I'll followup with an update to clean up (suppress) some of these false warnings.

nthcircle:
I reran inspection and that other unresolved reference (for xml view id) is still being reported. I'll just close this PR since it doesn't actually resolve those inspection warnings/errors.

PR #590, opened 2018-11-03
Fix leaky async used by Wear create note.

nthcircle:

- Fix potential Context leak in the create note AsyncTask

Test Steps:

1. Create a new note from Wear app(?).

Note: This appears to used by Wear only. I was unable to test and validate this fix.


roundhill:
:shipit:

PR #592, opened 2018-11-03
Fix leaky async for save note task.

nthcircle:

- Fix potential Context leak in the save note AsyncTask

Test Steps:

1. Open up a note and start typing. Note will auto-save after a few seconds.


roundhill:
:shipit: thanks @nthcircle!

Issue #602, opened 2018-12-05
Privacy: Recent app screen showing note contents, even when passcode is enabled


kriskorn:
#### Steps to reproduce
1. Open Simplenote app
2. Write a note and minimize the app
3. Open the recent apps on your mobile device

#### What I expected

To see a Simplenote app splash screen or a logo.

#### What happened instead

Saw all of the contents of my note.

#### Simplenote version

1.6.3

#### OS version

8.1.0

#### Screenshot / Video

![android-recent-app-screen](https://user-images.githubusercontent.com/12064241/49517201-4160fb80-f8a4-11e8-8fcf-15408933cf2e.jpg)

#1620989-zen

iskandergaba:
Good suggestion indeed. I will look into it soon.

PR #604, opened 2018-12-08
Sensitive content is now hidden from the recent apps when app lock is enabled

iskandergaba:
- Added the requested feature #602. Sensitive content (note list, tag list, note content, and settings) are now hidden from recent apps when app lock is enabled.
- Some code clean up along the way.

**App lock enabled:** Non-snesitive content screen

screenshot_1

**App lock enabled:** Snesitive content screen

screenshot_2

@roundhill Please check this out and let me know.


iskandergaba:
@roundhill So basically I am specifying the flag `FLAG_Secure` **in the activities containing sensitive information when the application lock is enabled**. This is a common practice that banking applications and privacy centered applications like Firefox Focus follow to hide the thumbnail of the application from `Recent Apps` menu. The problem with this approach is that it disables screenshots from being taken on that activity. So basically, when the application lock is enabled, thumbnail from `Recent Apps` are hidden and no screenshot can be taken on those sensitive information activities. When the application lock is disabled, the behavior is normal.

I tried to add the flag for instance in `OnPause()` (To hide thumbnail in `Recent Apps`) and remove it in `OnResume()` (to enable screenshot functionality) method for instance to get the exact desired behavior (remove thumbnail but keep screenshot functionality) without sacrificing taking screenshots, which worked a bit but inconsistently I am afraid. The lifecycle calls are not guaranteed to be made synchronously and therefore rarely this worked. I played with a lot of other combinations lifecycle methods but in vain. So my question here is: Is it fine to sacrifice the functionality of taking screenshots on sensitive information activities when the application is locked or not? If so, I will do a little tweak to this approach and will be good to go. Otherwise, I am afraid there is not any other way of doing this unless we cover the layout with another layout before the activity is paused (which is not so cool I think).

Let me know so that I can proceed.

roundhill:
> Is it fine to sacrifice the functionality of taking screenshots on sensitive information activities when the application is locked or not?

Sounds OK to me. Would we be able to show a Toast or something if a screenshot is attempted that it is disabled because the passcode lock is enabled?

iskandergaba:
@roundhill I am not sure how to listen to when the user attempts to take a screenshot in a clean and effective way so that I show a toast when such an event occurs.

An easier indication however would be to add description to `Turn PIN Lock on` setting that describes this behavior briefly.

What do you think?

roundhill:
> An easier indication however would be to add description to `Turn PIN Lock on` setting that describes this behavior briefly.

Sure, let's go for that 👍

iskandergaba:
> Sure, let's go for that 👍
Done.

iskandergaba:
> Sure, let's go for that 👍

Done.

iskandergaba:
> Sure, let's go for that 👍

Done.

screenshot

roundhill:
Great. How about shortening it to `App screenshots will be disabled while this option is on.`?

iskandergaba:
> Great. How about shortening it to `App screenshots will be disabled while this option is on.`?

Sure. Here you go.

screenshots

roundhill:
Sorry for the delay in reviewing this @iskandergaba! 

I was a little confused at first because this wasn't working after I enabled the pin lock in `NotesActivity`. I think it's because it isn't calling `onCreate` again because the activity already exists. It works great after I restarted the app though 👍Any workarounds for that?

roundhill:
> Let me know how it tests on yours.

It still isn't working for me, unfortunately.

1. Disable PIN lock.
2. Enable PIN lock, and return to notes activity.
3. Press multitask button, observe notes list is still visible.

iskandergaba:
@roundhill I am not sure what is going wrong. I tested many times using my Galaxy S6 Edge and on emulated Pixel 2 Device and it worked like a charm :/

roundhill:
> I tested many times using my Galaxy S6 Edge and on emulated Pixel 2 Device and it worked like a charm :/

OK, I'll check again!

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
```diff
@@ -41,4 +43,16 @@ public void onCreate(Bundle savedInstanceState) {
...
+    @Override
+    protected void onResume() {
+        super.onResume();
+
+        // Disable screenshots if app is lock is on
+        if (AppLockManager.getInstance().getAppLock().isPasswordLocked()) {
+            getWindow().addFlags(WindowManager.LayoutParams.FLAG_SECURE);
+        } else {
+            getWindow().clearFlags(WindowManager.LayoutParams.FLAG_SECURE);
+        }
```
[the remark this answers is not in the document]
iskandergaba: Got it. Will do it tonight.

iskandergaba:
@theck13 I implemented the changes you requested.

Simplenote/src/main/java/com/automattic/simplenote/utils/DisplayUtils.java
```diff
@@ -72,8 +75,15 @@ public static int getChecklistIconSize(Context context) {
         if (context == null) {
             return 18;
         }
-
         return DisplayUtils.dpToPx(context, PrefUtils.getFontSize(context)) + DisplayUtils.dpToPx(context, 6);
     }
-}
 
+  // Disable screenshots if app is lock is on
```
theck13:
```suggestion
  // Disable screenshots if app PIN lock is on
```

I think this comment has a typographical error.  Should it be "// Disable screenshots if app PIN lock is on" instead?

iskandergaba: Yes you are right. Sorry it slipped my mind.

Issue #598, opened 2018-11-26
Menu misaligned with status bar


jakebornheimer:
#### Steps to reproduce
1. Use Essential Phone (or maybe phone with notch)
2. Install new version of Simplenote
3. Open the hamburger button menu

#### What I expected

The menu should open _below_ the space that the status bar occupies.

#### What happened instead

As in the screenshot below, the menu opens _behind_ the status bar. The menu options are still accessible, but the "All Notes" menu item is partially obscured by the status bar, making it difficult to tap.

#### Simplenote version

1.1.6

#### Hardware

Phone: Essential Phone PH-1 (ver. 1.8)
(This phone has a central front-camera notch - may be related to this behaviour.)

#### Android version

9

#### Screenshot

![simplenotebug](https://user-images.githubusercontent.com/45367248/49031465-00683900-f178-11e8-9a36-b17c8b6edcf3.png)

roundhill:
Thanks for the report, I can confirm this on an emulator with a notch:

<img width="264" alt="screen shot 2018-11-26 at 10 36 11 am" src="https://user-images.githubusercontent.com/789137/49034542-20dbc780-f167-11e8-8465-88e2d51fe41c.png">

PR #606, opened 2018-12-16
Fixing menu misaligned (#598)

hrossi:
This commit fix the issue #598 

You can see prints of device with and without notch.

Print 1 - Device with notch
image

Print 2 - Device without notch
image

### DIFF
--- a/Simplenote/src/main/res/layout/activity_notes.xml
+++ b/Simplenote/src/main/res/layout/activity_notes.xml
@@ -33,6 +33,7 @@
         android:id="@+id/navigation_view"
         android:layout_width="wrap_content"
         android:layout_height="match_parent"
+        android:fitsSystemWindows="false"
         android:layout_gravity="start">
 
         <LinearLayout


roundhill:
Thanks @hrossi! Can we preserve the menu opening underneath the system bar? I think we can use https://developer.android.com/reference/androidx/core/view/DisplayCutoutCompat to get the inset values, and then change the top padding of the app drawer to accommodate for the notch.

hrossi:
Hi @roundhill.

This class is part of the androidx library and the project isn't using it yet.

I'll take a look if we have any alternative

roundhill:
Any luck @hrossi? Surely there must be an appcompat way of getting that padding value 🤔

hrossi:
Hi @roundhill, sorry for the delay. I was travelling on vacation in the last weeks.

I hope have some luck in my studies today to fix it.

hrossi:
@roundhill I'm having a lot of work these days and I am not having enough time to search about it.

I'll close the pr and as soon as I have more free time I'll dedicate myself into it.

Issue #616, opened 2019-01-06
Manual switch from Dark at Night Only to Day only updates text area in Markdown preview


edequalsawesome:
#### Steps to reproduce
1. Activate Dark at Night Only
2. Switch to Day mode
3. Preview a note written with Markdown

#### What I expected
Entire view should be in the lighter Day mode

#### What happened instead
Text view displayed in Day mode, the rest of the interface remained in Dark mode

#### Simplenote version
1.6.7

#### OS version
Android 9.0 (Android One) running on a Xiaomi Mi A2

#### Screenshot / Video
![Preview of Markdown](http://cld.wthms.co/tueIyi+)

iskandergaba:
Hello, I noticed this as well. I believe this and #615 are both side effects of #580. Before this fix was introduced, things rendered just fine. @roundhill @robertoissc

iskandergaba:
Thank you for reporting the issues in #615 and #616. I have located the problem and it appears to be a simple fix. I will push a fix for this in few minutes.

iskandergaba:
@roundhill, Check #617

Issue #615, opened 2019-01-06
"Null" string appearing in Markdown preview when Dark at Night Only is enabled


edequalsawesome:
#### Steps to reproduce
1. Create a new note in Markdown
2. Click the Preview button

#### What I expected
Preview will only show text from the note with Markdown formatting applied

#### What happened instead
Markdown formatting was applied, but there is a "null" string at the top of the note -- turning off Dark at Night Only resolves the issue, though

#### Simplenote version
1.6.7

#### OS
Android 9.0 (Android One) running on a Xiaomi Mi A2

#### Screenshot / Video
![Note Text](http://cld.wthms.co/GlVNPJ+)
![Note Preview](http://cld.wthms.co/NVQhGG+)

iskandergaba:
Hello, I noticed this as well. I believe this and #616 are both side effects of #580. Before this fix was introduced, things rendered just fine. @roundhill @robertoissc

iskandergaba:
Thank you for reporting the issues in #615 and #616. I have located the problem and it appears to be a simple fix. I will push a fix for this in few minutes.

PR #617, opened 2019-01-07
Bugs #615 and #616 Fixed

iskandergaba:
There were two problems:
1- There was no case to what `css` file `NoteMarkdownFragment` should choose when `Dark at night only` theme is selected. This led to weird rendering of the Markdown preview including the `null` appearing at the top.
2- `isLightTheme()` method in `ThemeUtils` only checked the value in `SharedPreferences` and not the actual applied theme. If the value is `THEME_AUTO`, one cannot tell whether dark or light theme is chosen.

I fixed both problems by simplifying `isLightTheme()` method to only check the actual UI state (night mode or not, regardless of the value in `SharedPreferences`) which is simple and sufficient. I then swapped all the code instances that checks what theme is chosen with `ThemeUtils.isLightTheme(context)`. This ensures better and cleaner code as well as fixing the issue.

Check it out @roundhill. I believe this needs to be pushed to production asap.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
@@ -5,6 +5,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.res.Configuration;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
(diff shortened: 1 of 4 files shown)


roundhill:
Tested well for me, thanks @iskandergaba! :shipit:

PR #619, opened 2019-01-09
Bugfix. Dark theme is now correctly applied in bottom sheets

### LINKED ISSUE #613: Open in Browser button and others do not appear correctly in Dark Mode
#### Steps to reproduce
1. Turn on Dark Mode
2. Exit and reopen Simplenote
3. Open a note and tap on a link

#### What I expected
Two things:
1. To see the Open in Browser button
2. For the buttons and top bar to match the colors for the app

#### What happened instead
The first time you tap any link after opening Simplenote in dark mode, you do not get the Open in Browser button and the coloring is wrong. If you go back to the notes list and tap any link, it will then work.

#### Simplenote version
1.6.7

#### OS version
Android 8.0.0

#### Screenshot / Video
First tap:
incorrect styling

After trying again:
48971461_289016515110941_615896304496672768_n

iskandergaba:
However, this does not fix #613 and the `RecyclerView` showing the sharing app buttons displays incorrect text colors THE FIRST TIME ONLY, which is caused by the same issue in #613. For some reason, when the `NoteEditorActivity` is opened the first time in since opening the app, the app suddenly switches from night mode to day mode. I spent the whole day looking for the cause with no success.

The theming however is fixed and I believe can be safely merged. The slight bug remaining is a result of #613. After a lot of testing, I am confident that fixing #613 will also fix it here. But that is another issue for another day/night.

Here are the results in dark theme:

**Info Bottom Sheet**

**Share Bottom Sheet**

**History Bottom Sheet**

@roundhill The pull request comes also with many minor code quality improvements.
Cheers!

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/InfoBottomSheetDialog.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/InfoBottomSheetDialog.java
@@ -10,6 +10,7 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.widget.CompoundButton;
 import android.widget.ImageButton;
+import android.widget.LinearLayout;
 import android.widget.TextView;
 
 import com.automattic.simplenote.models.Note;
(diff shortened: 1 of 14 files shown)


roundhill:
Can the background color of the sheets match the background color we are using in the other areas of the app? I think it's `attr/mainBackgroundColor`

iskandergaba:
@roundhill The requested changes are implemented. Check screenshots below.

### Dark
dark

### Light
light

roundhill:
Thanks @iskandergaba! :shipit:

PR #622, opened 2019-01-18
Add support for nested checklists

roundhill:
Depends on #608 

Markdown checklists can be nested, so I've added support for that:

screen shot 2019-01-18 at 9 54 30 am

**To Test**
* Create a checklist at any position on a line. It should insert at the proper position.
* Pressing enter should 'autobullet' a new checklist item on the next line at the same position.
* Selecting multiple lines and/or one line of text and then inserting/removing a checklist should still work as expected.


Simplenote/src/androidTest/java/com/automattic/simplenote/utils/ChecklistsTest.java
```diff
@@ -23,9 +23,9 @@
 public class ChecklistsTest {
     @Test
     public void testRegexMatching() {
```
theck13: What do you think about leaving the `testRegexMatching` unmodified and creating a new method for testing regular expression matching for nested checkboxes?  That way we will know the feature before these changes still works.

roundhill: Good idea, did that in https://github.com/Automattic/simplenote-android/pull/622/commits/56fd38ef5c555790790893e3ccdc7bee7b23b723

PR #624, opened 2019-01-22
New Tracks Event for Checklist Button

roundhill:
Adding a new tracks event for when the checklist button is tapped on the toolbar.

**To Test**
* Ensure analytics is enabled in the app settings.
* Open a note and tap on the checklist button.
* Confirm the event shows up in the tracks live event view. The event is `spandroid_editor_checklist_inserted`


roundhill:
> should we also track when they create a checklist manually by typing `- [ ]` or `- [x]`?

Thanks @theck13! I think we're good without that, and the other apps don't track that either.

PR #628, opened 2019-01-30
Update Markdown Parser to Support Checklists

roundhill:
This PR fixes the issue described in #608 where checklists would not render properly in the Markdown preview:

screen shot 2019-01-29 at 4 06 39 pm

I achieved this by replacing the `AndDown` library we were previously using to parse markdown with [Flexmark](https://github.com/vsch/flexmark-java), which supports many more markdown formatting options, including task lists.

Unfortunately it doesn't seem to support rendering nested checklists, because indented text gets placed in code blocks:

screen shot 2019-01-29 at 4 10 41 pm

**To Test**
* Create a Markdown formatted note that includes checklists. They should render in the markdown preview as disabled checkboxes.
* Ensure that the Markdown preview still works as it did previously (tables, quotes, etc).
* HTML added to a note should be escaped in the markdown preview (like `<input type="text" />`)
* Javascript `<script>` tags should also be escaped.

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -112,8 +112,10 @@ android {
 
 	lintOptions {
 	    checkReleaseBuilds false
+        
         // Flexmark library includes some Java packages that aren't in Android
         disable 'InvalidPackage'
+        disable 'DuplicatePlatformClasses'
 	}
 
     useLibrary 'org.apache.http.legacy'
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -108,9 +108,8 @@ android {
     }
 
     packagingOptions {
-        exclude 'META-INF/LICENSE-LGPL-3.txt'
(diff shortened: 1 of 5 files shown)


Simplenote/build.gradle
```diff
@@ -103,6 +103,13 @@ android {
         }
     }
 
+    packagingOptions {
+        exclude 'META-INF/LICENSE-LGPL-3.txt'
+        exclude 'META-INF/DEPENDENCIES'
+        exclude 'META-INF/LICENSE-LGPL-2.1.txt'
+        exclude 'META-INF/LICENSE-W3C-TEST'
```
roundhill: There seemed to be a conflict with Flexmark and another library with these files. Adding these exclusions fixed compiling for me.

gemmagarner:
This issue has been raised [on Twitter](https://twitter.com/RisingAboveDark/status/1093242510959407106).

belcherj:
Closing for now.

PR #633, opened 2019-02-01
Fix the checkbox keyboard issue

0nko:
Prevents keyboard from popping up when a checkbox is clicked.


roundhill:
We used inspiration from this in #629. Thanks @0nko!

PR #634, opened 2019-02-01
Avoid setting the selection to negative index

### LINKED ISSUE #631: Checking a checkbox without first activating the keyboard exits to the note list
#### Steps to reproduce
1. Open an existing note with a checkbox list
2. Do not touch the text input area (in order to avoid activating the keyboard) 
3. Tap a checkbox directly (in order to check it) 

#### What I expected

The checkbox becomes checked and the keyboard does not activate. 

#### What happened instead

The screen flashes black, then dark grey-blue (regardless of theme light or dark setting) then returns to the note list. The checkbox is not checked.

#### Simplenote version

1.6.8

#### OS version

Both Android 9.0.0.159 and android 8.0.0.159

0nko:
Fixes #631.


roundhill:
Thanks @0nko! :shipit:

PR #640, opened 2019-02-07
Fix SimplenoteMovementMethod from clearing the selection

roundhill:
I noticed that the cursor would jump to the start when you click around the editor:

feb-07-2019 09-03-50

Tracked it down to some code that was left over from when I subclassed `SimplenoteMovementMethod` from `LinkMovementMethod`. Since we are subclassing `ArrowKeyMovementMethod`, we shouldn't be clearing the selection after a click.

**To Test**
* Change the cursor position in the editor, it should behave normally. Clicking checklists and links should also still work as expected.


roundhill:
Thanks @theck13!

PR #644, opened 2019-02-21
Use Status Bar Height as Padding Instead of Static Value

### LINKED ISSUE #598: Menu misaligned with status bar

robertoissc:
Fixes #598 by requesting status bar height when configuring drawer and using this value as top padding for DrawerView instead of an static value.

**To Test**
- Open menu in devices with different status bar sizes like devices with notch and without it.


robertoissc:
Hey @theck13!
I'm sorry. Should I work on the changes you suggest or close this PR and look for a different solution?

theck13:
If we can fix the device rotation bug mentioned [above](https://github.com/Automattic/simplenote-android/pull/644#pullrequestreview-233302593), using this type of solution will be better than what we have now.  If we can't fix that rotation bug, we might want to close this until we have a better solution.  If we can figure out how to use [`OnApplyWindowInsetsListener`](https://developer.android.com/reference/android/support/v4/view/OnApplyWindowInsetsListener), that would be the best.

robertoissc:
Thanks for the information @theck13, I'll work on this!

robertoissc:
Hey @theck13! I've been trying to implement `OnApplyWindowInsetsListener` as you suggested. I tried different approaches but I think this is the best one, it sets the right padding to the drawer and it automatically adjusts it when rotating the device.

Also, I've added the `clipChildren `and `clipToPadding `attributes to the `drawerView` so items can go below the status bar, however, I'm noticing that list items are disappearing earlier when scrolling the drawer, as soon as the item's bottom gets to the status bar's bottom they're gone...

I've been trying to solve this without success. Any clues?

Simplenote/src/main/res/layout/activity_notes.xml
```diff
@@ -40,7 +40,8 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:orientation="vertical"
-            android:paddingTop="@dimen/nav_drawer_navigation_padding_top">
```
theck13: Now that `nav_drawer_navigation_padding_top` is no longer being used, we can remove it from `values/dimens.xml` and `values-v21/dimens.xml`.  Since that was the only resource in `values-v21/dimens.xml`, we can remove that file entirely.

PR #645, opened 2019-02-21
Checklist Cursor and Keyboard Adjustments

### LINKED ISSUE #637: [BUG] UPDATE - New Checklist Feature: After first touch of text, SN jumps to different position when a box is ticked
#### Steps to reproduce
1. Edit text in note
2. Click check box
3. SimpleNote jumps to various positions in text and Keyboard pops up everytime I tick a box

#### What I expected
No jumping in text and no popping up of keyboard, if only check box is ticked. 
The issue does not occur, when note was opened until touching text. 

#### What happened instead
First edit of text leads to jumping to different positions in text and popping up of keyboard when ticking boxes. only closing of note and opening again leads to not popping up of keyboard. 

#### Simplenote version
1.7.0

#### OS version
7.0, 8.0, 9.0 (3 devices)

robertoissc:
Fixes #637 
- Hide cursor when toggling a checkbox until the selection position can be restored to prevent the jumping cursor effect.
- Clear editor focus when closing soft keyboard, this prevent the keyboard from showing up again when the user toggles another item.

**To Test**
- Open checklist note and toggle some items.
- Edit text in checklist note and toggle some items, the cursor should stay where it is.
- Close soft keyboard and toggle some items, the keyboard shouldn't show up.
- Edit text in checklist note, keyboard should show up again.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
@@ -61,6 +61,8 @@ public class SimplenoteEditText extends AppCompatEditText {
 
     // Updates the ImageSpan drawable to the new checked state
     public void toggleCheckbox(final CheckableSpan checkableSpan) {
+        setCursorVisible(false);
+
         final Editable editable = getText();
 
         final int checkboxStart = editable.getSpanStart(checkableSpan);
@@ -92,6 +94,7 @@ public class SimplenoteEditText extends AppCompatEditText {
                             && selectionStart <= editable.length()
                             && selectionEnd <= editable.length()) {
                         setSelection(selectionStart, selectionEnd);
+                        setCursorVisible(true);
                     }
 
                     if (mOnCheckboxToggledListener != null) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
@@ -1,6 +1,7 @@
 package com.automattic.simplenote.widgets;
 
 import android.content.Context;
+import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.os.Handler;
 import android.support.v7.widget.AppCompatEditText;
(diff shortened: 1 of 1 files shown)


Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
```diff
@@ -92,6 +103,7 @@ public void run() {
                             && selectionStart <= editable.length()
                             && selectionEnd <= editable.length()) {
```
theck13:
```suggestion
                            && selectionEnd <= editable.length() && hasFocus()) {
```

During my testing, these conditions are always `true`, which means the `setCursorVisible(true);` statement is always called.  That results in the cursor appearing randomly when it shouldn't.  Follow the steps below to reproduce the random cursor appearance.

1. Open any note with checklist.
2. Notice cursor is not shown.
2. Tap only checkbox to toggle checked/unchecked.
3. Repeat Step 2 indefinitely.
4. Notice cursor is shown/hidden before first character in note randomly.

The random cursor appearance can reproduced after selecting anywhere in the note as well.  Follow the steps below to reproduce that case.

1. Open any note with checklist.
2. Tap anywhere in the note other than a checkbox.
3. Notice cursor and soft keyboard are shown.
4. Tap back button to dismiss soft keyboard.
5. Notice cursor is not shown.
6. Tap only checkbox to toggle checked/unchecked.
7. Repeat Step 6 indefinitely.
8. Notice cursor is shown/hidden before first character in note randomly.

In order to stop the random cursor appearance, we could add the `&& hasFocus()` condition to this `if` statement.  Consequently, that would also cause the `setCursorVisible(true);` statement to be called only when the view has focus.  That means if the back button is pressed or a checkbox is toggled, the cursor will no longer be shown.  To fix that, we can override `onFocusChanged` in `SimplenoteEditText` and call `setCursorVisible(true);` if the view has focus like this.

```java
@Override
protected void onFocusChanged(boolean focused, int direction, Rect previouslyFocusedRect) {
    if (focused) {
        setCursorVisible(true);
    }

    super.onFocusChanged(focused, direction, previouslyFocusedRect);
}
```

Let me know if you see the same random cursor behavior described above and what you think about the fix.

robertoissc: Hey @theck13, you're right, I'm seeing the same behavior.
I've applied the changes you suggest and it looks good after some testing!

PR #648, opened 2019-02-28
Fix Inspection Warnings

robertoissc:
Fixed warnings:
- Use @NonNull annotation to match parent method.
- Use character class which is faster and simpler instead of char alternation in regex.

This is a partial code structure cleanup for #564


roundhill:
@robertoissc Thanks for your recent PRs. I've had a change of focus, so won't be able to get to these. But someone will soon 👍

robertoissc:
Thanks for the heads up! @roundhill

robertoissc:
Hey @theck13!
I agree with you and I'll complete the fixes this week.

PR #652, opened 2019-03-11
Add Icon for Shortcut

### LINKED ISSUE #467: Need an Icon for Shortcut
#459 added a handy way to create a new note from the app icon. It is currently using a black icon that we use for the new note icon in the app, and it doesn't look so good:

screen shot 2017-10-09 at 9 54 19 am

We can't dynamically tint the icon so we'll want to create one with a blue tint, and probably with a circular background?

robertoissc:
Fixes #467 by adding a new icon for the shortcut.
The design is based on the existing icon, platform guidelines and the comments in #467.

Here's how it looks like among other app's shortcuts:
Screenshot_20190311-170106


robertoissc:
Thanks, sounds great. I'll use vector and apply suggested adjustments.
In progress @theck13!

robertoissc:
Hey @theck13! The icon is now on vector format and adjusted to match your suggestions and the guidelines.

Issue #446, opened 2017-05-15
Any option to show md files in preview mode first


dxnacho:
I found out that note always open first in Edit mode. As most of the times I just wanna check notes and not write on them, I was wondering if a new option to "Show md notes in Preview first" could be added.

Many thanks!

theck13:
Thanks for the feedback!  It looks like the same feature was requested in https://github.com/Automattic/simplenote-android/issues/434.  I'll close this issue since it's a duplicate.

dxnacho:
Thanks!!

PR #656, opened 2019-04-02
Add Setting for Preview Mode First

### LINKED ISSUE #434: Allow the user to default to preview mode
Automatically load a note in edit mode or read-only _'preview'_ mode depending on a settings toggle.

robertoissc:
Fixes #434 and #446 by adding a preference to show preview mode as default on markdown notes.
If this setting is enabled, markdown notes will be loaded in 'preview' tab as default.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
@@ -17,6 +17,7 @@ import android.view.View;
 import android.view.inputmethod.InputMethodManager;
 
 import com.automattic.simplenote.utils.DisplayUtils;
+import com.automattic.simplenote.utils.PrefUtils;
 import com.automattic.simplenote.utils.ThemeUtils;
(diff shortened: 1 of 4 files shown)


theck13:
I started reviewing this pull request weeks ago, but I had second thoughts about how we should handle this feature.  Sorry for taking so long to respond.

Having a setting for all notes to show either the ***Edit*** or ***Preview*** tab when a note is opened may not make sense if all notes don't have markdown enabled.  Since markdown is set per note, having this setting per note may be a better option.  That would allow users to choose which view to show when opening each markdown note rather than having one setting for all markdown notes.  I imagine some users have notes that are mostly being edited and notes that are mostly being previewed.  Giving users the flexibility to choose may be best.

That solution would be much more complex though.  I have a rough draft of that implementation.  Once the code is more polished, I'll create a pull request.  Thanks for the inspiration!

robertoissc:
Hey @theck13! That sounds good to me, please let me know if I can be of any help with this.

PR #697, opened 2019-06-07
Fix accessibility warnings

robertoissc:
### Fix
This is a partial code cleanup. #564 #255
Fix accessibility warnings:
* Add missing `focusable` attributes.
* Implement `performClick` on `NoteEditorViewPagerJava` for accessibility.

### Test
* Verify `gradlew lintRelease` report doesn't containt accessibility warnings.
* Verify continuous integration build succeeds.
* Verify app builds and runs without error.
* Verify note editor pager is responding as expected on markdown enabled notes or disabled otherwise.

### Review

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
@@ -32,7 +32,7 @@ public class NoteEditorViewPager extends ViewPager {
 
     @Override
     public boolean performClick() {
-        return super.performClick();
+        return this.mIsEnabled && super.performClick();
     }
 
     public void setPagingEnabled(boolean enabled) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
@@ -19,10 +19,20 @@ public class NoteEditorViewPager extends ViewPager {
         return this.mIsEnabled && super.onInterceptTouchEvent(event);
     }
 
-    @SuppressLint("ClickableViewAccessibility")
     @Override
     public boolean onTouchEvent(MotionEvent event) {
-        return this.mIsEnabled && super.onTouchEvent(event);
+        if(mIsEnabled) {
+            if(event.getAction() == MotionEvent.ACTION_UP) {
+                performClick();
+            }
+            return super.onTouchEvent(event);
+        }
+        return false;
+    }
+
+    @Override
+    public boolean performClick() {
+        return super.performClick();
     }
 
     public void setPagingEnabled(boolean enabled) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
@@ -1,6 +1,5 @@
 package com.automattic.simplenote.widgets;
 
-import android.annotation.SuppressLint;
 import android.content.Context;
 import android.support.v4.view.ViewPager;
 import android.util.AttributeSet;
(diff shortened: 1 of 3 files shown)


Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
```diff
@@ -18,6 +19,7 @@ public boolean onInterceptTouchEvent(MotionEvent event) {
         return this.mIsEnabled && super.onInterceptTouchEvent(event);
     }
 
+    @SuppressLint("ClickableViewAccessibility")
```
theck13:
```suggestion
    @Override
    public boolean onTouchEvent(MotionEvent event) {
        performClick();
        return this.mIsEnabled && super.onTouchEvent(event);
    }

    @Override
    public boolean performClick() {
        return this.mIsEnabled && super.performClick();
    }
```

Rather than suppressing this warning, we may want to override the `performClick()` method mentioned in the warning.  I haven't tested or researched the consequences of this, but ignoring the warning could be an anti-pattern.

Simplenote/src/main/res/layout/nav_drawer_footer.xml
```diff
@@ -11,14 +11,15 @@
...
-        android:background="?attr/dividerColor"/>
+        android:background="?attr/dividerColor" />
 
     <LinearLayout
         android:id="@+id/nav_settings"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:background="@drawable/selectable_background_simplestyle"
         android:clickable="true"
+        android:focusable="true"
```
theck13: We definitely want to add the `android:focusable="true"` attribute to this view as you've done, but the rest of the changes to the file are just moving existing attributes around and are unnecessary.  We can revert those.

Simplenote/src/main/res/layout/nav_drawer_row.xml
```diff
@@ -11,49 +11,50 @@
...
-            android:textSize="@dimen/nav_drawer_item_text_size"/>
+            android:textSize="@dimen/nav_drawer_item_text_size" />
 
         <com.automattic.simplenote.widgets.RobotoMediumTextView
             android:id="@+id/edit_tags"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:background="?attr/selectableItemBackgroundBorderless"
             android:clickable="true"
+            android:focusable="true"
```
theck13: This is another instance where adding the `android:focusable="true"` attribute is good, but can revert the rest of the changes.

Simplenote/src/main/res/layout/nav_drawer_footer.xml
robertoissc: Hey @theck13!
You're right, I wasn't sure about the formatting changes but I've reverted those!

Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
robertoissc: Hey @theck13! I tried to do some research about this and I think we should call `performClick()` only on `MotionEvent.ACTION_UP` action but I'm not sure my implementation is the right one.

Please let me know if you prefer that I use this approach instead or have any other suggestion.

theck13: Good idea about the `MotionEvent.ACTION_UP` action.  I think your solution works.  We might want to update the `onTouchEvent` method like the code below.  It's effectively the same thing as your solution, but it uses the early exit pattern used throughout the codebase.  What do you think?

```java
    @Override
    public boolean onTouchEvent(MotionEvent event) {
        if (!mIsEnabled) {
            return false;
        }

        if (event.getAction() == MotionEvent.ACTION_UP) {
            performClick();
        }

        return super.onTouchEvent(event);
    }
```

PR #705, opened 2019-06-13
Update Passcode Layout

theck13:
### Fix
Update the passcode layout with vector drawables and theme changes to match the app style.  See the screenshots and animations below for illustration.

update_passcode

Before|After
-|-
update_passcode_before|update_passcode_after

### Test
1. Open navigation drawer.
2. Tap ***Settings*** item.
3. Tap ***Turn PIN lock on*** under ***PIN Lock*** section.
4. Enter PIN lock passcode.
5. Reenter PIN lock passcode.
6. Exit app.
7. Wait five seconds.
8. Open app.
9. Notice updated app logo, number buttons, and fingerprint image.
10. Long-press any number button.
11. Notice updated button background.

### Review

#### Note
Since these changes require a `minSdkVersion` value of `21` or higher, the pull request will be in draft status until https://github.com/Automattic/simplenote-android/pull/689 is merged.


SylvesterWilmott:
This looks great @theck13 , thanks!

Do you think it would be worth adding some vibration feedback when tapping the number since this exists on Android's own PIN keypad?

theck13:
<blockquote>
I keep getting this crash when entering a wrong pin in the app
</blockquote>

Good find, @anitaa1990!  I missed the `AppUnlockLinearLayout1` identification for the passcode lock layout, which is causing that crash.  I added the attribute and value for both layouts in 97b1cee.

<blockquote>
Btw, I love that can test both the debug and release builds at the same time 😄
</blockquote>

Me too!  That's the magic of setting the [`applicationIdSuffix` for different build types](https://github.com/Automattic/simplenote-android/blob/146cc9bb014f39912bcee7565e09663421277733/Simplenote/build.gradle#L96).

<blockquote>
Do you think it would be worth adding some vibration feedback when tapping the number since this exists on Android's own PIN keypad?
</blockquote>

Good idea!  I [added haptic feedback to the Passcode Lock library](https://github.com/wordpress-mobile/PasscodeLock-Android/pull/58) and updated the dependency in 9926afa.  Now, the device will vibrate when tapping a number or the backspace key and haptic feedback is enabled on the device.

theck13:
Thanks, @anitaa1990!  The `[status] needs design review` label has been removed.  Merge whenever you're ready.

Issue #708, opened 2019-06-18
Clicking on the first tag in the list does not open the list of notes


AmandaRiu:
### Expected
Clicking on the first tag in the list opens a list of notes assigned to that tag.

### Observed
Clicking on the first tag in the list does not do anything.

### Reproduced
1. Open the navigation drawer
2. Click on the first tag in the list - nothing happens

### Devices
Make|Model|Android Version|App Version
-|-|-|-
Nexus|6P|27|1.7.0
Google|Pixel 3A|28|1.7.0

theck13:
I'm unable to reproduce that behavior with the devices below.  I tested both alphabetical and non-alphabetical tag sorting.  The first tag in the navigation drawer list showed the correct notes with that tag in both cases.  See the animation [here](https://drive.google.com/open?id=1_diT74VBAyR2bw1wP3WuVdR6uPEp_18o) for illustration.

Make|Model|Android Version|App Version
-|-|-|-
Nexus|6P|27|1.7.0
Google|Pixel 2|25|1.7.0

This _may_ be related to some open tag issues (https://github.com/Automattic/simplenote-android/issues/239, https://github.com/Automattic/simplenote-android/issues/279).  Can you confirm your devices have the same tags and counts as the web (i.e. app.simplenote.com)?  If not, then it's probably related to one or more of the open tag issues.

PR #723, opened 2019-07-03
Issue/708 first tag focus

theck13:
### Fix
Remove the `android:focusable="true"` attribute from the ***Edit*** view in the tags header group of the navigation drawer item layout in order to close https://github.com/Automattic/simplenote-android/issues/708.  The `android:focusable="true"` attribute makes the first tag view unable to be tapped/clicked.  Removing the attribute allows the tag view to be tapped/clicked.  The attribute was added to remove the lint warning stating when a view is `clickable` and not `focusable`, it cannot be accessed by the keyboard.  The `tools:ignore="KeyboardInaccessibleWidget"` attribute was added to suppress the lint warning.  Even though adding the `android:focusable="true"` attribute removed the lint warning and the ***Edit*** view was able to be focused using a keyboard, the view was not able to be clicked by the keyboard due to the custom navigation drawer layout.  Therefore, removing the `android:focusable="true"` attribute and adding the `tools:ignore="KeyboardInaccessibleWidget"` attribute does not change the ability to click the ***Edit*** view using a keyboard, removes the lint warning, and resolves the first tag view issue.

### Test
0. Use account with at least one tag.
1. Open navigation drawer.
2. Tap first tag in list.
3. Notice tag is selected as expected.

### Review

#### Note
@loremattei, this bug was introduced with the 1.8.0 changes and requires a hotfix.

### DIFF
--- a/Simplenote/src/main/res/layout/nav_drawer_row.xml
+++ b/Simplenote/src/main/res/layout/nav_drawer_row.xml
@@ -42,7 +42,8 @@
             android:textColor="@color/simplenote_blue"
             android:textSize="@dimen/nav_drawer_item_text_size"
             android:textStyle="bold"
-            tools:text="Edit"/>
+            tools:ignore="KeyboardInaccessibleWidget"
+            tools:text="Edit" />
 
     </LinearLayout>
 
--- a/Simplenote/src/main/res/layout/nav_drawer_row.xml
+++ b/Simplenote/src/main/res/layout/nav_drawer_row.xml
@@ -36,7 +36,6 @@
             android:layout_height="wrap_content"
             android:background="?attr/selectableItemBackgroundBorderless"
             android:clickable="true"
-            android:focusable="true"
             android:padding="@dimen/padding_extra_small"
             android:text="@string/edit"
             android:textAllCaps="true"


theck13:
Nice catch, @anitaa1990!  I agree; opening the first tag after tapping the ***Tags*** header is a little weird.  These changes revert to the previous behavior.  In other words, that is the same behavior in Simplenote 1.7.0.  Even though I don't think it's great, it's better than not being able to tap the first tag at all.  Because of that, I would say we merge this to fix the issue and work on the ***Tags*** header issue when we rethink the navigation drawer.  What do you think?

anitaa1990:
I'm good with fixing this issue in another PR @theck13 👍 I'm not sure if an issue for this exists previously. If not, I'll create one first thing tomorrow!

anitaa1990:
@theck13 I'll merge the changes as soon as the CircleCI tests are done 👍

theck13:
Thanks, @anitaa1990!  I'm not sure why the ***ci/circleci: Build Error*** is still pending.  @loremattei, do you know what could cause that task to remain pending for hours?

loremattei:
@theck13 it looks like a hiccup in CircleCi<->GitHub communication... all the expected tasks seem to be completed and green.
I think we can move this ahead and merge the changes. 

cc/ @jtreanor do you have any idea about this?

loremattei:
@theck13 it turned out it was a misconfiguration in the branch protection. It's now fixed.

PR #735, opened 2019-07-08
Fix Correctness>Messages warnings

robertoissc:
### Fix
This is a partial code cleanup. #564 #255
Fix Correctness>Messages warnings:
* ~~Ignore `Potential Plurals` warnings for `note_selected` and `notes_selected` strings since they are used by the plural already defined in `plurals.xml`.~~
* Remove untranslatable string `about_twitter_handle` from `strings.xml(iw)`.

### Test
* Verify `gradlew lintRelease` report doesn't contain Correctness>Messages warnings.
* Verify continuous integration build succeeds.
* Verify app builds and runs without error.

### Review

### DIFF
--- a/Simplenote/src/main/res/values/strings.xml
+++ b/Simplenote/src/main/res/values/strings.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
 
     <string name="app_name" translatable="false">Simplenote</string>
     <string name="app_launcher_name" translatable="false">@string/app_name</string>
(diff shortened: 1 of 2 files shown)


Simplenote/src/main/res/values/strings.xml
```diff
@@ -36,8 +36,8 @@
     <string name="website">Visit simplenote.com</string>
     <string name="more_info">More Information</string>
     <string name="note_pinned">Note pinned.</string>
-    <string name="note_selected">%d note selected</string>
-    <string name="notes_selected">%d notes selected</string>
+    <string name="note_selected" tools:ignore="PluralsCandidate">%d note selected</string>
+    <string name="notes_selected" tools:ignore="PluralsCandidate">%d notes selected</string>
```
theck13: Both the `note_selected` and `notes_selected` resources are referenced in the [`plurals`](https://github.com/Automattic/simplenote-android/blob/8ae53102bedeace1f843017d5bf84b98001094db/Simplenote/src/main/res/values/plurals.xml#L4-L5) file so we don't want to add the `tools:ignore="PluralsCandidate"` attribute to them.  The changes to the `strings.xml` file in this pull request can be reverted.

robertoissc: Hey Tyler, thanks for the feedback!
I thought it'd be nice to get rid of those warnings in `strings.xml` file since those strings are already referenced in the `plurals.xml` file as you mention.
I've reverted the changes to `strings.xml` ;)

PR #738, opened 2019-07-10
Add Sentry (and the secret store)

jkmassel:
### Fix
Adds Sentry crash logging (and as a side-effect, support for the secret store).

### Release
These changes are not user-facing, and don't require a release notes update.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/CrashUtils.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/CrashUtils.java
@@ -1,0 +1,114 @@
+package com.automattic.simplenote.utils;
+
+import android.content.Context;
+import android.content.res.Resources;
+import android.os.Build;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import com.automattic.android.tracks.CrashLogging.CrashLogging;
+import com.automattic.android.tracks.CrashLogging.CrashLoggingDataProvider;
+import com.automattic.android.tracks.TracksUser;
+import com.automattic.simplenote.BuildConfig;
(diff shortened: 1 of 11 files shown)


Simplenote/src/main/java/com/automattic/simplenote/utils/CrashUtils.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public static void clearCurrentUser() {
+        sharedInstance.currentUser = null;
+        CrashLogging.setNeedsDataRefresh();
+    }
+
+    @Override
+    public String sentryDSN() {
+        return BuildConfig.SENTRY_DSN;
```
loremattei: This doesn't seem to work well with what we have in the secrets repository. The entry there is "sentryDSN" which would translate to SENTRY_D_S_N here. 
The entry looks correct in the example file here, so I guess you want to change the file in the repo.

loremattei:
The code looks good to me and the secret store integration works well. 
I had a small build issue, as reported above. 
The other issue I have is that, after having fixed the error above locally, my crashes don't show up in Sentry.

Simplenote/src/main/java/com/automattic/simplenote/utils/CrashUtils.java
jkmassel: 🤦‍♂ – you're totally right! Had to rework the BuildConfig logic when merging with `develop` and didn't add it to the store. Updated the PR with the new commit hash and tested this all again from scratch. Should be good now.

Sorry about that!

loremattei:
> Is it possible you have the "Share analytics" switch turned off in Settings? 
It was definitely it! Thanks for the suggestion :-)

PR #749, opened 2019-07-12
Add finalize release lane

loremattei:
This PR adds a lane to automate some of the final release steps.
The PR depends on and should be merged after https://github.com/Automattic/simplenote-android/pull/748.

# To test:
### Setup
1. Run `bundle install`.
2. Reopen the `1.8`milestone.
3. Checkout a new `release/1.88` (or what you want) branch and push it with the -u option (the PR will commit and push on the current branch).
4. Edit `versionName` in `build.gradle` to be `1.8` (anything that hasn't a tag yet and that has an open milestone works as well).

### Run
5. Run `bundle exec fastlane finalize_release`.
6. At the prompt, verify that the version is right and confirm.
7. Verify that the code pulls all the translations and run the release checks.
8. Verify that the metadata is downloaded in the `fastlane/metadata/android` folder and that the `release_notes.xml` file is updated.
9. Verify that the `1.8` tag is created and pushed.
10. Verify that the `1.8` milestone is closed.

### Wrap up
11. Delete the `1.8` tag on your local repository and on GitHub.
12. Delete the test branch.

PR #750, opened 2019-07-12
Add script to automate builds

loremattei:
This PR adds a lane to automate the pre-build checks and the build itself.
This PR builds on top of https://github.com/Automattic/simplenote-android/pull/749 and should be merged after that one. 

# To Test
1. Run `bundle install`.
2. Run `bundle exec fastlane build_release`.
3. When prompted, verify that the version is correct and confirm.
4. If prompted, update the secret file.
5. Verify that the script finishes without errors.
6. Verify that the apk has been generated in the `build` folder.
7. Test the apk.


fastlane/Fastfile
```diff
@@ -8,16 +8,277 @@ platform :android do
...
+  #####################################################################################
+  # build_bundle
+  # -----------------------------------------------------------------------------------
+  # This lane builds an app bundle
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane build_bundle [skip_confirm:<skip confirm>]
+  #####################################################################################
+  desc "Builds an app bundle"
+  lane :build_apk do | options |
```
jtreanor: Is there a reason we are building an apk rather than an app bundle? I guess we just haven't gotten around to that yet.

theck13: Yep, we are still releasing the full APK.  There isn't a large size savings when using app bundle for Simplenote Android.  The app is only 3.8MB at the moment and we would reduce the download size by about 12%.

jtreanor: Makes sense. We may want to make the change at some point anyway for consistency but not a big deal.

loremattei: Oh yes, just something to do.

PR #782, opened 2019-08-26
Merge Release/2.0 into develop

loremattei:
Merge release/2.0 into develop.

### Release

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,6 @@
+2.1
+-----
+ 
 2.0.0
 -----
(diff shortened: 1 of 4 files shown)


loremattei:
@theck13 it looked less important than the others, so I skipped it to keep the release notes around the suggested 250 chars. Some translations end up being quite longer and we easily hit the 500 chars limit. 
Anyway, if you think it's important, I can re-add it and then manually cut it from the translations that turn out to be too long.

loremattei:
Thank you @theck13! We can try and upload the complete set for translation. I'll remove the entry you suggested from the translations that won't fit the limit.

PR #791, opened 2019-08-28
Update Note List

theck13:
### Fix
Update the note list interface for better performance and readability.  These changes include updating the note list item layout and floating action button icon.  The navigation bar style will be included in a separate pull request.  See the screenshots below for illustration.

update_note_list_expanded_light

update_note_list_expanded_dark

update_note_list_condensed_light

update_note_list_condensed_dark

### Test
Follow the steps below with the ***Condensed note list*** setting enabled and disabled with both ***Light*** and ***Dark*** themes applied to verify the layouts shown in the screenshots above.
1. Go to note list.
2. Notice updated note list item layout.
  a. Notice list dividers are hidden between items.
  b. Notice published icon is shown for published notes.

### Review

### Release
`RELEASE-NOTES.txt` was updated in 42f2345 with:
> Updated note list interface for better performance and readability


SylvesterWilmott:
Looking good @theck13 ! Thanks

**A few bits and bobs:**
- Text sizes appear off but looks to be inherited from the current list. The note title should be 16sp and note excerpt 14sp
- We have this behaviour where if I create a new note, type nothing and then return to the notes list I see a "New note..." row there that is empty. Unless there is some important background to this, can we kill this behaviour?

theck13:
> Text sizes appear off but looks to be inherited from the current list. The note title should be 16sp and note excerpt 14sp

Yep, those are the values used.  You can see them in the [`dimens.xml` file](https://github.com/Automattic/simplenote-android/pull/791/files#diff-051a7688623d9e5d2efa9ba60e6788c4R52-R53).

> We have this behaviour where if I create a new note, type nothing and then return to the notes list I see a "New note..." row there that is empty. Unless there is some important background to this, can we kill this behaviour?

That might be a regression.  It should have been fixed in https://github.com/Automattic/simplenote-android/pull/532.  If it's still occurring, we should revisit that issue.

SylvesterWilmott:
>Yep, those are the values used. You can see them in the dimens.xml file.

Hmm there must be an issue somewhere. The text sizes in the APK are 17 and 15 when measured.
example:

Screenshot 2019-08-30 at 10 28 00
(The red overlayed text above is 17 and 15)

>That might be a regression. It should have been fixed in #532. If it's still occurring, we should revisit that issue.

Yeah this was still happening in the APK you sent me

Also, I don't want the note status discussion to be a blocker for this PR, can we align the icons side by side just so we have something decent while we continue exploring options for the note status?

Screenshot 2019-08-30 at 10 42 27

Screenshot 2019-08-30 at 10 43 36

Screenshot 2019-08-30 at 10 44 55

theck13:
> Hmm there must be an issue somewhere. The text sizes in the APK are 17 and 15 when measured.

Those values are being overridden by the font size value set in ***Settings*** where the user can specify ***Extra small***, ***Small***, ***Normal***, ***Large***, or ***Extra large*** font size.  I updated the logic so that the title is `16` and the excerpt is `14` for the ***Normal*** setting in a6d95049.

> can we align the icons side by side just so we have something decent while we continue exploring options for the note status?

I updated the icons to be horizontally aligned for both the default and condensed note list views in facbc72d.

See the screenshots below for illustration of both changes.

update_note_list_icons

SylvesterWilmott:
Looks good @theck13 , thank you

theck13:
@AmandaRiu, the design review has passed so the pull request is good to merge whenever you're ready.

Issue #837, opened 2019-10-09
Crash after double-tapping to delete a tag


rachelmcr:
### Expected

I expect that the app won't crash while I delete tags.

### Observed

If I double-tap on the trash icon next to a tag, the app crashes.

I confirmed that I can double-tap on the trash icon for a note without any issues.

### Reproduced

1. Add a new tag to a note.
2. Remove the tag from the note (so the tag exists but isn't used on any notes).
3. Open the navigation drawer and edit the existing tags.
4. Find the tag you added and double tap the trash icon. Result: crash.

Crash in Sentry: SIMPLENOTE-ANDROID-A9

Make|Model|Android Version|App Version
-|-|-|-
Motorola|e5 play|8.1.0|2.1-rc-1

PR #838, opened 2019-10-09
Issue/837 crash delete tag

theck13:
### Fix
Add a check to ensure the tags list has the item to delete when the ***Delete tag*** button is tapped before attempting to delete the tag to close https://github.com/Automattic/simplenote-android/issues/837.

### Test
1. Tap any note in list.
2. Add new tag to ***Add tag...*** field.
3. Remove new tag from Step 2.
4. Tap back arrow in app bar.
5. Open navigation drawer.
6. Tap ***Edit*** button in ***Tags*** section.
7. Double-tap ***Delete tag*** button for tag from Step 2.
8. Notice app does not crash.
9. Notice tag from Step 2 is removed.

### Review
 @rachelmcr, I also requested you as a reviewer since you caught the associated bugs and to ensure they are fixed.

### Release

#### Note
@loremattei, these changes will require a new release candidate build once they are merged.


rachelmcr:
I tested with the same device and steps as before and confirmed that this fixes the issue! 👍 Thanks!

PR #851, opened 2019-10-21
Update fastlane

loremattei:
This PR updates `fastlane` to `2.133` to match the version we have on the other apps. It also update bundle to `2.0.2` to fix some installation issues with XCode 11.1.

### Test
- Verify that `bundle install` can be completed without errors.
- Verify that CI is green.

### Review

### Release


loremattei:
@theck13 it allows to run `bundle install` on any system where XCode 11 is installed. So, while it's not strictly necessary for the project, it solves a problem for any developer that has XCode 11 installed on their machine. 

If you see issues with this, we can try to figure out other solutions.

theck13:
@loremattei, thanks for the explanation!  Keeping the `2.0.2` changes are fine with me.

PR #870, opened 2019-11-05
Issue/844 add recent suggestions

### LINKED ISSUE #844: Search: Add Recent to Suggestions
Add recent search queries to the search input field as shown in the mockup below.

<a href="https://user-images.githubusercontent.com/3827611/66611876-12c74780-eb7d-11e9-90c7-835d1f01f666.png"></a>

theck13:
### Fix
Add recent search queries to the search input field to close #844.  The recent search list is ordered from most recent at the top to least recent at the bottom.  Each submitted search is saved as a recent search with a maximum of five recent searches.  When a search is submitted after five recent searches have been saved, the least recent search is removed.  When a recent search is tapped, it becomes the most recent search.  Each recent search can be deleted with the ***X*** (i.e. ***Delete item***) button to the right of the search item row.  Deleting a recent search can be undone by tapping the ***Undo*** action in the ***Deleted recent search item*** snackbar.  See the screenshots and animation below for illustration.

844_add_recent_suggestions

<a href="https://user-images.githubusercontent.com/3827611/68254204-34361a80-ffe7-11e9-9abb-a39732934d6f.gif"></a>

### Test
1. Tap ***Search*** action in top app bar.
2. Notice no recent searches are shown.
3. Enter text into search field.
4. Notice suggestions appear based on entered text.
5. Tap any suggestion.
6. Notice suggestion text is inserted into search field.
7. Notice note list is shown based on inserted text.
8. Tap ***X*** button in search field.
9. Notice previously searched item is shown in recent searches.
10. Enter text into search field.
11. Tap magnifying glass button in keyboard.
12. Notice note list is shown based on input text.
13. Tap ***X*** button in search field.
14. Notice search from Step 10 is shown in recent searches followed by search from Step 5.
15. Tap recent search from Step 5.
16. Notice recent search text is inserted into search field.
17. Notice note list is shown based on inserted text.
18. Tap ***X*** button in search field.
19. Notice recent search from Step 5 is shown in recent searches follow by search from Step 10.
20. Long-press ***X*** button to right of any recent search.
21. Notice ***Delete item*** message is shown.
22. Tap ***X*** button to right of any recent search.
23. Notice recent search item from Step 22 is removed from recent searches.
24. Notice ***Deleted recent search item*** snackbar.
25. Tap ***Undo*** action in snackbar.
26. Notice recent search item from Step 22 is added to recent searches in previous position.

### Review

### Release
`RELEASE-NOTES.txt` was updated in 7f2f9c7 with:
> Added search history and suggestions


Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -709,6 +761,86 @@ public boolean hasSearchQuery() {
...
+        } catch (BucketObjectMissingException exception) {
+            // Create entity for recent searches since it doesn't exist.
+            try {
+                Search search = mBucketSearch.newObject(KEY_SEARCH);
+                search.save();
+            } catch (BucketObjectNameInvalid invalid) {
+                Log.e("addSearchItem", "Could not create search object", invalid);
+            }
+        }
+    }
```
dmsnell: Just for fun and to help re-familiarize myself with Java I have created a version of add/remove using Streams

https://repl.it/repls/ZealousOrchidLinkedlist

```java
public static final int MAX_SEARCHES = 5;

public static List<String> useSearch(List<String> searches, String search) {
  return Stream.concat(
    Stream.of(search),
    searches.stream().filter(s -> !s.equals(search))
  )
    .limit(MAX_SEARCHES)
    .collect(Collectors.toList());
}

public static List<String> removeSearch(List<String> searches, String search) {
  return searches
    .stream()
    .filter(s -> !s.equals(search))
    .collect(Collectors.toList());
}
```

SylvesterWilmott:
This looks great and works as expected, I like the fact that we don't duplicate recent searches in the list if we perform the same search twice.

One small detail, for the snackbar we are using the light blue colour for the "undo" text, can we just switch this for the blue we use for text in light mode?

SylvesterWilmott:
>One small detail, for the snackbar we are using the light blue colour for the "undo" text, can we just switch this for the blue we use for text in light mode?

Confirmed that we'll fix this in a separate PR

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -140,6 +148,7 @@ public void onNoteSelected(String noteID, int position, String matchOffsets, boo
     private TextView mSortOrder;
     private refreshListTask mRefreshListTask;
     private refreshListForSearchTask mRefreshListForSearchTask;
+    private int mIndex;
```
dmsnell: is this name a little generic for the size of this class?
what are the reasons for declaring here instead of as a local variable in `deleteSearchItem` and in the delete button callback?

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -242,7 +251,8 @@ public void onItemCheckedStateChanged(ActionMode actionMode, int position, long
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        mBucket = ((Simplenote) requireActivity().getApplication()).getTagsBucket();
```
dmsnell: this is a great change - thanks for making the name mean something specific 😊

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -447,6 +457,19 @@ public void onResume() {
         super.onResume();
         getPrefs();
         refreshList();
+        mBucketPreferences.start();
+        mBucketPreferences.addOnDeleteObjectListener(this);
+        mBucketPreferences.addOnNetworkChangeListener(this);
+        mBucketPreferences.addOnSaveObjectListener(this);
```
dmsnell: What's the purpose of putting these all in `NoteListFragment` instead of in their own class, something like a `PreferencesManager` and connecting changes/updates to that class vs. connecting to a smaller API like `mPreferencesManager.addChangeListener(this)`

Just asking out of curiosity - trying to learn the patterns and how we make decisions on Android

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -740,10 +761,71 @@ public boolean hasSearchQuery() {
...
+            preferences.setRecentSearches(recents.subList(0, recents.size() > MAX_RECENT_SEARCHES ? MAX_RECENT_SEARCHES : recents.size()));
+            preferences.save();
+        } catch (BucketObjectMissingException exception) {
+            // Create entity for recent searches since it doesn't exist.
+            try {
+                Preferences preferences = mBucketPreferences.newObject(PREFERENCES_OBJECT_KEY);
+                preferences.save();
+            } catch (BucketObjectNameInvalid invalid) {
+                Log.e("addSearchItem", "Could not create search object", invalid);
+            }
```
dmsnell: It seems like this function and some others would be much clearer if we didn't have to expose the logic to grab the preferences entity _and_ update it. The extra code noise at least makes it hard for me to understand what is essential here vs. what is incidental…

```java
public Preferences getPreferences() {
    try {
        return mBucketPreferences.get(PREFERENCES_OBJECT_KEY);
        // oops, we don't have the preferences entity yet
    }

    try {
        Preferences preferences = mBucketPreferences.newObject(PREFERENCES_OBJECT_KEY);
        preferences.save();
        return preferences;
    } catch (BucketObjectNameInvalid invalid) {
        Log.e("getPreferences", "Could not create preferences entity", invalid);
        return null; // nothing we can do
    }
}

public void addSearchItem(String item, int index) {
    Preferences preferences = getPreferences();
    if (null == preferences) return;
    List<String> recentSearches = preferences.getRecentSearches();

    recentSearches.remove(item);
    recentSearches.add(index, item);

    // Trim recent searches to MAX_RECENT_SEARCHES (currently 5) if size is greater than MAX_RECENT_SEARCHES.
    preferences.setRecentSearches(recentSearches.subList(0, Math.min(MAX_RECENT_SEARCHES, recentSearches.size())));
    preferences.save();
}
```

When I tested this it seemed to work as expected. I tried hard to understand what we're doing with the `query` and `cursor` but in the end could not. It seems like the `query` iterates over all the entities in the bucket (which in this case should be only one) and the cursor jumps to the first one.

The thing is, we grab the preferences when we do `mNoteBucket.get(PREFERENCES_OBJECT_KEY)` and this should always be the same as what the query and cursor give us - if they were any different at all we'd be in a very broken place.

So, is the query necessary at all? Why not just use `preferences.getRecentSearches()`?

I saw this pattern somewhere else - do you find the separation of getting the preferences entity helpful? I suspect that it's all complicated by trying to treat the bucket as an entity and not a major deal, just confusing.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -740,10 +761,71 @@ public boolean hasSearchQuery() {
...
+                recents = cursor.getObject().getRecentSearches();
+            }
+
+            Preferences preferences = cursor.getObject();
+            mIndex = recents.indexOf(item);
+            recents.remove(item);
+            preferences.setRecentSearches(recents);
+            preferences.save();
+        }
+    }
```
dmsnell: …if we were to have `getPreferences()`

```java
private void deleteSearchItem(String item) {
    Preferences preferences = getPreferences();
    if (null == preferences) return;
    ArrayList<String> recentSearches = preferences.getRecentSearches();

    mIndex = recentSearches.indexOf(item);
    recentSearches.remove(item);
    preferences.setRecentSearches(recentSearches);
    preferences.save();
}
```

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -740,10 +761,71 @@ public boolean hasSearchQuery() {
...
+                List<String> recents = cursor.getObject().getRecentSearches();
+
+                for (String recent : recents) {
+                    suggestions.add(new Suggestion(recent, HISTORY));
+                }
+            }
+        }
+
+        mSuggestionAdapter.updateItems(suggestions);
+    }
```
dmsnell: …here as well…

```java
private void getSearchItems() {
    Preferences preferences = getPreferences();
    if (null == preferences) return;
    ArrayList<Suggestion> suggestions = new ArrayList<>();

    for (String recent : preferences.getRecentSearches()) {
        suggestions.add(new Suggestion(recent, HISTORY));
    }

    mSuggestionAdapter.updateItems(suggestions);
}
```

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -961,6 +1043,46 @@ public void bindView(View view, Context context, Cursor cursor) {
         }
     }
 
+    @Override
+    public void onBeforeUpdateObject(Bucket<Preferences> bucket, Preferences object) {
+    }
```
dmsnell: is this required by an abstract base class? why have the empty method?

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -961,6 +1043,46 @@ public void bindView(View view, Context context, Cursor cursor) {
...
+    @Override
+    public void onDeleteObject(Bucket<Preferences> bucket, Preferences object) {
+        if (isAdded()) {
+            requireActivity().runOnUiThread(new Runnable() {
+                @Override
+                public void run() {
+                    getSearchItems();
+                }
+            });
+        }
```
dmsnell: if we invert the conditional we avoid the nesting. not sure what your style preferences are but I know that for me seeing fewer levels of nesting can help surface bugs and behaviors easier

```java
public void onDeleteObject(Bucket<Preferences> bucket, Preferences object) {
	if (!isAdded()) {
		return;
	}

	requireActivity().runOnUiThread(new Runnable() {
		@Override
		public void run() {
			getSearchItems();
		}
	});
}
```

dmsnell:
When I was playing with `getPreferences()` I didn't like that I was still checking `null == preferences` and I was trying to find a way to remove the possibility of a null exception. It was possible with `Consumer` but I don't think I like it very much. Anyway, here it is just for fun, though because I find the `new Consumer<Preferences>(){ @Override public void accept( … ) }` ugly I'm writing it with lambdas.

```java
public void withPreferences(Consumer<Preferences> callback) {
    try {
        callback.accept(mBucketPreferences.get(PREFERENCES_OBJECT_KEY));
        return;
        // oops, we don't have the preferences entity yet
    }

    try {
        Preferences preferences = mBucketPreferences.newObject(PREFERENCES_OBJECT_KEY);
        preferences.save();
        callback.accept(preferences);
    } catch (BucketObjectNameInvalid invalid) {
        Log.e("withPreferences", "Could not create preferences entity", invalid);
        // nothing to do
    }
}

public void addSearchItem(String item, int index) {
    withPreferences(preferences -> {
        List<String> recentSearches = preferences.getRecentSearches();

        recentSearches.remove(item);
        recentSearches.add(index, item);

        // Trim recent searches to MAX_RECENT_SEARCHES (currently 5) if size is greater than MAX_RECENT_SEARCHES.
        preferences.setRecentSearches(recentSearches.subList(0, Math.min(MAX_RECENT_SEARCHES, recentSearches.size())));
        preferences.save();
    });
}

private void deleteSearchItem(String item) {
    withPreferences(preferences -> {
        ArrayList<String> recentSearches = preferences.getRecentSearches();

        mIndex = recentSearches.indexOf(item);
        recentSearches.remove(item);
        preferences.setRecentSearches(recentSearches);
        preferences.save();
    });
}

private void getSearchItems() {
    withPreferences(preferences -> {
        ArrayList<Suggestion> suggestions = new ArrayList<>();

        for (String recent : preferences.getRecentSearches()) {
            suggestions.add(new Suggestion(recent, HISTORY));
        }

        mSuggestionAdapter.updateItems(suggestions);
    });
}
```

In my earlier attempts I could get rid of the check by throwing something like a `MissingResourceException` from within `getPreferences()`. I wouldn't know what to do other than crash the app if we can't get the preferences entity within the preferences bucket. I'm sure we could add a bunch of handling to put up a message.

With lambdas this would be much cleaner.

Can you share some thoughts on my different ideas and help me fit them into the style and customs of our Android code?

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
theck13: Java looks pretty bad here with the nesting, but I read this as, "If the fragment is added, then get the search items."  I think the early exit requires about the same cognitive load in this case.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
dmsnell: > it can't be a local variable

ahh…yeah I figured that out later and then forgot to come back and delete this comment or acknowledge that I figured it out. thanks

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
dmsnell: Sure, but isn't the _only_ thing that we're doing is updating the recent search list?

In other words, for the sake of the UI and the application, I'm curious why we care about these internal details in the `NoteListFragment`.

Is there anything done here that would be specific to one of those three things that couldn't be done automatically inside some kind of `RecentSearchManager`?

I'm asking because this module is already pretty big and the addition of this stuff and its new implementation functions is even more confusing when I'm trying to understand how it works and how the pieces fit together 😀

```java
RecentSearchManager recentSearches = getRecentSearchManager();

recentSearches.addOnChangeListener(new Consumer<ArrayList<String>>() {
	@Override
	public void accept(ArrayList<String> searches) {
		diffAndUpdateSearchAdapter(…);
	}
} );

new View.OnClickListener() {
	@Override
	public void onClick(View view) {
		recentSearches.add(item);
	}
}

holder.mButtonDelete.setOnClickListener(new View.OnClickListener() {
    @Override
    public void onClick(View view) {
        if (!isAdded()) {
            return;
        }

        final String item = holder.mSuggestionText.getText().toString();
        recentSearches.drop(item);

        Snackbar
            .make(getRootView(), R.string.snackbar_deleted_recent_search, Snackbar.LENGTH_LONG)
            .setActionTextColor(ThemeUtils.getColorFromAttribute(requireContext(), R.attr.colorAccent))
            .setAction(
                getString(R.string.undo),
                new View.OnClickListener() {
                    @Override
                    public void onClick(View view) {
                        recentSearches.restoreLastSearch();
                    }
                }
            )
            .show();
    }
});
```

dmsnell:
Thanks for the explanations!

> with the more recent ~Android~ language, Kotlin

FTFY 😉 

> We actually don't use lambdas anywhere in the Simplenote Android code.

Yeah I found that out and wasn't arguing to introduce them here; I was saying "since this option is clearly bad I'm going to write it the way would rather write it than a way that would be workable here"

> It's littered with null-checks everywhere. So much so that Android developers have come to expect them at this point…Since "clean" in Java is relative and null-checks are the norm, I would go with the verbose option

Interesting. Maybe this is still responding to my use of lambdas. I was asking more about `null`-checks in general since they are a massive portal for bugs and crashes. That is, I was only trying to find a way to make the failure-state unrepresentable in the code, something I wouldn't exactly frame as terse or verbose.

What's the value of returning `null` in `getPreferences()` over throwing an error and trapping it higher up? I guess I can see where there's a cascade of exceptions, but is it not possible to handle that at the top of the app where we would need to pause anyway if we can't load the preferences?

---

In any case this overlaps with a concrete question about `getPreferences` and the `null` checks everywhere - what's supposed to happen if we can't load `preferences`? If we don't do that we don't know whether to disable analytics tracking and if we silently keep going with `if ( null ) return` then won't we end up breaking in a way that hides the indication of the breakage?

In this patch if we can't load the preferences then we stop updating our recent searches - not the most jarring problem but we'll have a delete button that doesn't work.

How would you prefer to avoid this?

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -740,10 +761,70 @@ public boolean hasSearchQuery() {
...
 
+    public void addSearchItem(String item, int index) {
+        Preferences preferences = getPreferences();
+
+        if (preferences != null) {
+            List<String> recents = preferences.getRecentSearches();
+            recents.remove(item);
+            recents.add(index, item);
+            // Trim recent searches to MAX_RECENT_SEARCHES (currently 5) if size is greater than MAX_RECENT_SEARCHES.
+            preferences.setRecentSearches(recents.subList(0, recents.size() > MAX_RECENT_SEARCHES ? MAX_RECENT_SEARCHES : recents.size()));
```
dmsnell: it was a separate thing and may have been lost, but we can collapse this ternary expression into a single declaration…

```diff
- preferences.setRecentSearches(recents.subList(0, recents.size() > MAX_RECENT_SEARCHES ? MAX_RECENT_SEARCHES : recents.size()));
+ preferences.setRecentSearches(recents.subList(0, Math.min(MAX_RECENT_SEARCHES, recents.size())));
```

theck13: Interestingly, a ternary expression is used in the `Math.min(int a, int b)` source code.

```java
    /**
     * Returns the smaller of two {@code int} values. That is,
     * the result the argument closer to the value of
     * {@link Integer#MIN_VALUE}.  If the arguments have the same
     * value, the result is that same value.
     *
     * @param   a   an argument.
     * @param   b   another argument.
     * @return  the smaller of {@code a} and {@code b}.
     */
    public static int min(int a, int b) {
        return (a <= b) ? a : b;
    }
```

Since they are the nearly the exact same thing, there doesn't seem like much advantage other than developer preference.  Using the ternary expression could help reviewers or future maintainers understand the logic a little easier.  That's also why I added the comment above that statement to explain what's going on.

theck13:
@anitaa1990 and @dmsnell, is there anything else we would like to add/update to these changes or is this ready to merge?

dmsnell:
@theck13 I think it's worth renaming `mIndex` to something less generic, maybe `mDeletedItemIndex` as you suggested.

Apart from that I have no objections; thanks for sharing your thoughts with me on this.

theck13:
I updated the variable name in e21ae9dd.

theck13:
@anitaa1990 and @dmsnell, this is good to merge whenever you're ready.

PR #871, opened 2019-11-06
Support the unicode bullet as a list item bullet in unordered lists

jivox92362:
### Fix

Support the unicode bullet character `•` as a list item bullet because lists copied from HTML or word documents may contain this character.

### Test

1. Open Simplenote
2. Create an unordered list with unicode bullets
```
  • list item a
  • list item b
```
3. Press enter after the first item
4. A new list item should be opened, starting with a unicode bullet •

### Review

Everyone can review this if they want :)

### Release

`RELEASE-NOTES.txt` was updated in 9b9b644 with:
> Added support for the unicode bullet • in list items

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 2.3
 -----
 * Added search sorting by relevance, date created, date modified, and alphabetically
+* Added support for the unicode bullet • in list items
 
 2.2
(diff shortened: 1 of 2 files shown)


jivox92362:
See https://github.com/Automattic/simplenote-electron/pull/1551 for the change in simplenote-electron.

PR #880, opened 2019-11-12
Issue/879 note sync pause

### LINKED ISSUE #879: Notes do not sync if you switch apps directly from editor
When editing a note, if you do not return to the notes list or reopen the note again later, it will not sync from Android to other devices. This also occurs if you close the app from the Recent Apps screen.

### Expected
That Simplenote syncs as I type, or in such a way that it's not synced by specific actions such as returning to the notes list or reopening the note.

### Observed
When switching apps on Android, rather than hitting the home button, or the back button to return to the Notes List, any recent updates do not sync.

### Reproduced
1. Open a note in Android
2. Type a few words at least
3. Quickly tap the Recent Apps/Open Apps button in Android to switch to another open app or close Simplenote
4. Log in on web, MacOS, or other Simplenote app
5. You will not see the change
--
6. Open the note again in Android, or Simplenote at all
7. Change syncs right away

I can reproduce this consistently

-|-|-|-
Samsung | Galaxy Note 10 | Android 9 | Simplenote 2.2

theck13:
### Fix
Update the note editor to remove the note bucket listener after `onPause` and note save completes to close #879.  The note bucket listener was being removed before the note object was being saved when the `onPause` method was called in the `NoteEditorFragment` class.  Consequently, edits to a note were not being synced when exiting the app within two seconds of the edits.  These changes make sure the note is saved before the note bucket listener is removed.

### Test
In order to see whether the note edits are saved and synced when exiting the Android app, logging in to Simplenote on a web browser is helpful.

#### Pause
1. Tap any note in list.
2. Add text to note.
3. Notice added text appears in Android app.
4. Before two seconds pass, tap the ***Home*** button in the navigation bar.
5. Notice added text appears in web app.

#### Destroy
1. Tap any note in list.
2. Add text to note.
3. Notice added text appears in Android app.
4. Before two seconds pass, tap the ***Recent*** button in the navigation bar.
5. Swipe left or right to remove Simplenote from recent apps.
6. Notice added text appears in web app.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -127,6 +127,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
     private String mMatchOffsets;
     private int mCurrentCursorPosition;
     private HistoryBottomSheetDialog mHistoryBottomSheet;
+    private boolean mIsPaused;
     // Hides the history bottom sheet if no revisions are loaded
     private final Runnable mHistoryTimeoutRunnable = new Runnable() {
         @Override
@@ -434,9 +435,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
     @Override
     public void onPause() {
         super.onPause();  // Always call the superclass method first
-
-        mNotesBucket.removeListener(this);
-        mNotesBucket.stop();
+        mIsPaused = true;
 
         // Hide soft keyboard if it is showing...
         DisplayUtils.hideKeyboard(mContentEditText);
@@ -457,6 +456,13 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
 
         mHighlighter.stop();
         saveNote();
+    }
+
+    @Override
+    public void onDestroy() {
+        super.onDestroy();
+        mNotesBucket.removeListener(this);
+        mNotesBucket.stop();
     }
 
     @Override
@@ -1234,7 +1240,10 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
 
     @Override
     public void onSaveObject(Bucket<Note> noteBucket, Note note) {
-        // noop
+        if (mIsPaused) {
+            mNotesBucket.removeListener(this);
+            mNotesBucket.stop();
+        }
     }
 
     @Override


Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -434,9 +435,7 @@ public void run() {
     @Override
     public void onPause() {
         super.onPause();  // Always call the superclass method first
-
```
dmsnell: how would this have been different/broken if we had moved `saveNote()` up here instead of moving `mNotesBucket.stop()` to `onDestroy()`?

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -1234,7 +1240,10 @@ public void run() {
 
     @Override
     public void onSaveObject(Bucket<Note> noteBucket, Note note) {
-        // noop
+        if (mIsPaused) {
+            mNotesBucket.removeListener(this);
+            mNotesBucket.stop();
+        }
```
dmsnell: does this get called by `saveNote()`? is it called in a separate thread? why would it be called after pausing?

dmsnell:
Are there any remaining timing issues here (e.g. does this mitigate most of the timing issues vs. eliminate them altogether)?

Awesome fix though, and fast! ⚡️

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
dmsnell: Thanks for your reply. Are we intending to shut down all updates from the server during pause or are we mostly concerned here with sending out our local changes before sleeping?

theck13: We're mostly concerned about sending out our local changes before pausing/sleeping.  Android could destroy activities and fragments once they have been paused.  If that happens before the local changes are sent, it could result in content loss.

dmsnell: Why stop listening in `onPause` and not in `onStop`?

If we switch apps and come back won't we have to start the bucket connection again?
Don't we want updates to be processed in the background?

theck13: We're not stopping `onPause`.  We're stopping `onDestroy`, which is after `onStop`, if the fragment has been paused already (i.e. `mIsPaused`).

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
dmsnell: > I'm guessing the mNote.save(); statement in the saveNote() method doesn't save/sync immediately and queues the changes for later.

That should be true, and also it only sends one change at a time and waits for each to return before applying the next ones - it may have to rewrite the internal changes.

That is, I can see where we definitely want to avoid this…

```java
saveNote();
mNotesBucket.stop();
```

…because that's going to interrupt a process that's relatively slow (makes and waits for a network call) _and_ asynchronous.

Questioning as below, do we know that we need to stop the bucket connection on pause? According to [the `onPause` docs](https://developer.android.com/guide/components/activities/activity-lifecycle) it looks like they are suggesting that this kind of thing be done in `onSave()`

It's easy to see both arguments for stopping the bucket connection: stop sending and listening to network calls when the app's not in the foreground; keep sending queued changes and listening for new ones so that when the app returns it will be ready.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
dmsnell: > since Android could destroy activities and fragments once they have been paused.

ah, so they don't call `onStop` or `onDestroy` when they do this?

theck13: It _should_ call `onStop` and `onDestroy`, but `onPause` is more of a guarantee.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
theck13: > According to [the `onPause` docs](https://developer.android.com/guide/components/activities/activity-lifecycle) it looks like they are suggesting that this kind of thing be done in `onSave()`

Do you mean `onStop` rather than `onSave` here?  If so, I think that's being discuss in [the other comment thread](https://github.com/Automattic/simplenote-android/pull/880#discussion_r345260215).

dmsnell:
I missed your last comment!

> Are there ay in particular you're aware of or was that just a general question?

It was the timing issue between `saveNote()` and `mNoteBucket.stop()`

> Since this issue could result in content loss, I chose these changes to ensure the content is saved and synced. It's more of a short-term fix than a long-term solution.

If I'm reading you correctly then I think you might be indicating that there's a known bug in the `simperium-android` library.

The _only_ place we should experience loss on a hard crash (except when we're in the middle of executing the code in `saveNote()` and haven't stored the change in the local database yet) is in the debounce window before calling `saveNote()`. Once the note bucket is notified that content has changed it should hold on to those changes through crashes and network failures and restarts.

That is, calling `noteBucket.stop()` should make it less likely that our changes get sync'd with the server, but it shouldn't increase our chance of data loss (other than the fact that by not sending those local changes to the server more conflicting edits could occur before we reconnect).

> We may want to consider creating a service that will be able to save, delete, sync, etc. all in the background even when the app has been paused and destroyed

This seems like a good plan and related to my surprise to find `mNoteBucket` inside the `NoteFragment`. It seems like from the moment we open the application until we quit out of the application then our buckets should be active (unless we are told by Android to stop making network calls and processing things while in the background).

PR #891, opened 2019-11-20
Issue/890 add tag selected

### LINKED ISSUE #890: Search: Add Tag to Search When Tag Selected in Navigation Drawer
When a tag (e.g. `work`) is selected from the navigation drawer and the _*Search*_ action is tapped, add the `tag:` prefix with the tag name to the search input field (e.g. `tag:work`).

theck13:
### Fix
When a tag (e.g. `work`) is selected from the navigation drawer and the _*Search*_ action is tapped, add the `tag:` prefix with the tag name to the search input field (e.g. `tag:work`) to close #890.  A space was also added after the tag prefix and tag name so that the user can start typing additional search criteria without entering a space first.

### Test
1. Open navigation drawer.
2. Tap ***All Notes*** or ***Untagged Notes*** item in navigation drawer.
3. Tap ***Search*** action in top app bar.
4. Notice search field is empty.
5. Notice recently submitted searches are shown.
6. Tap back arrow in top app bar.
7. Open navigation drawer.
8. Tap any tag under ***Tags*** section.
9. Tap ***Search*** action in top app bar.
10. Notice search field contains `tag:` followed by tag name from Step 7 and a space.
11. Notice recently submitted searches are shown.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -86,6 +86,7 @@ public class NotesActivity extends AppCompatActivity implements
     private boolean mShouldSelectNewNote;
     private boolean mIsSettingsClicked;
     private boolean mIsTabetFullscreen;
+    private boolean mIsTagSelected;
 
     private String mTabletSearchQuery;
     private UndoBarController mUndoBarController;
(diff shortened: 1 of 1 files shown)


SylvesterWilmott:
Looks good!

However, I noticed one thing:

1. Navigate to a tag from the drawer
2. Tap search
3. Clear search field
4. Perform a search
5. Notice only the current tag is still searched

There's two ways I might expect this to behave:

- Treat the field as filled if the user backspaces or empty (clean) if the user clears the field which essentially returns it to an "All notes" search
- Fix the tag in the search bar so it cannot be backspaced or cleared when searching from a tag

Thoughts on this?

theck13:
Ahh, nice catch!  That's an artifact of the way search used to work.  It would first filter the note list by whatever tag was selected then filter by whatever search text was input by the user.

Since we have the `tag:` prefix available and we are making it more visible to users with suggestions as well as prefilling the search field with that when a tag is selected, it seems like _always_ searching all notes when search is opened would be the way to go.  That way users don't have to back out of search completely to select _*All Notes*_ then go back into search to find something outside of that tag, which would be the case if we made the tag unable to be backspaced.

theck13:
We've decided to not include this aspect in the search update project.

Issue #422, opened 2017-01-24
Switch Trash Icon for Empty and Restore Actions


theck13:
### Expected
The [restore icon](https://github.com/Automattic/simplenote-android/blob/develop/Simplenote/src/main/res/drawable/ic_empty_trash_24dp.xml) is used for restoring notes and the [trash icon](https://github.com/Automattic/simplenote-android/blob/develop/Simplenote/src/main/res/drawable/ic_trash_24dp.xml) is used for emptying the trash.

### Observed
The [trash icon](https://github.com/Automattic/simplenote-android/blob/develop/Simplenote/src/main/res/drawable/ic_trash_24dp.xml) is used for restoring notes and the [restore icon](https://github.com/Automattic/simplenote-android/blob/develop/Simplenote/src/main/res/drawable/ic_empty_trash_24dp.xml) is used for emptying the trash.

![trash](https://cloud.githubusercontent.com/assets/3827611/22227785/536bde5e-e18a-11e6-9a9c-5557880613ce.png)

#### Note
Renaming the `ic_empty_trash_24dp` icon to `ic_trash_restore_24dp` might avoid confusion also.

### Reproduced
1. Open navigation drawer.
2. Tap ***All Notes*** item.
3. Tap any note in ***All Notes*** list.
4. Tap trash icon in action bar.
5. Open navigation drawer.
6. Tap ***Trash*** item.
7. Long-press restore icon.
8. Notice "Empty trash" message.
9. Tap trashed note in ***Trash*** list.
10. Long-press trash icon.
11. Notice "Restore note" message.
12. Tap trash icon.

#### Tested
Google Pixel on Android 7.1.1 with Simplenote 1.5.5

PR #914, opened 2019-12-17
change layout_weight about #422

dbgprud:
### Fix
change layout_weight
When I added a note, I didn't see the add tag, so I had to scroll down, but now I can see it.


theck13:
The changes to `fragment_note_editor.xml` may fix a certain use case for a specific device, but they break many other cases for many devices.  Also, the net changes to `fragment_note_editor.xml` are modifying the `layout_weight` on two views, but multiple attributes were reordered unnecessarily and the file does not follow the code style for XML files throughout the codebase.  Additionally, the df382e24c76b0e50345955bc3aa45fae93bb030b commit should not be bundled with this pull request since it is completely unrelated to the original changes.

PR #916, opened 2019-12-24
Update Async Tasks

theck13:
### Fix
Update the class names, variable names, and code style of all `AsyncTask`s for code style consistency.

### Test
Since these changes have no user-facing aspect, smoke testing is all that is required.  Viewing the note list, viewing the ***Edit*** and ***Preview*** tabs on a note, and viewing the tag list should be sufficient.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -1539,17 +1539,17 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
 
     private static class trashNotesTask extends AsyncTask<Void, Void, Void> {
 
-        private SoftReference<NoteListFragment> fragmentRef;
+        private SoftReference<NoteListFragment> mNoteListFragmentReference;
         private List<String> mDeletedNoteIds = new ArrayList<>();
         private SparseBooleanArray mSelectedRows = new SparseBooleanArray();
 
         private trashNotesTask(NoteListFragment context) {
-            fragmentRef = new SoftReference<>(context);
+            mNoteListFragmentReference = new SoftReference<>(context);
         }
 
         @Override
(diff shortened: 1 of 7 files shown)


Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -1487,30 +1492,27 @@ protected void onPostExecute(ObjectCursor<Note> cursor) {
...
         }
 
         @Override
         protected void onPreExecute() {
-            NoteListFragment fragment = fragmentRef.get();
-            if (fragment.getListView() != null) {
-                mSelectedRows = fragment.getListView().getCheckedItemPositions();
-            }
+            NoteListFragment fragment = mNoteListFragmentReference.get();
+            mSelectedRows = fragment.getListView().getCheckedItemPositions();
```
liangmicha: Is there a reason the null check is removed?

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -1529,38 +1531,34 @@ protected Void doInBackground(Void... args) {
...
```
liangmicha: Is there a reason the null check is removed?

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
theck13: It is unnecessary since `fragment.getListView()` is never `null`.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
theck13: It is unnecessary since `fragment.getListView()` is never `null`.

PR #922, opened 2020-01-14
Update Fastlane to 2.140.0

jkmassel:
### Fix
Fixes a warning in Github about a vulnerable Ruby dependency. This is a minor version update and shouldn't break anything with fastlane.

### Test
Ensure all automated tests pas.

### Review
Just needs a quick 👍 from one person

### Release


jkmassel:
We'll merge against develop to kill the warning, but this doesn't go out as part of a release, so you're right, 2.4 is _probably_ better!

I just threw the first milestone in the list on it so Peril wouldn't get mad at me :D

PR #946, opened 2020-02-12
Remove Passcode Logout

theck13:
### Fix
Add removing the passcode if it is set when a user logs out.  Currently, the passcode is retained after logout causing the user to enter the passcode while logging in or immediately after login.

### Test
There are two scenarios when a user would be logged out when the passcode lock is enabled; manual logout from ***Settings*** and automatic logout from a password change.
#### Logout from Settings
1. Open navigation drawer.
2. Tap ***Settings*** item in navigation drawer.
3. Tap ***Turn passcode lock on*** item under ***Passcode Lock*** section.
5. Enter passcode.
6. Enter passcode again.
7. Tap recents/overview system navigation button.
8. Wait two seconds.
9. Tap Simplenote app preview in recents/overview.
10. Notice passcode lock is shown.
11. Enter passcode.
12. Tap ***Log out*** item under ***Account*** section.
13. Notice login/signup is shown.
14. Tap recents/overview system navigation button.
15. Wait two seconds.
16. Tap Simplenote app preview in recents/overview.
17. Notice login/signup is shown.

#### Logout from Password
1. Open navigation drawer.
2. Tap ***Settings*** item in navigation drawer.
3. Tap ***Turn passcode lock on*** item under ***Passcode Lock*** section.
5. Enter passcode.
6. Enter passcode again.
7. Tap recents/overview system navigation button.
8. Wait two seconds.
9. Tap Simplenote app preview in recents/overview.
10. Notice passcode lock is shown.
11. Enter passcode.
12. Tap recents/overview system navigation button.
13. Remove Simplenote app from recents/overview.
14. Change password at https://app.simplenote.com/settings.
15. Launch Simplenote app.
16. Notice login/signup is shown.
17. Tap recents/overview system navigation button.
18. Wait two seconds.
19. Tap Simplenote app preview in recents/overview.
20. Notice login/signup is shown.
21. Log in with new password.
22. Notice note list is shown.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -1089,6 +1089,9 @@ public class NotesActivity extends AppCompatActivity implements
         editor.remove(PrefUtils.PREF_WORDPRESS_SITES);
         editor.apply();
 
+        // Remove Passcode Lock password
+        AppLockManager.getInstance().getAppLock().setPassword("");
+
         Intent intent = new Intent(NotesActivity.this, SimplenoteAuthenticationActivity.class);
         startActivityForResult(intent, Simperium.SIGNUP_SIGNIN_REQUEST);
     }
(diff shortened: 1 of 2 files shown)


theck13:
Thanks for the detailed review, animations, and steps to reproduce it!

I think that was a mistake with my test steps and a timing issue too.  The mistake was that the login/signup screen would be shown before the passcode screen.  If a passcode is set, we should expect the user to enter the passcode before seeing anything in the app.  From a security perspective, the user's data is still safe in this case.  The timing issue is that the back end might not send the login/signup reset to Android instantaneously.  It may take a few seconds for Android to receive the notice and show the login/signup screen.  In other words, I think that behavior is expected, but let me know if you think otherwise.

Issue #884, opened 2019-11-17
App Starts with White Background in Dark Theme


jivox92362:
The app initializes with a white container, brightly flashing into your eyes at night.

### Expected
The app initializes with a transparent or black container.

### Observed
The app initializes with a white container.

### Reproduced
at night, in dark mode
1. Get in line at Sparrow, LA's hottest club
2. Open Simplenote to check what you got going on tomorrow
3. The screen flashes like a camera flash
4. Girl behind you yells about you taking a photo of her
5. You tell her you were checking your schedule on Simplenote
6. She asks wtf simplo? why you takin photo of me?
7. You just say your iPhone got confused about being in dark mode for a second

BitsInMyBlood:
The application preview window background color needs to be changed to black only if dark mode is enabled, or always ?

theck13:
The startup background should match the `mainBackgroundColor` attribute (i.e. `background_dark` for Dark and ``background_light` for Light).

manu-chroma:
Hi, I'm interested in taking up this issue. 

Problem is, I'm not able to reproduce this issue on my android device. I suppose this issue present on simplenote-ios? (since the description says iPhone)

theck13:
The iPhone reference is just a joke or mistake.  This is an Android issue.  See the animation below for illustration.

<a href="https://user-images.githubusercontent.com/3827611/73598844-5e21af00-44fa-11ea-9a6a-1b22d587bb7f.gif"><img src="https://user-images.githubusercontent.com/3827611/73598844-5e21af00-44fa-11ea-9a6a-1b22d587bb7f.gif" width="300"></a>

adityabhaskar:
Hi

I'll work on this bug unless someone's already on it.

Cheers!

Adi

adityabhaskar:
I'm unable to replicate this issue. I tried on a device (Pixel with API 29) and emulators (Pixel 3XL with API 29, and Nexus 5X with API 25).

I am quite confident that this behaviour is not due to missing dark mode background resource for startup window. The theme setup is correct (relevant lines below the fold). 

My hypothesis is that the white splash is occurring because the `AppCompatDelegate.setDefaultNightMode` is called in `ThemeUtils.setTheme` from `NotesActivity.onCreate`. [The recommendation](https://medium.com/androiddevelopers/appcompat-v23-2-daynight-d10f90c83e94) is to call `setDefaultNightMode` in `Application.onCreate`.
 
[...]

PR #951, opened 2020-02-19
Issue/884 app starts with white background in dark theme

adityabhaskar:
### Fix

Potential fix for #884 .
- Extracted the call to `AppCompatDelegate.setDefaultNightMode` from `ThemeUtils.setTheme` into a new static method `ThemeUtils.setNightMode`.
- `ThemeUtils.setNightMode` is now called from the Application class `Simplenote`.

Discussed in [comment here](https://github.com/Automattic/simplenote-android/issues/884#issuecomment-588291188).

### Test

> 1. Set the app theme to `Dark`, or set it to `System default` and set the device in night mode.  
> 2. Close the app, and remove it from recent apps to force system to stop and destroy the app.  
> 3. Open the app again. 

The app should reopen with a dark splash screen before displaying contents. If the app briefly shows a white splash screen before opening in dark mode, the bug is still not fixed.

### Review
A developer is required to review the changes. 

Specifically, someone needs to confirm the placement of call to `ThemeUtils.setNightMode` in `Simplenote.onCreate`. I have placed it as early as possible—right after CrashUtils initialisation, though it may be safely moved anywhere within the `onCreate` method.

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -18,6 +18,7 @@ import com.automattic.simplenote.models.Preferences;
 import com.automattic.simplenote.models.Tag;
 import com.automattic.simplenote.utils.CrashUtils;
 import com.automattic.simplenote.utils.PrefUtils;
+import com.automattic.simplenote.utils.ThemeUtils;
 import com.simperium.Simperium;
 import com.simperium.client.Bucket;
 import com.simperium.client.BucketNameInvalid;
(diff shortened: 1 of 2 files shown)


theck13:
I'm still seeing the light screen on launch before the app is changed to the ***Dark*** theme.  Specifically, it occurs on devices and emulators below API 29 _and_ with API 29 _if Android's Dark mode is not enabled_.  Android's Dark mode appears to handle the ***Dark*** theme correctly and the screen is dark on launch _if the ***Dark*** theme is selected in Simplenote_.  If the ***Light*** theme is selected in Simplenote and Android's Dark mode is enabled on the device running API 29, the dark screen is shown on launch before the app is changed to the ***Light*** theme.  I'm seeing the same behavior before and after the changes in this pull request.

Because of that, it's debatable whether the color of the launch screen as it is right now is expected behavior or not.  That's why we should keep the `[priority] low` label on the issue.

Thank you for attempting to fix the issue!

adityabhaskar:
Thanks for the detailed description. I was able to replicate it with these.

I'll have another look at it now that I can replicate them.

Cheers!

PR #961, opened 2020-02-24
Add RTL Support

theck13:
### Fix
Add support for right-to-left languages and layouts.  Fortunately, nearly all of the layouts were already optimized for right-to-left and adding the `android:supportsRtl="true"` attribute was the only thing needed to render them correctly.  The only thing I found that needed updating was the padding between the tag chips and ***Add tag...*** field at the bottom of the note editor.  The `ChipGroup` view was not honoring the `android:paddingEnd` attribute as it should.  As a workaround, I added a `View` with the same padding for its width between the two views.  See the screenshots below for illustration.

add_rtl_support

### Test
In order to see the layouts in right-to-left, a right-to-left language (e.g. Hebrew) must be selected at the system level _or_ the ***Force RTL layout direction*** setting can be enabled in ***Developer options***.  I've seen issues with the ***Force RTL layout direction*** setting giving false positives so I recommend selected a right-to-left language.

Note that even though a right-to-left language is selected, the app layout can show incorrectly _if left-to-right text is used_.  As an example, the ***Edit tags*** screen will show text to the left even when Hebrew is selected at the system level if the tags used English text.  Using Hebrew text will show the expected layout in that case.

Even though the test steps below only address the note editor since that's the only layout that was changed, it's best to smoke test the rest of the app to ensure no other issues are found.
1. Tap any note in list with no tags.
2. If note content does not overflow screen, notice ***Add tag...*** field is shown.
3. Notice no padding between ***Add tag...*** field and edge of note content.
4. If note content does overflow screen, scroll to view ***Add tag...*** field.
5. Notice no padding between ***Add tag...*** field and edge of note content.
6. Add tag to ***Add tag...*** field.
7. Notice no padding between tag and edge of note content.
8. Notice padding between tag chip and ***Add tag...*** field.

### Review

### Release
`RELEASE-NOTES.txt` was updated in 5778e6d with:
> Added support for right-to-left languages and layouts


theck13:
> I've never run into an app where if I turn on "Force RTL" english still ends up left-aligned.

That's not something Simplenote actively does coincidentally.  I'm not sure if that's part of the updated Android versions or not, but I thought it was interesting as well.

> It looks like the edit tags list view only supports LTR

It looks that way in your screenshot, but it's actually due to all those tags being in English.  Once they are in a right-to-left language, they are right-aligned.  See the screenshot below for illustration.

add_rtl_support_tags

> typing and previewing in an RTL language such as arabic works great, but I did notice that the "add tag..." text is still left-aligned

That looks like it's due to a missing translation for Arabic, which means the English variant is used and the field is left-aligned.  If the language has a translation for that field, it will be right-aligned for right-to-left languages like Hebrew.  See the screenshot below for illustration.

add_rtl_support_add_tag

Also, notice that part of the note content is left-aligned if the line begins with a left-to-right language word and right-aligned if the line begins with a right-to-left language word.  Crazy!

AmandaRiu:
PR code approved -- waiting on design approval before merging.

iskandergaba:
> That looks like it's due to a missing translation for Arabic, which means the English variant is used and the field is left-aligned. If the language has a translation for that field, it will be right-aligned for right-to-left languages like Hebrew. See the screenshot below for illustration.

@theck13 Great work! Since you are at it, I suggest you add the translation for "Add tag" in Arabic: 
إضافة وسم

P.S: It's my native tongue. I don't know what platform you use but I thought I should share it.

theck13:
Thanks for the suggestion, @iskandergaba!

We use the [Simplenote Android project](https://translate.wordpress.com/projects/simplenote%2Fandroid) under [GlotPress](https://translate.wordpress.com) where people can suggest, discuss, and validate translations.  Regardless of their native language, I encourage anyone that would like contribute to translations to join!

iskandergaba:
Oh OK. It is already there I just checked. It could be that the string resources in the app are not up to date with the translations. Anyhow, good to know.

AmandaRiu:
@theck13 FYI - this one has some conflicts

theck13:
Thanks for the ping, @AmandaRiu!  I didn't notice the conflicts.  I merged `develop` into this branch and resolved the conflicts.  It's ready to for review again.

AmandaRiu:
Did another round of quick testing since more code was merged in. Looks good! :shipit:

PR #967, opened 2020-02-27
Search Match Bar

theck13:
### Fix
Add a view at the bottom of the note editor to go through the matching terms from a search query.  After performing a search, tapping on a note in the search results takes you to the note editor with the searched terms highlighted.  Tapping the ***Next*** and ***Previous*** chevrons in the search match bar scrolls to the position of the next and previous matching term, respectively.  See the screenshots and animation below for illustration.

search_match_bar

<a href="https://user-images.githubusercontent.com/3827611/75377302-b5fbcd80-588e-11ea-81cf-da9d414ffb88.gif"></a>

### Test
1. Tap ***Search notes and tags*** action in app bar.
2. Enter search term(s) matching at least one note.
3. Notice search term(s) are highlighted in search results.
4. Tap any note in search results.
5. Notice search term(s) are highlighted in not editor.
6. Notice search match bar shows position and total number of search match terms.
7. If total number of search match terms is less than two, notice both chevrons are disabled.
8. If total number of search match terms is more than one, notice ***Next*** chevron is enabled and ***Previous*** chevron is disabled.
9. Long-press ***Next*** chevron.
10. Notice ***Next*** message is shown.
11. Tap ***Next*** chevron.
12. Notice note editor is scrolled to position of next search match term.
13. Notice ***Previous*** chevron is enabled.
14. Long-press ***Previous*** chevron.
15. Notice ***Previous*** message is shown.
16. Tap ***Previous*** chevron.
17. Notice note editor is scroll to position of previous search match term.

### Review

### Release
`RELEASE-NOTES.txt` was updated in 4371cbb with:
> Added a toolbar to find matching search terms within notes

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
@@ -4,10 +4,14 @@ import android.app.Activity;
 import android.content.Intent;
 import android.content.res.Configuration;
 import android.os.Bundle;
+import android.os.Handler;
+import android.view.HapticFeedbackConstants;
 import android.view.View;
+import android.widget.ImageButton;
+import android.widget.RelativeLayout;
+import android.widget.Toast;
 
 import androidx.annotation.NonNull;
-import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.Toolbar;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
@@ -19,6 +23,7 @@ import com.automattic.simplenote.models.Note;
 import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.ThemeUtils;
 import com.automattic.simplenote.widgets.NoteEditorViewPager;
+import com.automattic.simplenote.widgets.RobotoMediumTextView;
 import com.google.android.material.tabs.TabLayout;
 import com.simperium.client.Bucket;
 import com.simperium.client.BucketObjectMissingException;
@@ -26,22 +31,38 @@ import com.simperium.client.BucketObjectMissingException;
 import org.wordpress.passcodelock.AppLockManager;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
 
 import static com.automattic.simplenote.analytics.AnalyticsTracker.CATEGORY_WIDGET;
 import static com.automattic.simplenote.analytics.AnalyticsTracker.Stat.NOTE_LIST_WIDGET_NOTE_TAPPED;
 import static com.automattic.simplenote.analytics.AnalyticsTracker.Stat.NOTE_WIDGET_NOTE_TAPPED;
 import static com.automattic.simplenote.utils.DisplayUtils.disableScreenshotsIfLocked;
+import static com.automattic.simplenote.utils.MatchOffsetHighlighter.MATCH_INDEX_COUNT;
+import static com.automattic.simplenote.utils.MatchOffsetHighlighter.MATCH_INDEX_START;
 import static com.automattic.simplenote.utils.WidgetUtils.KEY_LIST_WIDGET_CLICK;
 import static com.automattic.simplenote.utils.WidgetUtils.KEY_WIDGET_CLICK;
 
 public class NoteEditorActivity extends ThemedAppCompatActivity {
-    private TabLayout mTabLayout;
+    private static final String STATE_MATCHES_INDEX = "MATCHES_INDEX";
+    private static final String STATE_MATCHES_LOCATIONS = "MATCHES_LOCATIONS";
+
+    private ImageButton mButtonPrevious;
+    private ImageButton mButtonNext;
     private Note mNote;
+    private NoteEditorFragment mNoteEditorFragment;
     private NoteEditorFragmentPagerAdapter mNoteEditorFragmentPagerAdapter;
     private NoteEditorViewPager mViewPager;
+    private RelativeLayout mSearchMatchBar;
+    private RobotoMediumTextView mNumberPosition;
+    private RobotoMediumTextView mNumberTotal;
+    private String mNoteId;
+    private TabLayout mTabLayout;
     private boolean isMarkdownEnabled;
     private boolean isPreviewEnabled;
-    private String mNoteId;
+    private int[] mSearchMatchIndexes;
+    private int mSearchMatchIndex;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
@@ -59,7 +80,7 @@ public class NoteEditorActivity extends ThemedAppCompatActivity {
             getSupportActionBar().setDisplayHomeAsUpEnabled(true);
         }
 
-        NoteEditorFragment noteEditorFragment;
+        mNoteEditorFragment = new NoteEditorFragment();
         NoteMarkdownFragment noteMarkdownFragment;
 
         mNoteEditorFragmentPagerAdapter =
@@ -83,13 +104,12 @@ public class NoteEditorActivity extends ThemedAppCompatActivity {
                 arguments.putString(NoteEditorFragment.ARG_MATCH_OFFSETS,
                         intent.getStringExtra(NoteEditorFragment.ARG_MATCH_OFFSETS));
 
-            noteEditorFragment = new NoteEditorFragment();
-            noteEditorFragment.setArguments(arguments);
+            mNoteEditorFragment.setArguments(arguments);
             noteMarkdownFragment = new NoteMarkdownFragment();
             noteMarkdownFragment.setArguments(arguments);
 
             mNoteEditorFragmentPagerAdapter.addFragment(
-                    noteEditorFragment,
+                    mNoteEditorFragment,
                     getString(R.string.tab_edit)
             );
             mNoteEditorFragmentPagerAdapter.addFragment(
@@ -143,6 +163,12 @@ public class NoteEditorActivity extends ThemedAppCompatActivity {
 
             isMarkdownEnabled = savedInstanceState.getBoolean(NoteEditorFragment.ARG_MARKDOWN_ENABLED);
             isPreviewEnabled = savedInstanceState.getBoolean(NoteEditorFragment.ARG_PREVIEW_ENABLED);
+            mSearchMatchIndex = savedInstanceState.getInt(STATE_MATCHES_INDEX, 0);
+            mSearchMatchIndexes = savedInstanceState.getIntArray(STATE_MATCHES_LOCATIONS);
+        }
+
+        if (intent.hasExtra(NoteEditorFragment.ARG_MATCH_OFFSETS)) {
+            setUpSearchMatchBar(intent);
         }
 
         mViewPager.setAdapter(mNoteEditorFragmentPagerAdapter);
@@ -202,6 +228,8 @@ public class NoteEditorActivity extends ThemedAppCompatActivity {
         }
         outState.putBoolean(NoteEditorFragment.ARG_MARKDOWN_ENABLED, isMarkdownEnabled);
         outState.putBoolean(NoteEditorFragment.ARG_PREVIEW_ENABLED, isPreviewEnabled);
+        outState.putInt(STATE_MATCHES_INDEX, mSearchMatchIndex);
+        outState.putIntArray(STATE_MATCHES_LOCATIONS, mSearchMatchIndexes);
         super.onSaveInstanceState(outState);
     }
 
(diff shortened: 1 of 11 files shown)


SylvesterWilmott:
Looks great @theck13 !

A couple of small things:

- Can we use grey for the chevrons? (Same as the appbar)
- Are we able to highlight the currently selected keyword as the user cycles through them? This could be done using `blue50` for the highlight colour and `white` for the text

theck13:
> Can we use grey for the chevrons? (Same as the appbar)

Using the same color as the app bar icons makes the disabled and enabled states indistinguishable.  See the screenshots below where the chevron on the left is disabled and the right is enabled.

search_match_bar_gray

> Are we able to highlight the currently selected keyword as the user cycles through them? This could be done using the light mode highlight style on dark mode and vice versa

I tried doing this originally, but there isn't a straightforward way to do it.  I think it would be a great enhancement to build on these changes in the future though.

theck13:
I updated the chevrons to use the toolbar icon color with 100% opacity for the enabled state and 40% opacity for the disabled state in d178b68b.  See the screenshots below for illustration.

search_match_bar_gray_disabled

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
```diff
@@ -235,6 +263,114 @@ public void showTabs() {
...
+            String[] matches = matchOffsets != null ? matchOffsets.split("\\s+") : new String[]{};
+            String[] matchesStart = new String[matches.length / MATCH_INDEX_COUNT];
+
+            // Get "start" item from matches.  The format is four space-separated integers that
+            // represent the location of the match: "column token start length" ex: "1 3 3 7"
+            for (int i = MATCH_INDEX_START, j = 0; i < matches.length; i += MATCH_INDEX_COUNT, j++) {
+                matchesStart[j] = matches[i];
+            }
+
+            // Remove duplicate items with linked hash set and linked list.
```
dmsnell: what is the reason we expect duplicates?

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
```diff
@@ -235,6 +263,114 @@ public void showTabs() {
...
+        mNumberTotal = findViewById(R.id.text_total);
+        mButtonPrevious = findViewById(R.id.button_previous);
+        mButtonNext = findViewById(R.id.button_next);
+
+        mButtonPrevious.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                if (mSearchMatchIndex > 0) {
+                    mSearchMatchIndex--;
+                }
```
dmsnell: if we were to flip this condition could not skip skip all the following updates if we're at the start of the list of matches (and unnecessary)…

```java
if (mSearchMatchIndex) {
  return;
}

mSearchIndex--;

mNoteEditorFragment.…
```

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
```diff
@@ -235,6 +263,114 @@ public void showTabs() {
...
+                        @Override
+                        public void run() {
+                            updateSearchMatchBarStatus();
+                        }
+                    },
+                    getResources().getInteger(android.R.integer.config_mediumAnimTime)
+                );
+            }
+        });
+        mButtonPrevious.setOnLongClickListener(new View.OnLongClickListener() {
```
dmsnell: what's the purpose of the long-click? I expected it to go to the first result but here it looks like it's supposed to pop up the search result in a banner on screen? is that the Toast?

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
theck13: The string we get from `ARG_MATCH_OFFSETS` and full-text search can have duplicates.  One case is when the title text has a match.  An example might help.  Below is a string we get with the set of matches in the format `column token start length`.  We want to grab the `start` value from those matches to know where to scroll in the note.  You can see the first two sets are not identical, but they have the same `start` value.  That makes it seem like there are seven matches for the search term when there are actually six; the title is counted twice.

```
INDEX  00 01 02 03 | 04 05 06 07 | 08 09 10 11 | 12 13 14 15 | 16 17 18 19 | 20 21 22 23 | 24 25 26 27
MATCH   0  0  0  6 |  1  0  0  6 |  1  0 17  6 |  1  0 25  6 |  1  0 43  6 |  1  0 55  6 |  1  0 89  6
START         0    |        0    |       17    |       25    |       43    |       55    |       89 
```

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
theck13: The long-press shows a hint about the button.  It's a typical feature on Android to long-press on an icon to show text about what action that icon represents since icons can be ambiguous.  It *is* a toast, which is ephemeral lasting two seconds and doesn't block other interface actions.

When opening the note, you *are* scrolled to the first match as you expected.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
theck13: Sorry, I don't follow.

```java
if (mSearchMatchIndex) {
  return;
}
```

`mSearchMatchIndex` is an integer rather than a boolean.

dmsnell: that was meant to say `if (mSearchMatchIndex === 0) { return }`

sorry 'bout that

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
dmsnell: is this due to the `title`/`content` data structure or a regexp we're using?

I don't doubt what you are saying is true, just trying to understand _why_ it happens and maybe hint at what things would be helpful in a comment to the next dev who comes around and naively says "there shouldn't be any duplicates, let's just remove this" 🙂

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
```diff
@@ -143,6 +163,12 @@ public void onPageScrollStateChanged(int state) {
 
             isMarkdownEnabled = savedInstanceState.getBoolean(NoteEditorFragment.ARG_MARKDOWN_ENABLED);
             isPreviewEnabled = savedInstanceState.getBoolean(NoteEditorFragment.ARG_PREVIEW_ENABLED);
+            mSearchMatchIndex = savedInstanceState.getInt(STATE_MATCHES_INDEX, 0);
+            mSearchMatchIndexes = savedInstanceState.getIntArray(STATE_MATCHES_LOCATIONS);
+        }
+
+        if (intent.hasExtra(NoteEditorFragment.ARG_MATCH_OFFSETS)) {
```
dmsnell: is it right to assume that these were already present, coming from the fulltext-search performed already when searching?

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
dmsnell: makes so much more sense, thanks for clarifying

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
theck13: Ah, okay.  No worries.  That makes more sense.

Incidentally, the code below the `if (mSearchMatchIndex > 0) { mSearchMatchIndex--; }` statement is not called when `mSearchMatchIndex <= 0` since the `updateSearchMatchBarStatus()` method disables the button and the click won't be fired.

We could update that snippet to the following.  Do you think that is more readable/understandable than the current code?

```java
if (mSearchMatchIndex > 0) {
    mSearchMatchIndex--;

    mNoteEditorFragment.scrollToMatch(mSearchMatchIndexes[mSearchMatchIndex]);
    new Handler().postDelayed(
        new Runnable() {
            @Override
            public void run() {
                updateSearchMatchBarStatus();
            }
        },
        getResources().getInteger(android.R.integer.config_mediumAnimTime)
    );
}
```

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
dmsnell: > Do you think we should be explicit and say it's due to title/content and full-text search in a comment?

I've seen too many people come around and "fix" things or "improve" things by messing with stuff that's non-obvious from the code. The comment indicates that we're removing duplicates, which at worst can be inferred by the immediately-following code.

What can't be reasonably inferred is why we expect them in the first place.

Something on the order of…

```java
// due to the way that sqlite interacts with the title and content
// we can get duplicate matches from time to time and therefore
// need to remove them
```

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
dmsnell: I don't think it's more or less readable. it seems reasonable to provide this guard since we can't count on the fact that the button will be disabled. when it comes to style I'm flexible but I generally find that early-aborting and removing nesting makes the control flow simpler to follow and chart and less prone to bugs as a result of that - as a result of exiting the scope before we have more assignments/variables and conditions that we could forget to handle. it's not a blocker here and I know it could sound like I'm making a bigger deal out of this than needs be; I'm not trying to do that but rather just respond to the question "do you find this more understandable?"

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
dmsnell: yeah I was just making sure we weren't looking for this to be newly added as a part of this PR which might have been missed vs. it already existed in the codebase but we weren't using it

dmsnell:
> If we were reusing the search match bar component, I would definitely do that. Since the views are in the activity_note_editor layout, it seemed more intuitive to me to have the logic in the NoteEditorActivity class.

Re-use definitely isn't what I was thinking about when I brought this up - it's more about the design of the API we're presenting and using inside the app. for example, it looks like the `NoteEditorActivity` and `NoteEditorFragment` have to be aware of some really specific properties of how we construct the search match bar. I'm thinking about thinks like does that view really need to know about the fact that we store search results in an array and reference them by that index?

Could we instead say that the editor activity only need know a few things about the search bar:
 - Instantiate it and store it `mSearchMatchBar = …`
 - Pass the `Intent` when it comes in, at maybe just the extra string value for search
 - Provide a callback or a method to call which the search bar can call to tell the note editor to scroll to a given location

That could leave all of the logic behind things like clamping the selected index, connecting the button clicks, and parsing the result/index string into a separate place.

So it's more about code organization and separating those concerns is why I brought it up. At least from someone who is less familiar with how the app is put together and historically so all these extra additions make it harder for me to grasp the components and how they interact, but then again, I'm not maintaining it so 🤷‍♂

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
theck13: Yep, it was already there, but just not used for this purpose.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
theck13: I updated the comment in 7b6f67ad.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
theck13: In cases when multiple `if`/`else if`/`else` statements are nested, I agree that aborting early is preferable.  In this case, there is only one `if` so the cognitive load is small.  I think you make a good point about being aware of the button enabled/disabled later.  I updated the conditional as I mentioned above to make is clear when the scroll occurs and the search match bar is updated in a34fa9ab.

SylvesterWilmott:
Thanks for the changes!

I'm still thinking about showing/hiding the bar. on iOS the bar and the highlights disappear when the note is focused. Here on Android, the bar is hidden when the note is focused and the highlights are hidden once a character is typed. I think the bar and the highlights should be hidden simultaneously after a character is typed. The bar can remain behind the keyboard.

theck13:
I moved hiding the search match bar from when the editor is focused to when content is edited in f2a715ed.

SylvesterWilmott:
Looking really good.

Discussed: It's possible to tap 'through' the bar so you can focus the note by tapping on an area on the bar near the center.

theck13:
I added the `clickable` and `focusable` attributes to the search match bar to avoid being able to tap through the bar in 0f04b713.

PR #972, opened 2020-03-03
Update Title Span

theck13:
### Fix
Add removing the `StyleSpan` (i.e. `BOLD`) before reapplying the `RelativeSizeSpan` and `StyleSpan` to the note title when a note is edited.   If a span is applied repeatedly without removing the span first, performance issues occur.  While the `RelativeSizeSpan` was already being removed, the `StyleSpan` was not, which made the app lag when editing a note.

### Test
1. Tap ***New Note*** floating action button.
2. Enter more than 200 characters in note title.
3. Notice app doesn't lag.

### Review

### Release
`RELEASE-NOTES.txt` was updated in 46e2c9c5 with:
> Fixed a bug with the note title that made the app lag while editing a note

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -887,9 +887,14 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
         mContentEditText.addTextChangedListener(this);
     }
 
+    /**
+     * Set the note title to be a larger size and bold style.
+     *
+     * Remove all existing spans before applying spans or performance issues will occur.  Since both
+     * {@link RelativeSizeSpan} and {@link StyleSpan} inherit from {@link MetricAffectingSpan}, all
+     * spans are removed when {@link MetricAffectingSpan} is removed.
+     */
     private void setTitleSpan(Editable editable) {
-        // Set the note title to be a larger size and bold style.
-        // NOTE: Remove all existing spans before applying spans or performance issues will occur.
         for (MetricAffectingSpan span : editable.getSpans(0, editable.length(), MetricAffectingSpan.class)) {
             editable.removeSpan(span);
         }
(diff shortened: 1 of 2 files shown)


Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -887,15 +887,22 @@ public void onTextChanged(CharSequence charSequence, int start, int before, int
...
-        for (RelativeSizeSpan span : spans) {
+        // Set the note title to be a larger size and bold style.
+        // NOTE: Remove all existing spans before applying spans or performance issues will occur.
+        for (RelativeSizeSpan span : editable.getSpans(0, editable.length(), RelativeSizeSpan.class)) {
             editable.removeSpan(span);
         }
+
+        for (StyleSpan span : editable.getSpans(0, editable.length(), StyleSpan.class)) {
+            editable.removeSpan(span);
+        }
```
dmsnell: would it be appropriate to call `clearSpans()` here instead of iterating over all of them and removing them?
https://developer.android.com/reference/android/text/Editable#clearSpans()

```java
private void setTitleSpan(Editable editor) {
	// if we don't clear out the old spans then we will leak them
	// and our performance will suffer as the title gets longer
	// we will add back the appropriate ones below
	editable.clearSpans();

	int newLinePosition = getNoteContentString().indexOf("\n");
	…
}
```

theck13: You would think so according to the documentation, but `Editable.clearSpans()` doesn't exactly work as expected.  All spans are removed, but it seems to disable applying future spans.  Also, it moves the cursor to the first position making text input very difficult.  Finally, there are significant performance issues even with a small number of characters.  See the animation below for illustration.

<a href="https://user-images.githubusercontent.com/3827611/75818482-2f476480-5d56-11ea-864b-4c1983d7e54f.gif"></a>

theck13:
I realized that both `RelativeSizeSpan` and `StyleSpan` inherit from `MetricAffectingSpan` so we can optimize `setTitleSpan(Editable editable)` by cutting the looping in half with e0a473d.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
dmsnell: Wow. That’s surprising. I guess it is really removing more spans that aren’t obviously there but which are important for proper functionality.

PR #998, opened 2020-04-02
Update Password Policy

theck13:
### Fix
Update the simperium dependency, which includes updating the password policy as described in https://github.com/Simperium/simperium-android/pull/219.  The updated password policy requirements are the following:
- Password cannot match email
- Minimum of 8 characters
- No new lines
- No tabs

### Test
In order to test the login password length updates, change your Simplenote password to more than four and less than eight characters.

0. Clear app data.
1. Tap ***Sign Up*** button.
2. Enter email address in ***Email*** field.
3. Notice ***Sign Up*** button is disabled.
4. Enter email address, new lines, or tabs in ***Password*** field.
5. Notice ***Sign Up*** button is enabled after eight characters are input.
6. Tap ***Sign Up*** button.
7. Notice ***Error*** dialog with password requirements is shown.
8. Tap ***OK*** button in dialog.
9. Tap back arrow in top app bar.
10. Tap ***Log In*** button.
11. Tap ***Log In with Email*** button.
12. Enter email address in ***Email*** field.
13. Notice ***Log In*** button is disabled.
14. Enter password of more than four and less than eight characters in ***Password*** field.
15. Notice ***Log In*** button is enabled after four characters are input.
16. Tap ***Log In*** button.
17. Notice ***Error*** dialog with password requirements is shown.
18. Tap ***Reset*** button in dialog.
19. Notice external browser is opened showing the https://app.simplenote.com/reset/ page.

### Review

### Release

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -68,7 +68,7 @@ dependencies {
         exclude module: 'recyclerview-v7'
     }
     // Automattic and WordPress dependencies
-    implementation 'com.simperium.android:simperium:0.9.6'
+    implementation 'com.simperium.android:simperium:0.9.7'
     implementation 'com.automattic:tracks:1.2'
     implementation 'org.wordpress:passcodelock:2.0.2'
 
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -68,7 +68,7 @@ dependencies {
         exclude module: 'recyclerview-v7'
     }
     // Automattic and WordPress dependencies
-    implementation 'com.simperium.android:simperium:0.9.7'
+    implementation 'com.simperium.android:simperium:0.9.8'
     implementation 'com.automattic:tracks:1.2'
     implementation 'org.wordpress:passcodelock:2.0.2'
 
--- a/Simplenote/src/main/res/values/strings.xml
+++ b/Simplenote/src/main/res/values/strings.xml
@@ -240,6 +240,7 @@
     <string name="simperium_dialog_message_login_reset">Your password is insecure and must be reset.  The password requirements are:\n\n- Password cannot match email\n- Minimum of %1$d characters\n- No new lines\n- No tabs</string>
     <string name="simperium_dialog_message_network">There is no network available.  Please, connect to a network and try again.</string>
     <string name="simperium_dialog_message_password">The password requirements are:\n\n- Password cannot match email\n- Minimum of %1$d characters\n- No new lines\n- No tabs</string>
+    <string name="simperium_dialog_message_password_login">Passwords must be a minimum of %1$d characters excluding new lines, spaces, and tabs.</string>
     <string name="simperium_dialog_message_signup">Could not sign up with the provided email address and password.</string>
     <string name="simperium_dialog_message_signup_existing">The email address you entered is already associated with a Simplenote account.</string>
     <string name="simperium_dialog_progress_logging_in">Logging in&#8230;</string>
--- a/Simplenote/src/main/res/values/strings.xml
+++ b/Simplenote/src/main/res/values/strings.xml
@@ -248,7 +248,7 @@
     <string name="simperium_error_email">Invalid email</string>
     <string name="simperium_error_password">Invalid password</string>
     <string name="simperium_footer_login">%1$s%2$s%3$sForgot password?%4$s</string>
-    <string name="simperium_footer_login_url">https://app.simplenote.com/forgot?email=%1$s</string>
+    <string name="simperium_footer_login_url">https://app.simplenote.com/forgot/?email=%1$s</string>
     <string name="simperium_footer_signup">By creating an account, you agree to our %1$s%2$s%3$sTerms and Conditions%4$s</string>
     <string name="simperium_footer_signup_url">https://simplenote.com/terms/</string>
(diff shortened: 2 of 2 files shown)


dmsnell:
We're invaliding a password on login if it doesn't match the requirements but this has a couple problems with it:

 - there are existing passwords that don't meet the requirements
 - we force people to reset before they even make a valid login

It seems like it would be better on our part if we waited to tell them to reset their passwords until after they successfully login. If you follow the testing steps and enter the wrong password you still get the same warning and that is a bit annoying, even though you can cancel it.

I also wonder if we shouldn't give people a grace period to reset their passwords. For instance maybe we could set a flag on their account and say "you are logged in now but you will not be able to login again with this password - please check your email for a reset link or click here"

that way, if someone where doing something like starting a meeting with their phone on the projector they wouldn't be immediately interrupted and embarrassed. just a thought.

theck13:
> We're invaliding a password on login if it doesn't match the requirements

We're not invalidating the password.  We're forcing accounts with passwords that don't meet the requirements to reset the password.

> there are existing passwords that don't meet the requirements

That's expected.  Those passwords will need to be reset.

> we force people to reset before they even make a valid login

That's expected.  If they attempt to log in with an incorrect password that doesn't meet the passwords requirements, they will be asked to reset the password.  Do you think that's a vulnerability?

I don't think we should give a grace period.  Then they would never _have_ to reset their password.  The current password requirements are very weak and the security advantage outweighs the inconvenience of requiring the password reset.

theck13:
> In my testing this was not the case. For instance, if I mistype my password, or just enter `a` and try to login the app took me to the password reset page. It never attempted to check if `a` was actually my password. In other words, the app didn't care if my account had a good password or not, it rejected my password outright before asking.

These changes don't do anything to the password itself.  It forces accounts with passwords that don't meet the requirements to reset the password.  If the user's password is `1234` and they enter `1235`, they will prompted to reset their password, but they can still cancel and retry.

> People can still login with weak passwords and if we shut it down in Android I feel like we'll be causing an interruption without doing much to increase security.

If these changes get merged in the next two weeks, they will be included in the [Simplenote Android 2.6.0](https://github.com/Automattic/simplenote-android/milestone/18) freeze on April 20th so that it will coincide with the [Simplenote Electron 1.16](https://github.com/Automattic/simplenote-electron/milestone/29) freeze and [Simplenote iOS 4.19](https://github.com/Automattic/simplenote-ios/milestone/29) freeze.  That way the password policy will be released across platforms at the same time.

> It feels like jumping the gun to send someone to a password-reset page if they entered the wrong password. They might have even forgotten their password but now instead of trying it and failing they'll have multiple jumps in and out of the app and a webview.

If they have forgotten their password, then it needs to be reset anyway.  They can still tap the ***Cancel*** button to retry without resetting.

> I was thinking of something on the order of "You have 1 hour to reset your password" and that's something the backend can enforce.

That might be something we can add in a future iteration, but it would require backend updates like you mentioned as well as changes to each frontend app since they will have to differentiate between failed login due to incorrect email/password and expired grace period.

SylvesterWilmott:
>we force people to reset before they even make a valid login

This is a good point @dmsnell 

>It seems like it would be better on our part if we waited to tell them to reset their passwords until after they successfully login.

I think that would be great especially working alongside a grace period. It does feel odd to input the incorrect password and be told my password is insecure without validating the password first. 

And this will be a permanent issue, I can reset my password today and still receive a message about having an insecure password tomorrow if I input my password incorrectly (7 characters instead of 8 for example).

Can we dodge the need for the extra effort to include a grace period by validating the password first with some logic like:

- If the password is valid and doesn't meet the password requirements then show "password insecure" error dialog and prevent login

mchowning:
I'm just curious if there is a reason we don't just have the server be solely responsible for enforcing any password requirements? It could return a particular error with a relevant user message in cases where the password did not meet any of the requirements. Seems like it would be nice to not be implementing the logic to enforce our password requirements across all the different platforms.

theck13:
<blockquote>
<i><b>Dennis:</b></i> It seems like it would be better on our part if we waited to tell them to reset their passwords until after they successfully login.
<br><br>
<blockquote>
<i><b>Sylvester:</b></i> It does feel odd to input the incorrect password and be told my password is insecure without validating the password first.
</blockquote>
</blockquote>

I'll update that behavior in the Simperium library and add the changes to this pull request subsequently.

<blockquote>
<i><b>Matt:</b></i> I'm just curious, is there is a reason we don't just have the server be solely responsible for enforcing any password requirements? It could return a particular error with a relevant user message in cases where the password did not meet any of the requirements. It would be nice to not be implementing the logic to enforce our password requirements separately on each platform.
</blockquote>

I agree.  Ideally, we wouldn't have to make any changes to the client apps and the server would handle everything.  Unfortunately, we can't do that for a few reasons.  All old app versions will not be able to handle the new password policy requirements set by the server and will result in uninformative or incorrect error messages if at all.  Also, the messages sent by the server are not translated and would result in poor internationalization.  Plus, the password policy changes will not be released at the exact same time across all platforms and will cause issues if we make changes to the server without all platforms being able to handle those changes.

theck13:
The updates have been merged in the Simperium library and added here.  The changes are ready for another review!

PR #1006, opened 2020-04-22
Revert Websocket Threading

theck13:
### Fix
Update the `simperium` dependency from `0.9.8` to `0.9.9`, which includes reverting the `AsyncWebSocketProvider` changes in https://github.com/Automattic/simplenote-android/pull/927 to avoid data loss as described in https://github.com/Simperium/simperium-android/pull/221.

### Test
Since these changes have no user-facing aspect, smoke testing is all that is required.  The changes affect the networking protocol class so it's best to test creating and editing notes and tags to ensure syncing is working as expected across devices.  Using two emulators side-by-side is a good way to see if syncing is working.

### Review

### Release
`RELEASE-NOTES.txt` was updated in 9898a01 with:
> Fixed networking bug causing data loss

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 2.6
 -----
- 
+* Fixed networking bug causing data loss
+
 2.5
 -----
 * Added a dark version of the note widget to view and open a note in the app from the home screen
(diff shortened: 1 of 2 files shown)


theck13:
> This may be related to the issue I opened earlier #1007.

That's possible, because I can't reproduce that out-of-sync behavior.  Since these changes are just reverted another change, I'm confident they aren't creating a new sync issue that is most likely unrelated.

PR #1010, opened 2020-05-01
Update Color 2.3.0

theck13:
### Fix
Update the color scheme to [Color Studio v2.3.0](https://color-studio.blog/).  These changes include updating color value of `gray_30` and all shades of the `yellow` palette.  There are no functional changes.  See the screenshots below for illustration.

update_color_230

### Test
An alpha build can be downloaded once the continuous integration checks pass.  It will be installed alongside the production build (i.e. two Simplenote apps will be on the device; one production and one debug) so changes can be compared on the same device.

### Review
Only one developer and one designer are required to review these changes, but anyone can perform the review.  The code review should not worry about the color names or specific changes to the color of the interface, but it should check that there are no breaking changes.  The interface and color values should be reviewed by a designer.

### Release


SylvesterWilmott:
Looks good!

It seems the blues in search results keyword highlighting weren't updated though.

theck13:
Good catch, @SylvesterWilmott!  I updated those references in ea99f167.

PR #1027, opened 2020-05-12
Add encrypted credentials for upcoming screenshot user

mokagio:
### Fix
Adds the credentials to use for the user that will drive the screenshot automation via `.configure`.

### Test
1. `bundle install && bundle exec fastlane run configure_apply`
1. Open Android Studio and make the project
1. Verify that `SCREENSHOT_USER_EMAIL` and `SCREENSHOT_USER_PASSWORD` are present in `BuildConfig.java`

### Review
Only one developer is required to review these changes, but anyone can perform the review. Thanks.

### Release


Simplenote/build.gradle
```diff
@@ -150,7 +150,9 @@ android.applicationVariants.all { variant ->
             key == "LOGIN_EMAIL" ||
             key == "LOGIN_PASSWORD" ||
             key == "SENTRY_DSN" ||
-            key == "WPCOM_CLIENT_ID") {
+            key == "WPCOM_CLIENT_ID" ||
+            key == "SCREENSHOT_USER_EMAIL" ||
+            key == "SCREENSHOT_USER_PASSWORD") {
```
mokagio, line 155: @theck13 I've added the values like this to keep everything in line with the current setup, but what do you think of adopting the same strategy as [WooCommerce does](https://github.com/woocommerce/woocommerce-android/blob/6f9bde468dd8d9bca2cf59eec3f2745e8ce06ce3/WooCommerce/build.gradle#L263-L276). That is, just pipe through all the values in `gradle.properties` (optionally matching a certain pattern).

If you're up for that, I'll open a PR for it later on.

theck13: Is there any disadvantage to using the current setup?

theck13, line 155:
```suggestion
            key == "SCREENSHOT_EMAIL" ||
            key == "SCREENSHOT_PASSWORD") {
```

Is `_USER` necessary?  We could remove it to make the variables `SCREENSHOT_EMAIL` and `SCREENSHOT_PASSWORD` to mimic the `LOGIN_EMAIL` and `LOGIN_PASSWORD` variables.

mokagio: Only the minor annoyance of having two files to change, `gradle.properties` and this `build.gradle`, rather than one.

Coming from the WooCommerce codebase, it took me a while to figure out why the values didn't appear in the generated `BuildConfig.java` automatically. But, that could be just my little familiarity with Android.

mokagio: Yeah, will do 👍 .

theck13:
@rachelmcr, would you like to review this pull request also or is it fine for me to merge it?

PR #1026, opened 2020-05-12
Fastlane setup for local screenshot automation

mokagio:
### Fix
Sets up the components required for running the screenshots automation locally, that is integrates `screengrab` in the projects and adds a new lane to the `Fastfile` for it.

### Test
1. Launch an Emulator (I've been using a Pixel 2 XL with API 29)
1. If you already have Simplenote on the Emulator, please logout. The full automation will take care of this.
1. Run `bundle install && bundle exec fastlane take_screenshots`

When everything is done, your browser should open rendering the content of `fastlane/screenshots/screenshots.html`, which is a screenshot of the login screen.

Screen Shot 2020-05-12 at 10 08 05 pm

The next step will be adding all of the code to the screenshot test to take the ones we need.

### Review
Only one developer is required to review these changes, but anyone can perform the review. Thanks.

### Release
These changes do not require release notes.

PR #1038, opened 2020-05-21
Add logic to automate all required screenshots

mokagio:
### What it does
This adds to the already existing test to take the screenshots all the logic to take the actual screenshots we want for the Play Store.

### Test
With the same setup as #1026, run `bundle exec fastlane take_screenshots`.

The browser should open showing screenshots like these

screenshot-2020-05-23-204820

### Review
I'd like to know if I've committed any major Java sin. It had been almost a decade since I'd written any.

I'm not happy with the fact that all the code is in a single file, but at the same time I can't think of any neat way to separate all this that doesn't require adding more abstractions. So, given that this code -in this current iteration- will run only for the occasional screenshots and smoke testing, I think having it all in the same file is a reasonable compromise.

### Release

### DIFF
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
@@ -164,7 +164,7 @@ public class ScreenshotTest {
 
     // Flows
 
-    private void logoutIfNeeded() throws InterruptedException {
+    private void logoutIfNeeded() {
         if (isViewDisplayed(getViewById(R.id.list_root)) == false) {
             return;
         }
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
@@ -154,7 +154,7 @@ public class ScreenshotTest {
         loadPasscodeUnsetterFromSettings();
         typeFullPasscode();
         // Disable darkmode
-        disableDarkModeFromSettings();
+        enableLightModeFromSettings();
     }
 
     @AfterClass
@@ -329,7 +329,7 @@ public class ScreenshotTest {
         selectSettingsOption(R.string.theme, themePosition);
     }
 
-    private void disableDarkModeFromSettings() {
+    private void enableLightModeFromSettings() {
         loadThemeSwitcherFromSettings();
         // The options have no id, and I couldn't find a way to access them by their text
         onView(childAtPosition(withId(R.id.select_dialog_listview), 0)).perform(click());
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
@@ -68,7 +68,6 @@ import static tools.fastlane.screengrab.cleanstatusbar.IconVisibility.SHOW;
 @LargeTest
 @RunWith(AndroidJUnit4.class)
 public class ScreenshotTest {
-    private static final String SIMPLENOTE_PACKAGE = "com.automattic.simplenote.debug";
     private static final int LAUNCH_TIMEOUT = 5000;
     private static final String NOTE_TITLE = "# Lemon Cake & Blueberry";
 
@@ -229,14 +228,14 @@ public class ScreenshotTest {
 
         // Launch the app
         Context context = ApplicationProvider.getApplicationContext();
-        final Intent intent = context.getPackageManager()
-                .getLaunchIntentForPackage(SIMPLENOTE_PACKAGE);
+        String packageName = context.getPackageName();
+        final Intent intent = context.getPackageManager().getLaunchIntentForPackage(packageName);
         // Clear out any previous instances
         intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK);
         context.startActivity(intent);
 
         // Wait for the app to appear
-        device.wait(Until.hasObject(By.pkg(SIMPLENOTE_PACKAGE).depth(0)), LAUNCH_TIMEOUT);
+        device.wait(Until.hasObject(By.pkg(packageName).depth(0)), LAUNCH_TIMEOUT);
     }
 
     // Notes List Screen
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
@@ -1,58 +1,527 @@
 package com.automattic.simplenote.screenshots;
 
-
+import android.content.Context;
+import android.content.Intent;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewParent;
 
+import androidx.appcompat.widget.SearchView;
+import androidx.test.core.app.ApplicationProvider;
+import androidx.test.espresso.PerformException;
+import androidx.test.espresso.UiController;
+import androidx.test.espresso.ViewAction;
 import androidx.test.espresso.ViewInteraction;
+import androidx.test.espresso.action.ViewActions;
+import androidx.test.espresso.assertion.ViewAssertions;
+import androidx.test.espresso.contrib.RecyclerViewActions;
 import androidx.test.espresso.matcher.ViewMatchers;
+import androidx.test.espresso.util.HumanReadables;
+import androidx.test.espresso.util.TreeIterables;
 import androidx.test.filters.LargeTest;
+import androidx.test.platform.app.InstrumentationRegistry;
 import androidx.test.rule.ActivityTestRule;
 import androidx.test.runner.AndroidJUnit4;
+import androidx.test.uiautomator.By;
+import androidx.test.uiautomator.UiDevice;
+import androidx.test.uiautomator.Until;
 
+import com.automattic.simplenote.BuildConfig;
 import com.automattic.simplenote.NotesActivity;
 import com.automattic.simplenote.R;
+import com.google.android.material.textfield.TextInputLayout;
 
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeMatcher;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
+import java.util.concurrent.TimeoutException;
+
 import tools.fastlane.screengrab.Screengrab;
+import tools.fastlane.screengrab.UiAutomatorScreenshotStrategy;
+import tools.fastlane.screengrab.cleanstatusbar.CleanStatusBar;
 import tools.fastlane.screengrab.locale.LocaleTestRule;
 
 import static androidx.test.espresso.Espresso.onView;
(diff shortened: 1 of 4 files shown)


Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -1,58 +1,528 @@
...
+    public void screenshotTest() throws InterruptedException {
+        // Pre-checks if the state is dirty
+        enterThenDisablePasscodeIfNeeded();
+        logoutIfNeeded();
+
+        login();
+
+        // Wait for notes to load.
+        // TODO: this should be some kind of loop/polling code, not a dumb and fragile sleep
+        Thread.sleep(5000);
```
mokagio, line 103: Should be able to use

```java
waitForViewMatching(allOf(withId(R.id.note_title), withText(NOTE_TITLE)), 1000);
```

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -1,58 +1,528 @@
...
+        // Swipe to perform a scroll (because I couldn't get a reference to a scrollable view)
+        // that will reveal the logout button.
+        onView(withId(R.id.preferences_container)).perform(swipeUp());
+
+        // Logout
+        selectSettingsOption(R.string.log_out, logoutPosition);
+
+        // Give time to the logout to finish
+        // TODO: this should be some kind of loop/polling code, not a dumb and fragile sleep
+        Thread.sleep(3000);
```
mokagio, line 180: Given we have logic to properly wait for the notes list, this could be removed.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -1,58 +1,528 @@
...
+        getViewById(R.id.input_password)
+                .perform(click())
+                .perform(replaceTextInCustomInput(BuildConfig.SCREENSHOT_PASSWORD))
+                .perform(ViewActions.closeSoftKeyboard());
+
+        getViewById(R.id.button).perform(click());
+
+        // This waits for the notes container to load, that is the login has been successful.
+        // We still have to wait for the notes to load from the backend, though.
+        waitForViewToBeDisplayed(R.id.list, 10000);
```
mokagio, line 214: We can remove this, too. Or, use the proper wait code.

fastlane/Fastfile
```diff
@@ -337,6 +337,7 @@ SUPPORTED_LOCALES = [
       # Because the screenshot tests are together with other UI tests, we need
       # to specify to run only them and not all of the others.
       use_tests_in_classes: "com.automattic.simplenote.screenshots.ScreenshotTest",
+      use_timestamp_suffix: false,
```
mokagio, line 340: This way, when / if we'll run the composition, our code will be able to uniquely identify each screenshot.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -1,58 +1,527 @@
...
+
+        login();
+
+        // Wait for notes to load.
+        // Not 100% sure why, but without this little sleep call, the call to wait for the note that
+        // we need to interact with to be loaded in the list fails. I suspect it might be due to the
+        // app changing root activity (from login to notes list) but I haven't had the time to
+        // research it properly
+        Thread.sleep(500);
+        waitForViewMatching(allOf(withId(R.id.note_title), withText(NOTE_TITLE)), 5000);
```
mokagio, line 107: Keen to hear why this might be the case.

mokagio:
Not sure about `aapt` either 🤔 .

I think the automation is failing because of this, though

```
[07:36:57]: ▸ com.automattic.simplenote.screenshots.ScreenshotTest:

[07:36:57]: ▸ Error in 
    screenshotTest(com.automattic.simplenote.screenshots.ScreenshotTest):

[07:36:57]: ▸ androidx.test.espresso.NoMatchingViewException: No views in 
    hierarchy found matching: (with id: 
	com.automattic.simplenote.debug:id/recycler_view and has descendant: with 
	string from resource id: <2131820758>[passcode_turn_on] value:
    Turn passcode lock on)
```

mokagio:
@theck13 I tried to reproduce a few times, with no luck.

From reading the stack trace (thanks again for adding it) the error happens the when the tests try to select the settings option to turn on the passcode.

Here's how they run on my machine (_big GIF, might take a while to load_)



How do yours look? Is there any chance you could record them for me if they fail consistently? Thanks.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -1,58 +1,527 @@
...
+                )
+        )
+                .perform(RecyclerViewActions.actionOnItemAtPosition(position, click()));
+    }
+
+    private void loadThemeSwitcherFromSettings() {
+        selectSettingsOption(R.string.theme, themePosition);
+    }
+
+    private void disableDarkModeFromSettings() {
```
theck13, line 333:
```suggestion
    private void enableLightModeFromSettings() {
```

This was confusing at first since disabling **_Dark_** is the same as enabling **_Light_**.  What do you think about changing the name of this method to mimic the similar `enableDarkModeFromSettings` method?

mokagio:
Bumping milestone to 2.8 because this tooling work is not required for 2.7, which we're about to code freeze.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -1,58 +1,527 @@
...
+    }
+
+    @AfterClass
+    public static void afterAll() {
+        CleanStatusBar.disable();
+    }
+
+    // Flows
+
+    private void logoutIfNeeded() throws InterruptedException {
```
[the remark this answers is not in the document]
mokagio: Sure! It must have ended up there at some point as I added a `Thread.timeout()`. I wonder why Android Studio doesn't tell me about this unnecessary one.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
mokagio: Oh yeah! Totally. Brain fart.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -1,58 +1,527 @@
...
+    @BeforeClass
+    public static void beforeAll() {
+        Screengrab.setDefaultScreenshotStrategy(new UiAutomatorScreenshotStrategy());
+
+        new CleanStatusBar()
+                .setBatteryLevel(100)
+                .setClock("1231")
+                .setWifiVisibility(SHOW)
+                .setBarsMode(TRANSPARENT)
+                .enable();
```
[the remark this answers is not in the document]
mokagio: It's hit and miss when I run the tests from Android Studio itself. But I consistently get it set when running via `bundle exec fastlane` 🤔

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -1,58 +1,527 @@
...
+
+    @AfterClass
+    public static void afterAll() {
+        CleanStatusBar.disable();
+    }
+
+    // Flows
+
+    private void logoutIfNeeded() throws InterruptedException {
+        if (isViewDisplayed(getViewById(R.id.list_root)) == false) {
```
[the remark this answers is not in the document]
mokagio: Sure. Is this the convention you use in the codebase?

For context. I tend to use this style because I find it clearer to read. I sometimes miss the `!` at the start, just a single thin character... the more verbose `== false` seems clearer.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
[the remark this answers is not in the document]
mokagio: Actually, it did tell me 😳 

Screen Shot 2020-06-01 at 9 30 00

😅

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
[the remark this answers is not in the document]
mokagio: Oh... I just saw that Android Studio offers that as an option. Cool, I guess that's just the expected style here. 

Applying 🔨

mokagio:
Thanks for the tips @theck13 👍 I applied them.

> We can skip the logout step and avoid any issues with the passcode lock by running this adb command at the start to clear the app data.
>
> `adb shell pm clear com.automattic.simplenote.debug`

Oh nice. I didn't know how to clear the app data via `adb`.

If it's ok with you, I'd like to keep the code. When working on these, my workflow is to first get them to work in Android Studio, and then make sure they work on Fastlane. So, that code to reset the state at the start and end is very handy because it spares me from manually resetting it every time.

Or, do you know a way that I can use the run configuration or something similar to reset the app every time I run the tests from Android Studio?

> Overall, the code feels pretty fragile since there are multiple instances of hard-coded numbers and strings as well as arbitrary wait times

Yeah, totally 😞 . @rachelmcr is working on adding more automation for smoke testing purposes to the apps. Hopefully, when it comes the turn of this one, we'll be able to remove some of this dirty brittleness.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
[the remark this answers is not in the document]
theck13: I'm unable to get those values to be set with Android Studio or via a terminal with the `bundle exec fastlane take_screenshots` command.  It's not a showstopper for this pull request, but it would be great if we can get it to work so we can modify the mobile network, wifi connection, battery level, and clock time.

mokagio:
@theck13 yep, we can run `adb` commands from Fastlane. I'll update the PR to do that because always having a clean slate would make things faster anyways and prevent errors as you showed me 👍 . Thanks.

I'd like to keep the code that handles with state cleanup, though, because it's helpful to me (and I think anyone else who'll need to modify these tests) when running the code via Android Studio. 

This is the kind of workflow I use when working on UI tests:

1. Write the tests on the IDE
2. Run it from the IDE, see what failures I get (it never works the first time does it?)
3. Make the tests work from the IDE
4. Make the tests work from the automation runner. In this case, Fastlane

It's handy to work within the IDE first, because it gives clearer error messages, and you can set breakpoints to try out the different UI queries to drive the tests.

If I removed the state cleanup code from my test, when running them from Android Studio, they'd fail after the first launch because of the state the app is in, e.g. already logged in when the test expects a login button.

I could circumvent this by manually resetting the state using `adb` or deleting the app from the emulator, but this would make my feedback loop slower, because I'd have to leave the IDE, run some commands, and go back, as opposed to just run the tests and let them do their thing.

If you don't think this is ok here, I can extract the logic a bit, so that it's there for me to call if I'll need it next time, but is not in the code that gets executed.

mokagio:
Thanks @theck13. I'll be keeping an eye on those issue, too 👀

PR #1041, opened 2020-05-26
Add Automation Test Flows

stavares843:
### Test
1. Go to `espressoUITests` folder
2. Right click and select on `Run tests in ..` button

Screenshot 2020-05-28 at 20 53 29

`espressoUITests`

added tests:

`AuthenticationTest.java` - includes the following automated flows:

- login with the correct password and logout
- trying to login with wrong credentials and verify the error alert message
- trying to login with wrong credentials and then login with correct credentials

`NoteActivityTest` - includes the following automated flows:

- adds note > delete > undo deletion > deletes note from all notes list > logout
- adds note > goes to note detail > delete > logout
- adds note > delete note > trash > empty trash > verify trash is empty > logout

`NoteEditorOptionsTest`- includes the following automated flows:
- add checklist on note 

Helpers:
`espressoUITestsHelpers`

- `EspressoAuthenticationHelpers` - includes actions used basically on sign up, login and such
- `EspressoGeneralHelpers` - includes general actions such as custom swipes
- `EspressoNoteActivityHelpers` - includes actions/flows used on notes activity such as add note, delete note, etc
- `EspressoNoteEditorHelpers` - includes actions from the popupwindows on note editor

Tests should be running properly and with a green status

Screenshot 2020-05-28 at 20 53 20

### Review
This PR adds the automated tests described above.

This also adds a Espresso dependency - https://github.com/AdevintaSpain/Barista which allows to have a list of common actions such as `openDrawer` without having to add them manually one by one what makes the process of adding automated tests even faster

Note - using sleep until adding a better improvement


Simplenote/src/androidTest/java/com/automattic/simplenote/LoginWithCorrectPasswordLogoutTest.java
```diff
@@ -0,0 +1,182 @@
...
+    public ActivityTestRule<NotesActivity> mActivityTestRule = new ActivityTestRule<>(NotesActivity.class);
+
+    @Test
+    public void loginWithCorrectPasswordLogoutTest() {
+
+        try {
+            sleep(5000);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
```
rachelmcr, line 52: I see a number of `sleep`s like this in the tests. Is there a specific condition the test is waiting for that we could make explicit here?

Simplenote/src/androidTest/java/com/automattic/simplenote/LoginWithCorrectPasswordLogoutTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+        //login with email and logout
+
+        ViewInteraction loginButton = onView(
+                allOf(withId(R.id.button_login), withText("Log In"),
```
rachelmcr, line 57: When possible, let's identify elements with just the ID so the tests can be run in any device locale (with localized strings).

Simplenote/src/androidTest/java/com/automattic/simplenote/LoginWithCorrectPasswordLogoutTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+        MotionEvent event = MotionEvent.obtain(downTime, eventTime, MotionEvent.ACTION_DOWN, 500, start, 0);
+        inst.sendPointerSync(event);
+        eventTime = SystemClock.uptimeMillis() + delay;
+        event = MotionEvent.obtain(downTime, eventTime, MotionEvent.ACTION_MOVE, 500, end, 0);
+        inst.sendPointerSync(event);
+        event = MotionEvent.obtain(downTime, eventTime, MotionEvent.ACTION_UP, 500, end, 0);
+        inst.sendPointerSync(event);
+        SystemClock.sleep(2000);
+    }
```
rachelmcr, line 157: Could you tell me more about this method and the benefit of using `swipeToBottom()` in this test, versus something like scrolling to a specific element that you want to interact with?

stavares843:
hello @rachelmcr, thanks for reviewing! 😄 

updated the PR based on your feedback, seems way better now and with less code as well 

modifications:
- added a file named `EspressoUITests.java` with basic actions such as login, logout, swipes, and others - all actions that can be reused on other tests without duplicating code and with better maintenance as well
- added a file named `SignUpWithCredentials.java` that includes the sign up automated flow 
- added a file named `Login.java` that includes login with the correct password and logout and trying to login with the wrong password and verify the error alert message


TODO: investigating atm the better way to avoid using sleep such as using espresso idling resources

please let me know your feedback, thanks 😄

Simplenote/src/androidTest/java/com/automattic/simplenote/LoginWithCorrectPasswordLogoutTest.java
stavares843: investigating atm the better way to avoid using sleep such as using espresso idling resources 😄

Simplenote/src/androidTest/java/com/automattic/simplenote/LoginWithCorrectPasswordLogoutTest.java
stavares843: updated, thanks! 😄

Simplenote/src/androidTest/java/com/automattic/simplenote/LoginWithCorrectPasswordLogoutTest.java
stavares843: I updated this with more swipe methods such as 

`swipeToBottom` - Swipe all the way to the bottom of the screen
`scrollSlowlyDown`- Swipe down one page at a time
`swipeToTop` - Swipe to the top
`scrollSlowlyUp`- Swipe up one page at a time


Those methods are nice to have a variety of swipe actions on automated tests, but we could also scroll automatically to the specific element

Also, was thinking instead of scrolling to the specific element to update the scroll to the `scrollSlowlyUp` where it scrolls slowly, I think would be more similar to a mainly user behaviour, any thoughts?

rachelmcr: > Also, was thinking instead of scrolling to the specific element to update the scroll to the scrollSlowlyUp where it scrolls slowly, I think would be more similar to a mainly user behaviour, because when going to the settings view, user doesn't see the logout option right away, I think it's a more natural user behaviour slowly scrolling until find it, any thoughts?

That's an interesting thought! What are the pros/cons you see for using a more natural scroll behavior in these tests? Are there specific risks you are trying to address with that?

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/EspressoUITests.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Rule
+    public ActivityTestRule<NotesActivity> mActivityTestRule = new ActivityTestRule<>(NotesActivity.class);
+
+    @Test
+    public static void loginWithCredentials() throws InterruptedException {
+        clickOn(R.id.button_login);
+        clickOn(R.id.button_email);
+        writeTo(R.id.input_email, "test.espresso.00001@gmail.com");
+        writeTo(R.id.input_password, "*-Re7]J4Ux8q)g?X");
```
rachelmcr, line 32: We should try to avoid putting valid credentials directly in the tests; could you move these to the `build.gradle` file so they can be added individually in `Simplenote/gradle.properties`? (Similar to other keys and credentials used in the app.)

Simplenote/src/androidTest/java/com/automattic/simplenote/Login.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Test
+    public void loginWithInvalidCredentials() throws InterruptedException {
+        clickOn(R.id.button_login);
+        clickOn(R.id.button_email);
+        writeTo(R.id.input_email, "test.espresso.00005@gmail.com");
+        writeTo(R.id.input_password, "*-Re7]J4Ux8q)g?X");
+        clickOn(R.id.button);
+        sleep(2000);
```
rachelmcr, line 39: I noticed this is almost the same set of steps used in `loginWithCredentials()`. What do you think of writing a method that can handle both valid and invalid login credentials? That way if the login flow changes, there are fewer places to update the tests.

stavares843: improved:


before:
Screenshot 2020-05-27 at 17 33 58

after:
Screenshot 2020-05-27 at 17 33 01

added a test only inserting the wrong credentials - `loginWithInvalidCredentials()` on `EspressoUITests.java`, that way we can also do more combinations such as try to login with wrong credentials and then use the correct ones and others

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/EspressoUITests.java
stavares843: updated, thanks! 😄

Simplenote/src/androidTest/java/com/automattic/simplenote/LoginWithCorrectPasswordLogoutTest.java
stavares843: The pros would be get the more similar to the user and also try to catch any crashes within the thread while scrolling, I noticed on previous projects, if the thread was still being loaded, scrolling slowly would crash (also reproducible manually) but scrolling automatically to the specific touch point wouldn't crash on the automated test.

stavares843:
updates since last review
- moved test user credentials
- improved login tests
- added a file named `DeleteNotes.java` that includes the following automated flows:
    adds note > delete > undo deletion > deletes note from all notes list > logout
    adds note > goes to note detail > delete > logout
- updated `EspressoUITests.java` with more actions

TODO: handle not synced dialog when trying to logout, to have only 1 logout test instead of 2

tried to address the sleep but wasn't working properly, but I added a post on p2 with all the info, I think it has useful stuff to be implemented on the future

any question let me know 😄

stavares843:
update
- added empty trash automated test flow

stavares843:
update
- added login with valid credentials after using wrong credentials automated test flow

stavares843:
update

- added add checklist on note automated test flow

stavares843:
update 

- added checklist on note automated test flow
- updated structure from `EspressoUITests` into smaller files
- updates naming from tests
- updated folders structure adding `espressoUITests` folder with all the espresso tests and a `espressoUITestHelpers` with all the helpers

any question let me know 😄

updated the PR description as well 😄

Simplenote/src/androidTest/java/com/automattic/simplenote/Login.java
rachelmcr: Thanks for that side by side comparison! What about writing a single `login()` method that takes the credentials as parameters, so you only need one method to try different combinations?

stavares843: updated, thanks! 😄

stavares843:
Screenshot 2020-05-30 at 04 01 38

stavares843:
Screenshot 2020-05-31 at 04 33 04

PR #1042, opened 2020-05-27
Add App Links

theck13:
### Fix
Add app links so that the password reset redirect `simplenote://launch` will open the Simplenote Android app without a disambiguation dialog requiring the user to choose the Simplenote Android app on their device.  This is the Android counterpart of https://github.com/Simperium/simplenote-gae/pull/292.

Note that the only functional changes are [adding an intent filter to the `NotesActivity` manifest element](https://github.com/Automattic/simplenote-android/compare/add-app-links?expand=1#diff-b6ee350ff52f9168d7b4f2cfb18f4368R67-R81), [updating the reset URL string resource](https://github.com/Automattic/simplenote-android/compare/add-app-links?expand=1#diff-19bb70917c478f507b793b466bd7d82eR239), and [updating the Simperium dependency version](https://github.com/Automattic/simplenote-android/compare/add-app-links?expand=1#diff-6d2d0e51a5d5de6aeee1e7014d80fce5R80).  The other changes to the `AndroidManifest.xml` file are to cleanup/organize the code and not actually remove any functionality.

### Test
These steps will open the Simplenote Android app automatically if only one instance of the app is installed on the device.  If the device has both debug and release versions of the app, a disambiguation dialog will be shown after Step 26 below.

0. Clear app data.
1. Tap ***Log In*** button.
2. Tap ***Log In with Email*** button.
3. Enter email address in ***Email*** field.
4. Notice ***Log In*** button is disabled.
5. Enter incorrect password of more than four and less than eight characters in ***Password*** field.
6. Notice ***Log In*** button is enabled after four characters are entered.
7. Tap ***Log In*** button.
8. Notice ***Error*** dialog without password requirements is shown.
9. Tap ***OK*** button in dialog.
10. Enter correct password of more than four and less than eight characters in ***Password*** field.
11. Notice ***Log In*** button is enabled after four characters are input.
12. Tap ***Log In*** button.
13. Notice ***Error*** dialog with password requirements is shown.
14. Tap ***Reset*** button in dialog.
15. Notice browser is opened showing ***Reset your password*** page with email address in ***Email*** field.
16. Tap ***Send reset email*** button.
17. Notice ***Okay! A link to reset your password will be sent if that account exists.*** message is shown.
18. Open email app.
19. Notice ***Simplenote Password Reset*** email received.
20. Open ***Simplenote Password Reset*** email.
21. Notice password reset link.
22. Notice ***?redirect=simplenote://launch*** appended to password reset link.
23. Tap password reset link.
24. Notice browser is opened showing ***Reset your password*** page with email address in ***Email*** field.
25. Enter password of more than seven characters in ***Password*** field.
26. Tap ***Change my Password*** button.
27. Notice app is opened showing ***Log In*** screen with email address in ***Email** field.
28. Notice ***Log In*** button is disabled.
29. Enter updated password from Step 25.
30. Notice ***Log In*** button is enabled after four characters are input.
31. Tap ***Log In*** button.
32. Notice app is successfully logged in.

### Review

### Release

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -77,7 +77,7 @@ dependencies {
     // Fastlane screengrab for screenshots automation
     androidTestImplementation 'tools.fastlane:screengrab:2.0.0'
     // Automattic and WordPress dependencies
-    implementation 'com.simperium.android:simperium:0.9.10'
+    implementation 'com.simperium.android:simperium:0.9.11'
     implementation 'com.automattic:tracks:1.2'
     implementation 'org.wordpress:passcodelock:2.0.2'
 
--- a/Simplenote/src/main/AndroidManifest.xml
+++ b/Simplenote/src/main/AndroidManifest.xml
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+
+<manifest
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     package="com.automattic.simplenote"
     android:installLocation="auto">
@@ -15,23 +17,29 @@
         android:label="@string/app_name"
         android:supportsRtl="true"
         android:theme="@style/Theme.Simplestyle.Splash">
+
         <activity
             android:name=".NotesActivity"
             android:configChanges="screenSize|smallestScreenSize|orientation|screenLayout"
             android:label="@string/app_launcher_name"
             android:resizeableActivity="true"
             android:windowSoftInputMode="adjustResize|stateHidden">
+
             <intent-filter>
+
                 <action android:name="android.intent.action.MAIN" />
 
                 <category android:name="android.intent.category.LAUNCHER" />
+
             </intent-filter>
 
             <meta-data
                 android:name="android.app.shortcuts"
-                android:resource="@xml/shortcuts" />
+                android:resource="@xml/shortcuts">
+            </meta-data>
 
             <intent-filter>
+
                 <action android:name="android.intent.action.SEND" />
                 <action android:name="com.google.android.gm.action.AUTO_SEND" />
                 <action android:name="com.google.android.voicesearch.action.AUTO_SEND" />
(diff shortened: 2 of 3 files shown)


theck13:
> If the user has logged into the web version of Simplenote on the device then the web version of the app will load to the notes list once the user clicks on the **RESET** button:

Yeah, that is an unfortunate side effect and known issue.  The web app automatically redirects the [reset URL](https://github.com/Automattic/simplenote-android/pull/1042/files#diff-19bb70917c478f507b793b466bd7d82eR239) to the base URL if the user is logged in.  We'll have to change some things on the web side.  I'll check with the team on the best way to do that.

> Also, under the above scenario if I close the browser and reopen the app I get a flash of the notes list before the sign in page is displayed.

That might be a consequence of the app architecture in regards to authentication.  The authentication screen is shown when the [`NotesAcitivity` detects the user authentication is invalid](https://github.com/Automattic/simplenote-android/blob/218435f8f4e545c61a6aae572afe063c70351931/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java#L241-L242).  Thus, the `NotesActivity`, which contains the note list you see flash on the screen, is always launched at least momentarily.  In your animation, it looks like the note list only contains the welcome note and not notes from a specific account.  Is that correct?  If that's the case, it _is_ an odd user experience, but it wouldn't  be considered a data breach.  What do you think?

These changes don't modify that behavior.  It is already present in `develop`.  The major functional part of these changes is [adding the `intent-filter` to `NotesActivity` in `AndroidManifest`](https://github.com/Automattic/simplenote-android/commit/0d5383a47d1169963116dddc05c33a54f0fbc8ca#diff-b6ee350ff52f9168d7b4f2cfb18f4368) in order to catch the `https://app.simplenote.com/account/.*/reset?redirect=simplenote://launch` URL.

AmandaRiu:
>  In your animation, it looks like the note list only contains the welcome note and not notes from a specific account. Is that correct? If that's the case, it is an odd user experience, but it wouldn't be considered a data breach. What do you think?

This is correct and it definitely wouldn't be a data breach, just odd. I'm find with leaving it as is though since that scenario would be an edge case very rarely seen (if ever).

PR #1060, opened 2020-06-05
Issue/1057 tag sync reverted

### LINKED ISSUE #1057: Sync: Remote changes to tags are always reverted
### Expected
I expect that my most recent changes to the tags on a note will always be saved and synced, whether the change was made locally or on another device.

### Observed
Opening a note in the Android app's note editor always syncs the local tags to other devices, even if tags were added/removed on another device more recently.

### Reproduced

1. Create a note in the Android app.
2. Add a tag to the note.
3. With the note open in the Android app, open the note on another device (I tested with both web and iOS clients) and confirm the tag is visible.
4. On the other device, add a tag.
5. Notice that after the note syncs, the newly added tag is removed.
6. In the Android app, go back to the note list.
7. On the other device, add the new tag again and notice that it stays added (and syncs to other non-Android clients).
8. In the Android app, open the note again. Notice that the new tag isn't visible in the Android app and, after the note syncs, it is removed again on the other device.

You can also repeat the steps, removing a tag instead of adding one, with the same results.

-|-|-|-
Motorola|moto e5 play|8.1.0|2.7-rc-1

Tested using Simplenote iOS 4.19 and Simplenote Electron 1.16.0-2078 (on macOS/Chrome) as the other device.

theck13:
### Fix
Add setting the notes bucket member variable when a change is detected and showing the markdown action (i.e. eye icon) or tabs (i.e. **_Edit_** and **_Preview_**) to close #1057.  See the animation below for illustration.

1057_tag_sync_reverted

The high-resolution source video used to create the animation can be seen [here](https://drive.google.com/file/d/1Jj9Q2lgok9TYchQ9H84AgQsaNy_U7MKP/view).

The animation and video show that the tags are not synced perfectly on Simplenote iOS.  I've notified @jleandroperez and he'll look into that separately.  If it requires changes to Simplenote Android, those will be in a subsequent pull request.

### Test
As shown in the animations above, the best way to test these changes is to have an Android device side-by-side with a device on another platform.
1. ***Android***: Tap any note in list.
2. ***Other***: Select note from Step 1 in list.
3. ***Android***: Add `android` tag to note.
4. ***Other***: Notice `android` tag in note.
5. ***Other***: Add `other` tag to note.
6. ***Android***: Notice `other` tag in note.
7. ***Android***: Remove `other` tag from note.
8. ***Other***: Notice `other` tag removed from note.
9. ***Other***: Remove `android` tag from note.
10. ***Android***: Notice `android` tag removed from note.

### Review
Only one developer is required to review these changes, but anyone can perform the review.  @rachelmcr, I also requested you as a reviewer since you caught the associated bug and to ensure it is fixed.

### Release

#### Note
@mokagio, these changes will require a new release candidate build once they are merged.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -716,7 +716,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
         editor.apply();
     }
 
-    protected void setMarkdownEnabled(boolean enabled) {
+    private void setMarkdownEnabled(boolean enabled) {
         mIsMarkdownEnabled = enabled;
 
         if (mIsMarkdownEnabled) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -1096,7 +1096,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
         mContentEditText.setText(content);
     }
 
-    protected void saveNote() {
+    private void saveNote() {
         try {
             if (mNote == null || mNotesBucket == null || mContentEditText == null || mIsLoadingNote ||
                 (mHistoryBottomSheet != null && mHistoryBottomSheet.getDialog() != null && mHistoryBottomSheet.getDialog().isShowing())) {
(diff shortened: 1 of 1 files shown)


mokagio:
> these changes will require a new RC

Noted, thanks. I'm subscribed to this 👍

rachelmcr:
> The animation and video show that the tags are not synced perfectly on Simplenote iOS. I've notified @jleandroperez and he'll look into that separately. If it requires changes to Simplenote Android, those will be in a subsequent pull request.

It looks like that's the Simplenote iOS issue reported in https://github.com/Automattic/simplenote-ios/issues/738 (which occurs when the tags field has focus on iOS).

PR #1082, opened 2020-06-29
Add california privacy

theck13:
### Fix
Add the **_Privacy Notice for California Users_** item to the **_About_** screen.  See the screenshots below for illustration.

add_california_privacy

### Test
1. Open navigation drawer.
2. Tap ***Settings*** item in drawer.
3. Tap ***About Simplenote*** item under ***More Information*** section.
4. Notice ***Privacy Notice for California Users*** link below ***Privacy Policy*** and ***Terms of Service*** links.
5. Tap ***Privacy Notice for California Users*** link.
6. Notice https://automattic.com/privacy/#california-consumer-privacy-act-ccpa site is shown if web browser does exist on device.
7. Notice "Unable to open website. No browser available." message is shown if web browser does not exist on device.

### Review
Only one developer is required to review these changes, but anyone can perform the review.

### Release
These changes do not require release notes.

#### Note
@mokagio, this requires a hotfix.

PR #1083, opened 2020-06-30
Merge hotfix release 2.7.1 into develop

mokagio:
See #1082.

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -30,9 +30,9 @@ android {
         if (project.hasProperty("versionName")) {
             versionName project.property("versionName")
         } else {
-            versionName "2.7"
+            versionName "2.7.1"
         }
-        versionCode 106
+        versionCode 107
         minSdkVersion 23
         targetSdkVersion 29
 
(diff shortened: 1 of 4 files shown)


mokagio:
I opened this while following our internal process documentation, but I do wonder whether it's needed at all?

This hotfix cherry picks #1082 from `develop`. Therefore, the only change that would be merged would be the change in the version number, but since we already cut `release/2.8` and merged it back into `develop`, #1080, there's no change to merge.

Am I seeing this right @loremattei? I'll leave this as a draft in the meantime.

mokagio:
I just realized the need to bump `versionCode`, see [`50e5b5d`](https://github.com/Automattic/simplenote-android/pull/1083/commits/50e5b5d586efb4bddbe8ced75eb35bdca8207e5c).

I guess in this case it _does_ make sense to merge into `master`, `develop`, and `release/2.8`, in order to have the latest `versionCode` value 🤔 . But then, which `versionName` should be used?

mokagio:
Closing in favor of #1086 .

PR #1092, opened 2020-07-09
Update Rename Tag

theck13:
### Fix
Update the entity identification value and user interface for renaming tags.  A hashed value based on the tag name is used to uniquely identify like tags to avoid duplication.  The **_Rename Tag_** dialog interface has been updated to include error messages when a tag name is invalid (i.e. contains a space or is empty).  The **_Error_** dialog is shown when a tag was unable to be renamed for an unknown reason rather than failing silently like the current implementation.  See the screenshots below for illustration.

update_rename_tag

### Test
1. Open navigation drawer.
2. Tap ***EDIT*** button in ***Tags*** header.
3. Tap any tag in list.
4. Notice ***Rename Tag*** dialog is shown.
5. Add space to tag name.
6. Notice ***Tags cannot contain spaces*** error message is shown.
7. Notice ***Save*** button is disabled in dialog.
8. Delete space from tag name.
9. Notice ***Tags cannot contain spaces*** error message is hidden.
10. Notice ***Save*** button is enabled in dialog.
11. Add any alphanumeric, special, or emoji character to tag name.
12. Tap ***Save*** button in ***Rename Tag*** dialog.
13. Notice tag is renamed.
14. Tap back arrow in app bar.
15. Close navigation drawer.
16. Tap note in list with renamed tag from Step 3.
17. Notice tag is renamed.

### Review

### Release
`RELEASE-NOTES.txt` was updated in 2b75529b with:
> Fixed a bug with duplicate tags and special characters in tag names


Simplenote/src/main/java/com/automattic/simplenote/TagsListFragment.java
```diff
@@ -403,38 +403,13 @@ public void onClick(View view) {
...
-                        // Do nothing
-                    }
-                });
-                alert.show();
+                //noinspection unchecked
+                TagDialogFragment dialog = new TagDialogFragment(
+                    ((Bucket.ObjectCursor<Tag>) getItem(getAdapterPosition())).getObject(),
+                    mNotesBucket,
+                    mTagsBucket
+                );
```
dmsnell, line 411: moving this out into its own class 👍 👍

Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param name      {@link String} to hash as the tag kay.
+     *
+     * @return          {@link String} hashed to use as tag key.
+     */
+    public static String hashTag(String name) {
+        try {
+            String normalized = Normalizer.normalize(name, Normalizer.Form.NFC);
+            String lowercased = normalized.toLowerCase(Locale.US);
+            String encoded = URLEncoder.encode(lowercased, StandardCharsets.UTF_8.toString());
+            return encoded.replace("*", "%2A").replace("+", "%20");
```
dmsnell, line 64: this is where we could catch tag names that encode too long…

```java
public static String hashTag(String name) throws TagNameTooLongException {
	…
	String encoded = URLEncoder
		.encode(lowercased, StandardCharsets.UTF_8.toString())
		.replace("*", "%2A")
		.replace("+", "%20");

	// or whatever our length limit is
	// note that our ASCII-encoded value will contain a 1:1 mapping
	// between characters and UTF-16 code units, as reported by `length()`
	if ( encoded.length() > 256 ) {
		throw new TagNameTooLongException();
	}

	return encoded;
```

theck13: That sounds like a good idea to add a check for encoded length.  If we add it to the `hashTag` method, then we'll need to add a `try`/`catch` block to everywhere `hashTag` is used, which is currently twice in the `Tag` class and twice in the `TagUtils` class.  That's not necessarily a big deal, but might make things confusing when we add some kind of error message to the interface notifying the user that the tag length is too long.

What if we add another method like `hashTagValid` to the `TagUtils` class that does the same thing `hashTag` does, but it returns a boolean if the length is under 256?  We could add that check to the **_Save_** button on the **_Rename Tag_** dialog and the tag creation detection in the **_Add tag..._** field of the editor.  I think we could display an error more gracefully in those cases.

theck13: This is what I was thinking for the interface.

Rename Tag Dialog| Add Tag Editor
-|-
<a href="https://user-images.githubusercontent.com/3827611/87092289-680db380-c1f8-11ea-96c1-e7db0fd22af5.gif"></a>|<a href="https://user-images.githubusercontent.com/3827611/87092297-6a700d80-c1f8-11ea-9890-e45a77d46e46.gif"></a>

dmsnell: > What if we add another method like hashTagValid to the TagUtils class that does the same thing hashTag does, but it returns true if the length is less than or equal to 256 and false otherwise?

Seems reasonable to me. It's just a tradeoff between having the compiler catch the bug vs. relying on us to manually do it.

Suggesting that someone contact support makes me nervous, like it could trigger unwanted and unnecessary support volume. I don't know better what to do other than eliminate the constraint the `tagHash` imposes.

Could we be a little fun with it in an informative way?

> Tag length too long (hard to explain); please choose a shorter name or choose something else.

theck13: In most cases, I wouldn't suggest contacting support immediately.  Also, it's not likely users will run into the issue where they are using a few "characters" and the encoded length is too long.  I'm a little tentative using humor in situations that could be very frustrating for users.  In the example above, it looks like four "characters" to the user and the tag is too long.  They may think you can only have tags with four characters, which is ludicrous.  That may exacerbate the frustration of an error.  That's why I usually go with a dry, straightforward error message.  What about something like this?

> The tag length is too long.  Please, try again with a shorter name.  If the problem continues, contact us at support@simplenote.com for help.

PR #1098, opened 2020-07-15
Update Tag Submission

theck13:
### Fix
Update the tag submission in the **_Add tag..._** field by adding detection of the **_Enter_**/**_Return_** key.  Research shows that users expect to tap the **_Space_** key or **_Enter_**/**_Return_** key to save a tag.

### Test
1. Tap any note in list.
2. Enter text in **_Add tag..._** field.
3. Tap **_Space_** key.
4. Notice tag is saved with text from Step 2.
5. Enter text in **_Add tag..._** field.
6. Tap **_Enter_**/**_Return_** key.
7. Notice tag is saved with text from Step 5.
8. Enter `🏴󠁧󠁢󠁥󠁮󠁧󠁿🏴󠁧󠁢󠁥󠁮󠁧󠁿🏴󠁧󠁢󠁥󠁮󠁧󠁿🏴󠁧󠁢󠁥󠁮󠁧󠁿` in **_Add tag..._** field.
9. Tap **_Space_** key.
10. Notice **_Error_** dialog is shown.
11. Notice space is trimmed from **_Add tag..._** field.
12. Tap **_Enter_**/**_Return_** key.
13. Notice **_Error_** dialog is shown.
14. Notice newline is trimmed from **_Add tag..._** field.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/TagsMultiAutoCompleteTextView.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/TagsMultiAutoCompleteTextView.java
@@ -41,11 +41,7 @@ public class TagsMultiAutoCompleteTextView extends AppCompatMultiAutoCompleteTex
         @Override
         public void onTextChanged(CharSequence s, int start, int before, int count) {
             if (count >= 1 && s.charAt(start) == SPACE) {
-                if (TagUtils.hashTagValid(s.toString())) {
-                    notifyTagsChanged();
-                } else {
-                    showDialogErrorLength();
-                }
+                saveTagOrShowError(s.toString());
             }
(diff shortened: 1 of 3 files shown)


theck13:
Ah, I should have been clearer about which flag to use in the tests.  Specifically, the flag of England (`🏴󠁧󠁢󠁥󠁮󠁧󠁿`) has is a longer URI-encoded string than other emojis.  That's why only four consecutive England flag emojis causes the error to be displayed since it is longer than the allowed 256 encoded characters.  It looks like you're using the flag of Faroe Islands (`🇫🇴`) in your animation.  Is that right?  Admittedly, those flags are very similar.  Do you get the error message when using four England flags?

Issue #70, opened 2013-07-28
Tag counts are incorrect for tags


beaucollins:
Add a tag to the note and the tag count doesn't update correctly. Happens with both a new note and an existing note.

PR #1112, opened 2020-07-29
Issue/1111 rename tag duplicate

### LINKED ISSUE #1111: Renaming tag to existing tag on the same note results in wrong notes count and duplicate tags on note
### Expected

When I rename a tag to an existing tag, I expect the tag to be de-duped everywhere. (No duplicate tags or notes count.)

### Observed

When I rename a tag to an existing tag, it is de-duped correctly in the tags list (a duplicate tag isn't created). However, if both tags were used on the same note, the notes count for the tag is erroneously increased by 1 and that tag appears twice on that note.

### Reproduced

1. Create a new note.
2. Add two tags to the note (e.g. `tag1` and `tag2`).
3. Open the navigation drawer.
4. Tap "Edit" next to "Tags."
5. Tap one of the tags you just added (e.g. `tag2`).
6. Rename that tag to the other tag on the note (e.g. rename `tag2` to `tag1`).
7. Notice that a duplicate tag is not created, but the notes count for the other tag (e.g. `tag1`) increases by 1 even though that note was already included in the count.
8. Open the note again and notice it now has two identical tags (e.g. `tag1` and `tag1`).

New note with tags|Tags list|Tags list after rename|Duplicate tags in note
-|-|-|-
Screenshot_20200729-124514|Screenshot_20200729-124521|Screenshot_20200729-124530|Screenshot_20200729-124540

Device Make: Motorola
Device Model: moto e5 play
Android Version: 8.1.0
Simplenote Version: 2.9-rc-1

theck13:
### Fix
Update the `Tag.renameTo` method to avoid duplicate tags when renaming a tag to an existing tag name to close #1111.

### Test
To ensure these changes do not affect the new tag key hash algorithm, performing the tag renaming steps below with both alphanumeric and special characters is best.  The steps are identical except for the tag names.
#### Alphanumeric
1. Tap any note in list.
2. Enter `tag1` in **_Add tag..._** field.
3. Tap **_Space_** or **_Enter_**/**_Return_** key.
4. Notice `tag1` is added to note.
5. Enter `tag2` in Add tag... field.
6. Tap **_Space_** or **_Enter_**/**_Return_** key.
7. Notice `tag2` is added to note.
8. Tap back arrow in app bar.
9. Open navigation drawer.
10. Tap **_Edit_** button in **_Tags_** header.
11. Notice `tag1` and `tag2` are in tag list with `1` next to **_Delete Tag_** button.
12. Tap `tag2` in tag list.
13. Notice **_Rename Tag_** dialog is shown.
14. Replace `tag2` with `tag1` in **_Tag name_** field.
15. Tap **_Save_** button in **_Rename Tag_** dialog.
16. Notice `tag1` is in tag list with `1` next to **_Delete Tag_** button.
17. Notice `tag2` is not in tag list.
18. Tap back arrow in app bar.
19. Tap note in list from Step 1.
20. Notice `tag1` is added to note.
21. Notice `tag2` is not added to note.
#### Special
1. Tap any note in list.
2. Enter `tag!` in **_Add tag..._** field.
3. Tap **_Space_** or **_Enter_**/**_Return_** key.
4. Notice `tag!` is added to note.
5. Enter `tag?` in Add tag... field.
6. Tap **_Space_** or **_Enter_**/**_Return_** key.
7. Notice `tag?` is added to note.
8. Tap back arrow in app bar.
9. Open navigation drawer.
10. Tap **_Edit_** button in **_Tags_** header.
11. Notice `tag!` and `tag?` are in tag list with `1` next to **_Delete Tag_** button.
12. Tap `tag?` in tag list.
13. Notice **_Rename Tag_** dialog is shown.
14. Replace `tag?` with `tag!` in **_Tag name_** field.
15. Tap **_Save_** button in **_Rename Tag_** dialog.
16. Notice `tag!` is in tag list with `1` next to **_Delete Tag_** button.
17. Notice `tag?` is not in tag list.
18. Tap back arrow in app bar.
19. Tap note in list from Step 1.
20. Notice `tag!` is added to note.
21. Notice `tag?` is not added to note.

### Review
@rachelmcr, I also requested you as a reviewer since you caught the associated bug and to ensure it is fixed.

### Release


Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
```diff
@@ -92,10 +92,11 @@ public void renameTo(String tagOld, String tagNew, int index, Bucket<Note> notes
 
                 // iterate and do a case insensitive comparison on each tag
                 for (String tag : tags) {
-                    // if it's this tag, add the new tag
-                    if (getSimperiumKey().equals(TagUtils.hashTag(tag))) {
+                    // Add new tag if it's this tag and note doesn't already have same tag.
+                    if (getSimperiumKey().equals(TagUtils.hashTag(tag)) && !tags.contains(tagNew)) {
```
dmsnell: we'll need to hash-compare against `tags` too won't we?

rename `cat` to `dog` but a note already has `Dog` as a tag…I think we'll still get a duplicate

```java
String thisHash = getSimperiumKey();

…
List<String> tags = new ArrayList<String>(note.getTags());
List<String> tagHashes = new ArrayList<String>();
Set<String> tagSet = new HashSet<String>();
for (String tag : tags) {
	String hash = TagUtils.hashTag(tag);
	tagHashes.add(hash);
	tagSet.add(hash);
}

List<String> newTags = new ArrayList<String>();
for (int i = 0; i < tags.size(); i++) {
	if (thisHash.equals(tagHashes[i]) && !tagSet.contains(tagHashes[i])) {
		newTags.add(newTag);
	} else {
		newTags.add(tags[i]);
	}
}
```

theck13: Yes, we will.  Good catch!  I updated the tag comparison in 8306736.

That case led me to find other bugs with the `Tag.renameTo` method.  We also need to add a condition when creating a new tag from the new name only if the hashed tag doesn't already exist (55ddf40).

Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
```diff
@@ -78,28 +80,43 @@ public void setIndex(Integer tagIndex) {
...
+
+                // Go through all note's tags and compare hashed tags.
+                for (int i = 0; i < tags.size(); i++) {
+                    // Add new tag if it's this tag and note doesn't already have same tag.
+                    if (tagHash.equals(tagHashes.get(i)) && !tagSet.contains(tagHashes.get(i))) {
                         newTags.add(tagNew);
-                    } else {
-                        newTags.add(tag);
+                    // Add tag if it's not old tag to be deleted.
+                    } else if (!tags.get(i).equals(tagOld)) {
```
dmsnell: I was confused at the `else if` until I realized that the condition should always hold since if `tags.get(i).equals(tagOld)` would be previously trapped in the first condition: if the tag name is identical to the old one, the one we're renaming, the `this` one, then the tag hashes would be equal as well

which seems to imply this second condition isn't needed and a basic `else` will do unless I'm missing something. if I were to come to this code and see the second condition I'd hunt to try and find an explanation for why it's there since it seems to be redundant with the first.

theck13: It might be best illustrated by an example.

A note has `Tag1` and `tag2`.  `tag2` is renamed to `tag1`.  If the branch in question is just `else`, then `tag2` will be added with `newTags.add(tags.get(i));` causing the old tag to persist in the note.

Unfortunately, I think I found another bug with the current code.

A note has `tag2` and `Tag1` already exists.  `tag2` is renamed to `tag1`.  The note has `tag2` removed, but neither `Tag1` nor `tag1` is added.

dmsnell: thanks for helping me understand. I think this is complicated by the fact that we haven't codified what happens when we rename into an existing tag name.

> If the branch in question is just else, then tag2 will be added with `newTags.add(tags.get(i));`

in this case wouldn't `tag2` hash-compare equal with `tagHash`? I assume that at the beginning of the function we're setting the hash based on the value of the tag before the rename

```java
String tagHash = getSimperiumKey(); // hash( tag2 ) = "tag2"
List<String> tags = new ArrayList<>(["Tag1", "tag2"]);
List<String> tagHashes = new ArrayList<>(["tag1", "tag2"]);
Set<String> tagSet = new HashSet<>(["tag1", "tag2"]);

for ( "tag2" ) {
	// if (tagHash.equals(tagHashes.get(i)) && !tagSet.contains(tagHashes.get(i)))
	if ( "tag2".equals( "tag2" ) && !tagSet.contains("tag2") ) {
		// doesn't follow this branch because `tag2` is already in the tags…
	// } else if ( !tags.get(i).equals(tagOld) ) {
	} else if ( ! "tag2".equals( "tag2" ) ) { // could have textual differences, e.g. "Tag2" vs. "tag2"
}
```

well yeah I see your point now.

for reference, here's how I'm doing it in the Electron rewrite…

```ts
// tag bucket responds to tag rename…
const oldHash = tagHash(oldName);
const newHash = tagHash(newName);

const prevTag = tagBucket.get( oldHash ) || {};
tagBucket.update( newHash, { ...prevTag, name: newName } );

if ( oldHash !== newHash ) {
	tagBucket.delete( oldHash );
}

// note bucket responds to tag rename…
for (const note of notes) {
	const index = note.tags.findIndex( tagName => tagHash( tagName ) === oldHash );
	if ( -1 === index ) {
		continue; // note doesn't contain renamed tag
	}

	note.tags[ index ] = newName;
	note.save()
}
```

so unlike the code here I'm not addressing duplicates if they exist and the Electron code might add `tag1` even if the note already contains `Tag1`

> A note has tag2 and Tag1 already exists. tag2 is renamed to tag1. The note has tag2 removed, but neither Tag1 nor tag1 is added.

apart from the fact that it seems questionable to allow renaming _into_ an existing tag, what if we tried this…

```java
String oldHash = TagUtils.hashTag(oldTag));
String newHash = TagUtils.hashTag(newTag));

List<String> tags = new ArrayList<>(note.getTags());
Set<String> tagSet = new HashSet<String>(tags.size());
List<String> newTags = new ArrayList<>(tags.size());

// start by appending the new tag name to the list of tags
tags.add(tagNew);

for (String tag : tags) {
	String hash = TagUtils.hashTag(tag);

	if (tagSet.contains(hash)) {
		continue; // skip any tag that's already in the list / already handled
	}

	tagSet.add(hash); // mark the tag as handled - some variation is now in the new list of tags

	if (!oldHash.equals(hash)) {
		newTags.add(tag); // this is not the tag we're renaming, so add it to the new list
		continue;
	}

	// this is the tag that was renamed…
	if (!tagSet.contains(newHash) {
		// only add the renamed version if it's not already present
		tagSet.add(newHash);
		newTags.add(newTag);
		continue;
	}
}
```

dmsnell: the bug in this I think is from the `tagSet.add(hash)` line which should only occur when adding the existing tag to the new tag list. in my tests this failed when renaming a variation of a tag and was fixed by the move

```java
String oldHash = TagUtils.hashTag(oldTag));
String newHash = TagUtils.hashTag(newTag));

List<String> tags = new ArrayList<>(note.getTags());
Set<String> tagSet = new HashSet<String>(tags.size());
List<String> newTags = new ArrayList<>(tags.size());

// start by appending the new tag name to the list of tags
tags.add(tagNew);

for (String tag : tags) {
	String hash = TagUtils.hashTag(tag);

	if (tagSet.contains(hash)) {
		continue; // skip any tag that's already in the list / already handled
	}

	if (!oldHash.equals(hash)) {
		tagSet.add(hash); // mark the tag as handled - some variation is now in the new list of tags
		newTags.add(tag); // this is not the tag we're renaming, so add it to the new list
		continue;
	}

	// this is the tag that was renamed…
	if (!tagSet.contains(newHash) {
		// only add the renamed version if it's not already present
		tagSet.add(newHash);
		newTags.add(newTag);
		continue;
	}
}
```

AmandaRiu (approved):
@theck13 I tested through all the steps and everything worked fine for me. I haven't merged however as @dmsnell has some open feedback items so I want to make sure all has been addressed.

theck13:
Thanks for the review, @AmandaRiu!  We've been having discussions about how to handle the issues @dmsnell mentioned and we haven't come to a conclusion yet.  I'll comment once we decide what to do.

theck13:
After some internal discussions and deciding to use the canonical representation of a tag (https://github.com/Automattic/simplenote-android/pull/1192), this pull request is ready for review again.

The changes include displaying a dialog when trying to rename a tag something for which a canonical representation already exists for the lexical variation entered.  As an example, imagine the tags "cat" and "dog" exist.  If you try to rename "dog" to a lexical variation of "cat" like "Cat" or "CAT" for instance, a **_Tag Conflict_** dialog will be shown asking to merge the renaming tag (i.e. "dog") into the canonical representation (i.e. "cat") or cancel the rename.  See the screenshots below for illustration.

1111_rename_tag_duplicate

@AmandaRiu, let me know if you would like to continue reviewing this pull request or if you would prefer I request another developer.  @SylvesterWilmott, I requested your review and added the `[status] needs design review` label since we made some changes to the user experience.

SylvesterWilmott:
I think the wording of the Tag Conflict dialog can be cut down. If I am attempting to rename "Dog" to "Cat" and "Cat" already exists, do you think it would make sense to completely simplify the copy to:

**Tag Conflict**
"The tag "Cat" already exists. Would you like to merge these tags?"
[Cancel][Merge]

theck13:
Shortening the message sounds good to me.  My only concern is the "these tags" part.  It should be obvious what that means, but we may want to be more explicit and specify which tag will be merged (i.e. replaced by) which tag to avoid confusion.  What about this?

<a href="https://user-images.githubusercontent.com/3827611/99676336-99637f00-2a35-11eb-93ac-e80f6ab864d9.png"></a>

theck13:
I updated the **_Tag Conflict_** dialog message in e9b49e26.  See the screenshot below for illustration.

<a href="https://user-images.githubusercontent.com/3827611/99676336-99637f00-2a35-11eb-93ac-e80f6ab864d9.png"></a>

@AmandaRiu, let me know if you would like to review the code or request another developer.

AmandaRiu:
Picking it up now @theck13

AmandaRiu (changes requested):
Hi @theck13 👋 Code changes look good overall, but found the following while testing: 

[This video](https://cloudup.com/c18ugm7pEhz) demonstrates the following items found during testing:
- When merging one tag into another the total count of items does not update to reflect the combined total. I tried restarting the app to see if the total items would update on loading, but they do not. 
- In the case of lexical variations in naming, the dialog uses the original tag case, but once merged, the tag takes on whatever changes were made in the request to rename the tag. For example, I have a tag named “Today” and another tag named “Yesterday”. I try to rename “Yesterday” to “tODAY” and the dialog will use “Today” in the text. I click to merge the two tags, but now the existing tag is “tODAY”. 

I’m not sure if this is by design, but renaming a tag lexically identical to it’s old name gives the same “merge” dialog message. Click on a tag to open the rename dialog, change nothing and click save. If nothing has changed I recommend we just cancel the rename dialog. WDYT?

theck13:
> When merging one tag into another the total count of items does not update to reflect the combined total. I tried restarting the app to see if the total items would update on loading, but they do not.

That may be related to an [old issue about incorrect tag counts](https://github.com/Automattic/simplenote-android/issues/70).  I'll have to test that some more and reopen that issue if it persists.

> In the case of lexical variations in naming, the dialog uses the original tag case, but once merged, the tag takes on whatever changes were made in the request to rename the tag.

I can't reproduce this.  I tried using your "Yesterday" and "Today" example, but it didn't change to the "tODAY" tag like you mentioned.  Here are the steps I took:

0. Create tags ***Yesterday*** and ***Today*** in a note.
1. Open navigation drawer.
3. Tap ***Yesterday*** tag.
4. Enter ***tODAY*** in ***Tag name*** field.
5. Tap. ***Save*** button in dialog.
6. Notice ***Tag Conflict*** dialog referencing ***Yesterday*** and ***Today*** tags.
7. Tap ***Merge*** button in dialog.
8. Notice ***Today*** tag is in list.
9. Notice ***Yesterday*** tag is not in list.

See the animation below for illustration.

<kbd><a href="https://user-images.githubusercontent.com/3827611/100144005-65cc8e80-2e53-11eb-9954-6948fa4a8687.gif"></a></kbd>

> If nothing has changed I recommend we just cancel the rename dialog.

I agree.  I updated the code to follow that logic in 555e7072.

AmandaRiu:
@theck13 

> That may be related to an old issue about incorrect tag counts. I'll have to test that some more and reopen that issue if it persists.

This is definitely still happening for me: 



> I can't reproduce this. I tried using your "Yesterday" and "Today" example, but it didn't change to the "tODAY" tag like you mentioned. Here are the steps I took:

Definitely still happening for me: 



I'm using an emulator and running API 29. What about you?

theck13:
<blockquote><blockquote>
That may be related to an old issue about incorrect tag counts. I'll have to test that some more and reopen that issue if it persists.
</blockquote>
This is definitely still happening for me:
</blockquote>

Okay.  I'll work on that bug next.

<blockquote><blockquote>
I can't reproduce this. I tried using your "Yesterday" and "Today" example, but it didn't change to the "tODAY" tag like you mentioned. Here are the steps I took:
</blockquote>
I'm using an emulator and running API 29. What about you?
</blockquote>

I'm using the stock emulator with a Pixel 2 API 29.  Maybe there is something else going on with tag syncing and the **_Yesterday_** and **_Today_** tags are having issues with your account.  Can you test a different set of tags like this?

1. Start with **_cat_** and _**dog**_ tags.
2. Rename **_cat_** to **_DoG_**.

AmandaRiu:
> Maybe there is something else going on with tag syncing and the Yesterday and Today tags are having issues with your account. Can you test a different set of tags like this? 

Same thing happened: 



I was curious to see if the new "DoG" tag synced to the Simplenote desktop app and discovered something interesting. The tag in the bar is labeled "DoG": 

Screen Shot 2020-11-27 at 7 06 29 PM

But if I click on the note that was originally assigned the "dog" tag, the tag is still in the proper case. I verified and it's identical on the device:

Desktop App | Mobile App
-- | --
Screen Shot 2020-11-27 at 7 06 44 PM|Screenshot_1606522075

But the document where the original tag of "cat" was renamed to "DoG" shows the mixed case version on both platforms:

Desktop App | Mobile App
-- | --
Screen Shot 2020-11-27 at 7 07 10 PM|Screenshot_1606522066

theck13:
Thanks for all the details, @AmandaRiu!  I was finally able to reproduce the behavior you described.  I was testing with both existing tags in the same note.  Once I separated the existing tags into two different notes, I saw the issue.

I updated the logic of the `Tag.renameTo` method in 4493407.  Along with fixing the issue, I tried to simplify the algorithm to make it more readable.  Let me know if it's not better than it was.

I noticed the tag conflict dialog was showing when changing the case of a tag like renaming "dog" to "Dog" for example.  I added a condition to skip showing the tag conflict dialog when renaming a tag to a lexical variation in 42ef4a1.

The pull request is ready for another review.  Thanks for the thorough testing!

AmandaRiu (approved):
Changes look good! :shipit:

PR #1114, opened 2020-07-31
Issue/1113 font resize swipe

### LINKED ISSUE #1113: Font Size Glitch in Non-Germanic Languages Switching From Preview to Edit
When the device's language is set to a language like Russian or a Turkish a glitch will show when you open a note that's in preview mode, then move to edit mode. 

### Expected
When I swipe between Edit and Preview modes the font should stay the same size through the whole process

### Observed
When the font on my device is set to a Cyrillic language or Türkçe language I see the font in the edit area momentarily larger when moving from Preview to Edit mode

### Reproduced
1. Update the language settings on your device so the desired language is the primary one on the device
2. Open a note and switch to the note preview
3. Open another note
4. Open the note in preview mode and swipe back to edit mode

Video of the issue can be seen in the ticket reporting it: #3162051-zen. I was also able to reproduce:

Device Make: Google
Device Model: Pixel 3
Android Version: 10.0.0
Simplenote Version: 2.9.0

theck13:
### Fix
Add setting the text size of the note editor when the screen is created to close #1113.  The size will still be set when the screen is resumed in order to update the text size if the value is changed in **_Settings_** and the note editor is open as with a tablet in landscape orientation.

### Test
1. Tap any note in list with markdown enabled.
2. Tap **_Preview_** tab.
3. Tap back arrow in app bar.
4. Tap note from Step 1.
5. Notice **_Preview_** tab is shown.
6. Swipe right to show **_Edit_** tab.
7. Notice text size does not resize.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -351,6 +351,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
         mContentEditText.setOnCheckboxToggledListener(this);
         mContentEditText.setMovementMethod(SimplenoteMovementMethod.getInstance());
         mContentEditText.setOnFocusChangeListener(this);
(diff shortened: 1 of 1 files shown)


anitaa1990 (approved):
Works great :shipit:

PR #1116, opened 2020-08-04
Add Style Setting (Part 1)

theck13:
### Fix
This is the first part in a series of pull requests to add a style setting to the app.  The pull requests are split into parts in an attempt to make the changes smaller.  This first part will be larger than the subsequent parts since it's setting the foundation for adding other styles.  Also, there are parts of the code that may seem unnecessary (e.g. `switch` statements with one or two cases).  Those are structured so with the additional styles in mind.

This part adds the **_Default_** and **_Classic_** styles.  There are subtle differences between the two styles.  The most noticeable ones are the shade of blue and the **_Classic_** style tints the action icons blue.  See the screenshots below for illustration.

add_style_setting_all_01

### Test
1. Open navigation drawer.
2. Tap ***Settings*** item in drawer.
3. Tap ***Style*** item under ***Appearance*** section.
4. Notice ***Style*** screen with ***Default*** and ***Classic*** styles.
5. Notice ***Default*** style is selected and app has ***Default*** styling.
6. Tap ***Classic*** style.
7. Notice ***Classic*** style is selected and app has ***Classic*** styling.

### Review

### Release
These changes do not require release notes.  `RELEASE-NOTES.txt` will be updated once the styles feature is complete.


SylvesterWilmott:
Looking good.

For the record, we've already had some back and forth on this and the styles have been reviewed.

Re: The settings, there's some subtle between what's implemented and the design since the designs were removed from Zeplin. I added those designs again. As far as I can tell we only need to modify the style title text size, the top/bottom page margins and the style card strokes when not selected.

theck13:
@SylvesterWilmott, i made the updates you mentioned.  See the screenshots below for illustration.

add_style_setting_all_01_style

SylvesterWilmott (approved):
Looks great! Thanks @theck13

Simplenote/src/main/res/layout-night/style_list_row_default.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<androidx.cardview.widget.CardView
```
AmandaRiu, line 3: Note: There may be a specific reason for this design, but just in case I thought I'd mention an optimization that could lead to less possible issues in the future. Instead of maintaining two separate layout files just to change the text color, you could use the same layout file with color variables that have day/night colors specified in different `values/colors.xml` and `values-night/colors.xml`. For example, you can create a color named `text_title` and set it as follows:

`values-night/colors.xml`: `<color name="text_title">@android:color/white</color>`
`values/colors.xml`: `<color name="text_title">@color/gray_80</color>`

Then just use `@color/text_title` in the layout file and it will automatically use the appropriate color for the active theme.

AmandaRiu (approved):
@theck13 Really nice work on this project! I'm excited to see all the styles when it's finished! I left a comment in the code, but it's not anything that would hold up this PR so I'm approving and merging.

Simplenote/src/main/res/layout-night/style_list_row_default.xml
theck13: Thanks for the tip!  I'll try that for subsequent styles.

Issue #992, opened 2020-03-17
Changes being reverted and not saved when moving between apps


rezzap:
### Expected

I would expect that when I make changes in the app that changes will be kept if I move to another app or end up on the lock screen. 

### Observed

Changes are not always kept in history when moving between apps or going to lock screen and sometimes are reverted. 

### Reproduced
We weren't able to reproduce this, but the user did provide us with a screen recording of it happening. 
It's available here: https://share.getcloudapp.com/WnuNDrqX
Relevant content between times 1:20~1:40 & 3:00.

Steps to reproduce seem to be: 
> 1. Go to Simplenote app and add some text to a note. 
> 2. Use the app switcher button in the bottom navigation and go to a new app.
> 3. Screen locks. 
> 4. Unlock the phone and go back to Simplenote
> 3. See your changed being reverted and look through history previous changes were not registered. 

Make|Model|Android Version|App Version
-|-|-|-
Samsung|Galaxy 9|10|2.4 & 2.5-rc-1

This user replied with two additional bits of information
>When i finish writing with a note for the moment, i go back to the list view and select the note again to save and sync. It does help. When i do that it syncs to the internet right away. Then the reverse problem doesn't happen.

rezzap:
We have another very similar report for this here: 2777335-zen

They provided these steps to reproduce: 

> 1) Login Chrome web first to check
> 2) Login and edited note on Android app
> 3) Android app goes to background while checking another app
> 4) Back to Android app and all of my editing was gone
> 4-1) That note page looked like refreshing and its page moved onto upside down even I didn't touch
> 4-2) I write several words to test and its gone too. But it wasn't easily reproduced.
> 4-3) Assumption : if android app is going to background while syncing, and back then it seems like to discard local client one and gets from server.
> 5) Checking history on Android app but nothing left
> 6) Checking history on Chrome web but nothing left too.

erricgunawan:
Also got a similar report: 2799194-zen 

Translated: 

> I have used this application a lot on my tablet but for more than a month I have been worried and I will have to find another alternative because the new thing I write does not stay put, when I change my note or when I exit and reenter the application I check what I put and it is not, I write it again and it happens again and, I am left without the new information. Many times I even see how the writing disappears.

Also, 

> There can be nothing in the trash because what disappears are not whole notes, the problem is that the last words or phrases that I add to the notes disappear, when I change the note or close the application, so that when I return to the note where I added those last words or phrases are no longer.

[...]

PR #1128, opened 2020-08-19
Add App Log

theck13:
### Fix
Add application logging to assist in troubleshooting data loss reports as described in https://github.com/Automattic/simplenote-android/issues/992 and https://github.com/Automattic/simplenote-android/issues/1101 among other issues.  The logs are not saved to device storage, but instead are stored in memory while the app is alive.  Once the app is killed by the operating system, the logs are cleared.  The logs are queued with a maximum of one hundred events to keep the memory footprint small.  The logs remain on the device and are never sent to anywhere unless explicitly done so by the user.  These aspects ensure an opt-in behavior and keeps the user in control of their information.  In order to share the logs, the **_Send logs_** item was added to the **_Settings_** screen.  When the **_Send logs_** item is tapped, a **_Share_** sheet is shown allowing the user to share the logs however they choose.  Since this feature is intended to help diagnose issues with data loss while editing/creating notes, the logs are focused on editor and syncing events.  See the details below for example logs.

<details>
<summary>
Log
</summary>
<pre>
Device: Google Pixel 4 (flame)
System: Android 10 (29)
App: Simplenote 2.9-rc-1 (111)<br>
Email: heckofatest@gmail.com
Notes: 21
Tags: 5<br>
2020-08-18 13:51:08.453 - LAYOUT: Small, portrait
2020-08-18 13:51:08.510 - NETWORK: WIFI (780.00 Mbps)
2020-08-18 13:51:08.511 - SCREEN: Created (NotesActivity)
2020-08-18 13:51:08.609 - NETWORK: WIFI (780.00 Mbps)
2020-08-18 13:51:08.610 - SCREEN: Created (NoteListFragment)
2020-08-18 13:51:08.662 - ACTION: App opened
2020-08-18 13:51:08.662 - SYNC: Started note bucket (NotesActivity)
2020-08-18 13:51:08.663 - SYNC: Started tag bucket (NotesActivity)
2020-08-18 13:51:10.155 - SCREEN: Paused (NoteListFragment)
2020-08-18 13:51:10.156 - SYNC: Stopped tag bucket (NotesActivity)
2020-08-18 13:51:10.157 - SYNC: Stopped note bucket (NotesActivity)
2020-08-18 13:51:10.157 - SCREEN: Paused (NotesActivity)
2020-08-18 13:51:10.171 - NETWORK: MOBILE (53.14 Mbps)
2020-08-18 13:51:10.171 - SCREEN: Created (NoteEditorActivity)
[...]

PR #1137, opened 2020-08-25
Add Sync Worker

theck13:
### Fix
Add a worker to run syncing tasks in the background to avoid data loss as described in #992 among other issues.  The background tasks are starting and stopping the note, tag, and preference buckets, which effectively triggers syncing.  Currently, the buckets are stopped after the app has been closed for ten seconds.  The worker added in these changes starts after the app has been closed for ten seconds.  It also has a backoff criteria of one minute and a constraint which requires an active network connection.  The worker starts the buckets and stops them after ten seconds.  The flow is as follows:
- After the app is closed for ten seconds, start the worker
- If there is an active network connection...
    - Start the buckets
    - Start timer for one minute to start the buckets again
    - Stop the buckets after ten seconds
- When the app is resumed, stop the worker

Effectively, the buckets are started for ten seconds every minute when there is a network connection.

### Test
There aren't any user-facing changes to this pull request.  The best way to see what's happening is to watch the logcat with the logs filtered to show "Debug" only and the "(Started)|(Stopped)" regular expression is used.  See the logs below for example.

```
2020-08-25 11:07:01.796 17642-17642/com.automattic.simplenote.debug D/Simplenote.onActivityPaused: Started worker
2020-08-25 11:07:11.866 17642-17883/com.automattic.simplenote.debug D/SyncWorker.doWork: Started buckets
2020-08-25 11:07:21.908 17642-17642/com.automattic.simplenote.debug D/SyncWorker.doWork: Stopped buckets
2020-08-25 11:08:11.912 17642-17991/com.automattic.simplenote.debug D/SyncWorker.doWork: Started buckets
2020-08-25 11:08:21.954 17642-17642/com.automattic.simplenote.debug D/SyncWorker.doWork: Stopped buckets
2020-08-25 11:08:24.860 17642-17642/com.automattic.simplenote.debug D/Simplenote.onActivityResumed: Stopped worker
```

While watching the logcat, launch the app then send it to the background by going to the home screen or opening another app.  Verify the buckets are started and stopped according to the flow mentioned above.

### Review
Only one developer is required to review these changes, but anyone can perform the review.  @malinajirka and @maxme, I requested you as reviewers too since you commented on the background service idea and seem to have experience with this type of implementation.

### Release
`RELEASE-NOTES.txt` was updated in 2f92028 with:
> Added syncing in the background to avoid data loss when exiting the app

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -10,6 +10,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.util.Log;
 
+import androidx.annotation.NonNull;
 import androidx.appcompat.app.AppCompatDelegate;
 import androidx.work.BackoffPolicy;
 import androidx.work.Constraints;
@@ -202,7 +203,7 @@ public class Simplenote extends Application {
         }
 
         @Override
-        public void onConfigurationChanged(Configuration newConfig) {
+        public void onConfigurationChanged(@NonNull Configuration newConfig) {
             AppLog.add(Type.LAYOUT, DisplayUtils.getDisplaySizeAndOrientation(Simplenote.this));
         }
 
@@ -213,7 +214,7 @@ public class Simplenote extends Application {
         // ActivityLifeCycle callbacks
         @SuppressLint("LongLogTag")
         @Override
-        public void onActivityResumed(Activity activity) {
+        public void onActivityResumed(@NonNull Activity activity) {
             if (mIsInBackground) {
                 AnalyticsTracker.track(
                         AnalyticsTracker.Stat.APPLICATION_OPENED,
@@ -229,16 +230,16 @@ public class Simplenote extends Application {
         }
 
         @Override
-        public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
+        public void onActivityCreated(@NonNull Activity activity, Bundle savedInstanceState) {
         }
 
         @Override
-        public void onActivityStarted(Activity activity) {
+        public void onActivityStarted(@NonNull Activity activity) {
         }
 
         @SuppressLint("LongLogTag")
         @Override
-        public void onActivityPaused(Activity activity) {
+        public void onActivityPaused(@NonNull Activity activity) {
             PeriodicWorkRequest syncWorkRequest = new PeriodicWorkRequest.Builder(
                 SyncWorker.class,
                 PeriodicWorkRequest.MIN_PERIODIC_INTERVAL_MILLIS,
@@ -258,15 +259,15 @@ public class Simplenote extends Application {
         }
 
         @Override
-        public void onActivityStopped(Activity activity) {
+        public void onActivityStopped(@NonNull Activity activity) {
         }
 
         @Override
-        public void onActivitySaveInstanceState(Activity activity, Bundle outState) {
+        public void onActivitySaveInstanceState(@NonNull Activity activity, @NonNull Bundle outState) {
         }
 
         @Override
-        public void onActivityDestroyed(Activity activity) {
+        public void onActivityDestroyed(@NonNull Activity activity) {
         }
     }
 }
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -9,6 +9,12 @@ import android.os.Bundle;
 import android.os.Handler;
 
 import androidx.appcompat.app.AppCompatDelegate;
+import androidx.work.BackoffPolicy;
+import androidx.work.Constraints;
+import androidx.work.ExistingPeriodicWorkPolicy;
+import androidx.work.NetworkType;
+import androidx.work.PeriodicWorkRequest;
+import androidx.work.WorkManager;
 
 import com.automattic.simplenote.analytics.AnalyticsTracker;
 import com.automattic.simplenote.analytics.AnalyticsTrackerNosara;
@@ -22,12 +28,15 @@ import com.automattic.simplenote.utils.AppLog.Type;
 import com.automattic.simplenote.utils.CrashUtils;
 import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.PrefUtils;
+import com.automattic.simplenote.utils.SyncWorker;
 import com.simperium.Simperium;
 import com.simperium.client.Bucket;
 import com.simperium.client.BucketNameInvalid;
 import com.simperium.client.BucketObjectMissingException;
 
 import org.wordpress.passcodelock.AppLockManager;
+
+import java.util.concurrent.TimeUnit;
 
 import static com.automattic.simplenote.models.Preferences.PREFERENCES_OBJECT_KEY;
 
@@ -211,6 +220,7 @@ public class Simplenote extends Application {
 
                 mIsInBackground = false;
                 AppLog.add(Type.ACTION, "App opened");
+                WorkManager.getInstance(getApplicationContext()).cancelUniqueWork(TAG_SYNC);
             }
         }
 
@@ -224,6 +234,21 @@ public class Simplenote extends Application {
 
         @Override
         public void onActivityPaused(Activity activity) {
+            PeriodicWorkRequest syncWorkRequest = new PeriodicWorkRequest.Builder(
+                SyncWorker.class,
+                PeriodicWorkRequest.MIN_PERIODIC_INTERVAL_MILLIS,
+                TimeUnit.MILLISECONDS
+            )
+                .setConstraints(new Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build())
+                .setBackoffCriteria(BackoffPolicy.LINEAR, ONE_MINUTE_MILLIS, TimeUnit.MILLISECONDS)
+                .setInitialDelay(TEN_SECONDS_MILLIS, TimeUnit.MILLISECONDS)
+                .addTag(TAG_SYNC)
+                .build();
+            WorkManager.getInstance(getApplicationContext()).enqueueUniquePeriodicWork(
+                TAG_SYNC,
+                ExistingPeriodicWorkPolicy.REPLACE,
+                syncWorkRequest
+            );
         }
 
         @Override
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -32,22 +32,23 @@ import org.wordpress.passcodelock.AppLockManager;
 import static com.automattic.simplenote.models.Preferences.PREFERENCES_OBJECT_KEY;
 
 public class Simplenote extends Application {
-
-    private static final int TEN_SECONDS_MILLIS = 10000;
-
-    // log tag
-    public static final String TAG = "Simplenote";
-
-    // intent IDs
-    public static final int INTENT_PREFERENCES = 1;
-    public static final int INTENT_EDIT_NOTE = 2;
     public static final String DELETED_NOTE_ID = "deletedNoteId";
     public static final String SELECTED_NOTE_ID = "selectedNoteId";
+    public static final String TAG = "Simplenote";
+    public static final int INTENT_EDIT_NOTE = 2;
+    public static final int INTENT_PREFERENCES = 1;
+    public static final int ONE_MINUTE_MILLIS = 60 * 1000;  // 60 seconds
+    public static final int TEN_SECONDS_MILLIS = 10 * 1000;  // 10 seconds
+
     private static final String AUTH_PROVIDER = "simplenote.com";
-    private Simperium mSimperium;
+    private static final String TAG_SYNC = "sync";
+
+    private static Bucket<Preferences> mPreferencesBucket;
+
     private Bucket<Note> mNotesBucket;
     private Bucket<Tag> mTagsBucket;
-    private static Bucket<Preferences> mPreferencesBucket;
+    private Simperium mSimperium;
+    private boolean mIsInBackground = true;
 
     public void onCreate() {
         super.onCreate();
@@ -141,10 +142,11 @@ public class Simplenote extends Application {
         return mPreferencesBucket;
     }
 
-    private class ApplicationLifecycleMonitor implements Application.ActivityLifecycleCallbacks,
-            ComponentCallbacks2 {
-        private boolean mIsInBackground = true;
+    public boolean isInBackground() {
+        return mIsInBackground;
+    }
 
+    private class ApplicationLifecycleMonitor implements Application.ActivityLifecycleCallbacks, ComponentCallbacks2 {
         // ComponentCallbacks
         @Override
         public void onTrimMemory(int level) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -1,5 +1,6 @@
 package com.automattic.simplenote;
 
+import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.Application;
 import android.content.ComponentCallbacks2;
@@ -7,6 +8,7 @@ import android.content.res.Configuration;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
+import android.util.Log;
 
 import androidx.appcompat.app.AppCompatDelegate;
 import androidx.work.BackoffPolicy;
@@ -209,6 +211,7 @@ public class Simplenote extends Application {
         }
 
         // ActivityLifeCycle callbacks
+        @SuppressLint("LongLogTag")
         @Override
         public void onActivityResumed(Activity activity) {
             if (mIsInBackground) {
@@ -221,6 +224,7 @@ public class Simplenote extends Application {
                 mIsInBackground = false;
                 AppLog.add(Type.ACTION, "App opened");
                 WorkManager.getInstance(getApplicationContext()).cancelUniqueWork(TAG_SYNC);
+                Log.d("Simplenote.onActivityResumed", "Stopped worker");
             }
         }
 
@@ -232,6 +236,7 @@ public class Simplenote extends Application {
         public void onActivityStarted(Activity activity) {
         }
 
+        @SuppressLint("LongLogTag")
         @Override
         public void onActivityPaused(Activity activity) {
             PeriodicWorkRequest syncWorkRequest = new PeriodicWorkRequest.Builder(
@@ -249,6 +254,7 @@ public class Simplenote extends Application {
                 ExistingPeriodicWorkPolicy.REPLACE,
                 syncWorkRequest
             );
+            Log.d("Simplenote.onActivityPaused", "Started worker");
         }
 
         @Override
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
@@ -1,0 +1,69 @@
+package com.automattic.simplenote.utils;
+
+import android.content.Context;
+import android.os.Handler;
+import android.os.Looper;
+
+import androidx.annotation.NonNull;
+import androidx.work.Worker;
+import androidx.work.WorkerParameters;
+
+import com.automattic.simplenote.Simplenote;
+import com.automattic.simplenote.models.Note;
+import com.automattic.simplenote.models.Preferences;
+import com.automattic.simplenote.models.Tag;
+import com.simperium.client.Bucket;
+
+import static com.automattic.simplenote.Simplenote.TEN_SECONDS_MILLIS;
+
+public class SyncWorker extends Worker {
+    private Bucket<Note> mBucketNote;
+    private Bucket<Preferences> mBucketPreference;
+    private Bucket<Tag> mBucketTag;
+
+    public SyncWorker(@NonNull Context context, @NonNull WorkerParameters params) {
+        super(context, params);
+        Simplenote application = (Simplenote) context.getApplicationContext();
+        mBucketNote = application.getNotesBucket();
+        mBucketTag = application.getTagsBucket();
+        mBucketPreference  = application.getPreferencesBucket();
+    }
+
+    @NonNull
+    @Override
+    public Result doWork() {
+        mBucketNote.start();
+        mBucketTag.start();
+        mBucketPreference.start();
+
+        new Handler(Looper.getMainLooper()).postDelayed(
+            new Runnable() {
+                @Override
+                public void run() {
+                    if (mBucketNote != null) {
+                        mBucketNote.stop();
+                    }
+
+                    if (mBucketTag != null) {
+                        mBucketTag.stop();
+                    }
+
+                    if (mBucketPreference != null) {
+                        mBucketPreference.stop();
+                    }
+                }
+            },
+            TEN_SECONDS_MILLIS
+        );
+
+        return Result.retry();
+    }
+
+    @Override
+    public void onStopped() {
+        super.onStopped();
+        mBucketNote.stop();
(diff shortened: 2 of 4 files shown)


obsantos:
Related to `WorkManager` there is an ongoing issue regarding how they handle the internal database to schedule the jobs that is creating a crash rate increase to many projects (Tumblr included). When asked for a solution they told us they. are working on an API to help handle these situations. More info on the crash here:
https://issuetracker.google.com/issues/152202040

In my opinion shouldn't be a blocker, but just wanted to raise this in case in the future you start seeing crashes, so there is awareness of this issue.

malinajirka:
Hi @theck13! I've been quite busy with reviews lately so I want to focus on project work for the next couple of days. I can review this PR on Tuesday unless someones beats me to it.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -209,31 +224,50 @@ public void onActivityResumed(Activity activity) {
...
         @Override
-        public void onActivityStarted(Activity activity) {
+        public void onActivityStarted(@NonNull Activity activity) {
         }
 
+        @SuppressLint("LongLogTag")
         @Override
-        public void onActivityPaused(Activity activity) {
+        public void onActivityPaused(@NonNull Activity activity) {
+            PeriodicWorkRequest syncWorkRequest = new PeriodicWorkRequest.Builder(
```
malinajirka, line 243: We might want to consider extracting the work manager logic into a separate class in order to follow the SRP. Wdyt?

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -209,31 +224,50 @@ public void onActivityResumed(Activity activity) {
...
+                SyncWorker.class,
+                PeriodicWorkRequest.MIN_PERIODIC_INTERVAL_MILLIS,
```
malinajirka, line 245: I'm not sure I fully understand why we need to sync the data so often. AFAIU we are introducing the syncing logic mostly to make sure the user doesn't lose local data, right? If the first sync succeeds and the user doesn't open the app again - is there even anything to sync? Because I think the current setup will continue starting the work manager indefinitely every minute - I'm mostly concerned about battery life. For example in WPAndroid we run the manager every 12 hours. I realize that WorkManager uses some optimization techniques to preserve battery life, but I'd like to understand if we really need such an intense sheduling. Wdyt?

Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
```diff
@@ -0,0 +1,107 @@
...
+                        }
+
+                        Log.d("SyncWorker.doWork", "Stopped buckets");
+                    }
+                }
+            },
+            TEN_SECONDS_MILLIS
+        );
+
+        return Result.retry();
```
malinajirka, line 83: AFAIU this might not work as intended - the OS keeps the device awake and gives the app a higher priority only while the "doWork" method is being executed. If we return from the method right away we'll lose these privileges. Moreover, the system can kill/destroy the worker anytime. 

From the docs:
"Worker classes are instantiated at runtime by WorkManager and the doWork() method is called on a pre-specified background thread (see Configuration.getExecutor()). This method is for synchronous processing of your work, meaning that once you return from that method, the Worker is considered to be finished and will be destroyed."

Wdyt?

malinajirka (commented):
Thanks @theck13! I've reviewed the code and I left a few of comments. Let me know what you think.

P.S. Have you considered writing the new code in Kotlin? Coroutine worker is easier to implement and it's more readable.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
theck13: If we use this in multiple places or the code was repeated, I would extract it.  In this case, this is the only place it's used.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
theck13: > we are introducing the syncing logic mostly to make sure the user doesn't lose local data, right? If the first sync succeeds and the user doesn't open the app again - is there even anything to sync?

Syncing is not only local data.  Remote data from other platforms and collaborators can affect the local data and revision history.  Syncing in the background allows the app to update local and remote data to avoid overwriting and revisions out of order.

> I'd like to understand if we really need such an intense sheduling.

Reports we've received vaguely mention leaving the app and coming back to see data is lost.  We haven't received detailed reports to be sure.  We [added application logging](https://github.com/Automattic/simplenote-android/pull/1128) to attempt to troubleshoot the issue ourselves, but that hasn't been released yet.  That's why a short interval was chosen.

> I'm mostly concerned about battery life.

This is the first attempt to avoid data loss and mimic the background syncing, which was [fixed to lower battery usage](https://github.com/Automattic/simplenote-android/pull/414/commits/e88c4eab8e0baa5105535b1babc851591df1a809).  If it proves to increase battery usage, we'll either revert or update these changes like extending the interval.  Ultimately, data loss is more important than battery usage.  Since we aren't getting helpful reports from users and we can't reproduce the issue at all, this will help us understand what the root cause is.

Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
theck13: Which part do you think will not work as intended?  Are you talking about the post-delayed handler stopping the buckets?  When testing these changes, the buckets were stopped as expected.

```
2020-08-25 11:07:01.796 17642-17642/com.automattic.simplenote.debug D/Simplenote.onActivityPaused: Started worker
2020-08-25 11:07:11.866 17642-17883/com.automattic.simplenote.debug D/SyncWorker.doWork: Started buckets
2020-08-25 11:07:21.908 17642-17642/com.automattic.simplenote.debug D/SyncWorker.doWork: Stopped buckets
2020-08-25 11:08:11.912 17642-17991/com.automattic.simplenote.debug D/SyncWorker.doWork: Started buckets
2020-08-25 11:08:21.954 17642-17642/com.automattic.simplenote.debug D/SyncWorker.doWork: Stopped buckets
2020-08-25 11:08:24.860 17642-17642/com.automattic.simplenote.debug D/Simplenote.onActivityResumed: Stopped worker
```

theck13 (commented):
> Have you considered writing the new code in Kotlin?

Simplenote does not have any Kotlin at the moment.  Unless there is a significant advantage to writing something in Kotlin, I don't think it's worth the maintenance to add it.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
malinajirka: I personally separate things even when they are not used multiple times. It imo makes the code easier to read and maintain. Moreover, it's easier for future devs to find related/similar logic. Having said that I'm completely fine keeping it here.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
malinajirka: >Ultimately, data loss is more important than battery usage.

I can't say I agree with this statement. I'm sure there are solutions which won't drain battery live and won't result in data loss. 

> If it proves to increase battery usage, we'll either revert or update these changes like extending the interval.

Assuming we'll see a huge battery increase. However, usually these things cumulate - one small thing at a time.

>Since we aren't getting helpful reports from users and we can't reproduce the issue at all, this will help us understand what the root cause is.

What if we synced the data once after the user leaves the app and then when they come back to the app?

Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
malinajirka: I think both the syncing itself and stop() might not work as intended. When the worker finishes the OS assumes the app finished the job.

The fact that it works during our tests unfortunately doesn't prove anything. Perhaps the issue occurs only when the worker is started from a doze mode or on low batter or low on memory or something else. It's hard to tell. However, the documentation is quite specific about this. Wdyt?

malinajirka:
Thanks for the replies @theck13 ;)! 

I'm still not sure about this solution. The worker will be running 1/6 of the time (50seconds idle 10 seconds running). It'll be running for 10s even when there isn't anything to sync. I understand that any content loss experience is terrible for users. However, I still feel like we might want to consider using a less resource-hungry solution. 

If we need to be immediately informed about changes made on other platforms we might need to implement some kind of server-push mechanism. Running the worker every 50s feels like a workaround. Having said that, AFAIU the reports that's not the issue we are dealing with. The users don't mention editing the notes on other platforms - they just put the app to background. So it seems like we don't store the local changes and the app gets killed before it pushes them to the server. I'm not familiar with the architecture of the app - do we store the notes/edits into a local db or do we just pass them to Simperium and that's it? If it's the latter, does Simperium store them into a local db or it first tries to push them to the server + initiates sync?


Any thoughts @maxme @AmandaRiu @mchowning @anitaa1990 ?



P.S.
>Simplenote does not have any Kotlin at the moment. Unless there is a significant advantage to writing something in Kotlin, I don't think it's worth the maintenance to add it.

I think the additional maintenance is negligible and the speed of development, readability of the code and features such as null safety are worth it. But that's a discussion for another day:)

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
theck13: > What if we synced the data once after the user leaves the app and then when they come back to the app?

Data is continued to be synced for ten seconds after the app is exited.  After ten seconds, the buckets are stopped and syncing stops.  When the app is resumed, the buckets are started again and syncing continues.  So, what you're suggesting is already happening.

Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
theck13: > Perhaps the issue occurs

What issue?

Are you concerned that the buckets will remain started and never stopped in the post-delayed handler since the worker is destroyed?  If so, we can create another worker to stop them or update this worker to detect if the buckets are already started and stop them.

theck13 (commented):
> If we need to be immediately informed about changes made on other platforms we might need to implement some kind of server-push mechanism.

This is already done in Simperium _if the buckets are started_.  The buckets are listening for changes.  If the buckets are not started, they are not listening.  That's the problem these changes are trying to solve.

> Having said that, AFAIU the reports that's not the issue we are dealing with. The users don't mention editing the notes on other platforms - they just put the app to background.

We are getting reports with various _hints_ about what _could_ be the issue.  Nothing is definitive.  We're not getting much from users.  So, we're trying a solution based on the [fix from four years ago](https://href.li/?https://github.com/Automattic/simplenote-android/pull/414).

> do we store the notes/edits into a local db or do we just pass them to Simperium and that's it? If it's the latter, does Simperium store them into a local db or it first tries to push them to the server + initiates sync?

We store the data locally regardless of online or offline.  The syncing is initiated when there is a network connection and changes are detected.

Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
malinajirka: >What issue?

After the worker is done - the `doWork` method completes - the system can kill the app -> it can kill even the syncing. Tbh I don't see any reason why we should go directly against the documentation. I'm also worried that other devs might copy this implementation (about which we already know that sometime doesn't work) and use it for other jobs which can't be interrupted in the middle.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
malinajirka: So we already sync our local data and we send them to the server. The only thing which can by synced in this new worker is that we receive data from the server -> do we have any clues which would suggest that this solution will help to "restore/save" the local data?

malinajirka:
I tried to reproduce the issue and I'm wondering if this content-loss might be related. 



- Set poor network connection on an emulator
1. Open a note
2. Keep typing text -> copy pasting -> removing blocks
3. Restart the app using adb
4. Boom - all the changes you've made are lost

I realize the users don't force-restart the app. However, some of the reports mention that the users leave the app and when they get back their changes are lost. When the user puts the app into a background the system might kill it (eg. when it's low on resources) - so that's technically like restarting the app. 

What might also help is if we started a foreground service when the user is about to leave the app - to make sure it has a higher priority and the app doesn't get killed by the system during the 10s period after the user left.

Also perhaps some more frequent auto-save mechanism while the app is in foreground might help with this. 

Wdyt?

Btw do we have any logs from users which would suggest that the app got killed/crashed when they lost content?

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
theck13: > So we already sync our local data and we send them to the server. The only thing which can by synced in this new worker is that we receive data from the server

That's an assumption we've made in the past, but I don't think it's always the case.  The buckets used to be left started indefinitely, which was thought to be causing high battery consumption even though it was never proven.  Now, we stop the buckets after the app is exited for ten seconds.  During that time, it's not guaranteed sync occurs due to the network connection.

> do we have any clues which would suggest that this solution will help to "restore/save" the local data?

This solution doesn't change how the local data is saved.  It simply starts and stops the buckets to trigger syncing.

Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
theck13: We're not intentionally going against the documentation.

> about which we already know that sometime doesn't work

What do you mean?

theck13 (commented):
> I realize the users don't force-restart the app. However, some of the reports mention that the users leave the app and when they get back their changes are lost. When the user puts the app into a background the system might kill it (eg. when it's low on resources) - so that's technically like restarting the app.

By force-restarting the app, you skip steps in the lifecycle that we use to save data.  First, we [save changes after typing has stopped for two seconds](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L929-L933).  Second, we [save the note when the editor is paused](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L983).  Both of those are skipped when force-restarting the app that are not skipped when the user is exiting the app.

> What might also help is if we started a foreground service when the user is about to leave the app - to make sure it has a higher priority and the app doesn't get killed by the system during the 10s period after the user left.

This might be an option we try.  I was hoping to use this background service as a first step since it doesn't force Simplenote to linger around in the foreground and it mimics the previous background syncing the closest.

> Also perhaps some more frequent auto-save mechanism while the app is in foreground might help with this.

As I mentioned above, we [save changes after typing has stopped for two seconds](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L929-L933).  Are you suggesting shortening that interval?

> Btw do we have any logs from users which would suggest that the app got killed/crashed when they lost content?

We don't.  We don't have much to go on at all to be honest.  That's why I [added the application logging](https://github.com/Automattic/simplenote-android/pull/1128) to get some information that might help.

malinajirka:
> By force-restarting the app, you skip steps in the lifecycle that we use to save data. First, we save changes after typing has stopped for two seconds. Second, we save the note when the editor is paused. Both of those are skipped when force-restarting the app that are not skipped when the user is exiting the app.

Thanks for linking to the exact lines of the code where we save the data;)! 

I realize I'm skipping the lifecycle. However, the behavior seems to exactly correspond to some of the scenarios the users are reporting. Moreover, [the asynctask](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L983) might never get executed even when the user leaves the app using (home/back/switch app) buttons.

There are two issues I can see in the code
1. Right after the app goes to background, the system can freely kill it (it can theoretically kill it to gain resources even while it's in foreground, but that's an extreme edge case)
2. The async task keeps a weakreference to the fragment, however, when the user leaves the app the garbage collector can freely collect the fragment and [this line](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L1597) never gets executed. My personal bet is that this is the true cause of the issue and starting a worker won't fix it.

Wdyt?

>As I mentioned above, we save changes after typing has stopped for two seconds. Are you suggesting shortening that interval?

In WPAndroid we store changes when typing has stopped for 500ms or when the user makes 50changes (writes/removes 50 characters).

theck13:
The [`SaveNoteTask` asynchronous task](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L1585) is called after [typing has stopped for two seconds](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L929-L933).  The [`saveNote()` method](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L1146) is called [when the app is paused](https://github.com/Automattic/simplenote-android/blob/fce807283c67eeec1d1abfb3a2bbe0e74b5b02bd/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L511).  If the user leaves the app before the `SaveNoteTask` asynchronous task is triggered, the `saveNote()` method should be triggered.  It's true that we shouldn't count on the weak reference when the app is exited.  We don't.  That occurs while the app is still in the foreground.  When the app is put into the background, we save the note immediately.

malinajirka:
>The saveNote() method is called when the app is paused.

Ohh, you are right, it's called directly in onPause().

>That occurs while the app is still in the foreground. When the app is put into the background, we save the note immediately.

The issue might be in the "Syncable.save()/Bucket.sync()" method - it doesn't save the object synchronously. 

Btw do you know when [this exception](https://github.com/Automattic/simplenote-android/blob/5b51f583ee054f91b59b8939d68481b77ec9bae6/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L1181) might be thrown?

theck13:
The `BucketObjectMissingException` is thrown when trying to get a `Note` or `Tag` object from a `Bucket<Note>` or `Bucket<Tag>`, respectively.  In the `saveNote()` method, that occurs [here](https://github.com/Automattic/simplenote-android/blob/5b51f583ee054f91b59b8939d68481b77ec9bae6/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L1152).  Basically, if we don't find the note or tag in the associated bucket, we say the object is missing.

Simplenote/src/main/java/com/automattic/simplenote/utils/SyncWorker.java
malinajirka: How come? The documentation says `This method is for synchronous processing of your work`, but we are starting an asynchronous work there.

theck13: Ah, I see what you mean.  The post-delayed handler and the buckets by their nature are asynchronous.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
malinajirka: > > do we have any clues which would suggest that this solution will help to "restore/save" the local data?

> This solution doesn't change how the local data is saved. It simply starts and stops the buckets to trigger syncing.

AFAIU the users can't recover the content even when they go back to the app on the same device right? So AFAICT the issue is that the app doesn't save the data locally before it is exited/killed ==> PersistentStore.save(..) doesn't get executed. Do you think calling "Bucket.start()" will help with that issue? I still feel like we are trying to solve a different issue with the worker. If the issue was that users wouldn't see their data on other devices, I'd understand this solution (I still wouldn't agree with scheduling interval). However, afaik that's not the case, right? Could you please elaborate, why you think it'll help. Thanks!

theck13: > the issue is that the app doesn't save the data locally before it is exited/killed

This isn't clear from user reports.  We instruct them to check the **_History_** feature since the revisions could be out of the expected order.  Users have reported that they find their "lost" content there.  The problem is that users just say something like "data lost" or "sync broken" and don't contact us.  So, we're not certain if it's lost, out of order, or something else.

The idea behind these changes is to mimic the syncing behavior of the app before we started getting reports of "data lost" or "sync broken" while avoiding the potential high battery usage due to keeping the buckets started indefinitely.  Since we don't have much to go on, changing one thing that supposedly worked previously will help diagnose things.  If it doesn't help, then looking at the Simperium Android library would be the next step.

malinajirka: >while avoiding the potential high battery usage 

I don't think we'll be avoiding that - 10s every 50s is a huge battery usage. We'll keeping waking up the device during nights even when it lays still on a table. I personally have to recommend to not merge these changes. We still have ideas which could fix the issue and they are not so drastic.

Also if we don't have any detailed reports - is it a common issue? These changes will affect all users not just those who encountered the issue.

theck13: > I don't think we'll be avoiding that - 10s every 50s is a huge battery usage.

I think we will be avoiding that.  Just starting/stopping the buckets is not necessarily the battery usage issue.  There's nothing proving what was truly the cause of high battery usage previously, but we believe it's due to a disconnected network, out-of-control buckets trying to sync, or some combination of both.  These changes will avoid both of those situations by only triggering when there is a network connection and starting/stopping the buckets.

> Also if we don't have any detailed reports - is it a common issue?

We don't know.  We have been getting reports on Play Store reviews, but that's not the best gauge of whether it's common or not.  The fact that data loss happens at all is a problem.

malinajirka: > I think we will be avoiding that. Just starting/stopping the buckets is not necessarily the battery usage issue

I believe just opening a socket can run down the memory as it wakes the CPU. Of course if you do more, you drain more. It might not be such a huge usage that users would report it. However, I still think apps should try to use as little as possible and the proposed approach unfortunately doesn't follow that. The documentation says: 

```When an app connects to the mobile network in the background, the app wakes up the CPU and turns on the radio. Doing so repeatedly can run down a device's battery.```

I realize WorkManager improves the  battery consumption a lot as it includes several optimization techniques such as batching jobs from different apps. However, afaik it doesn't mean we can freely run a background work every 50s for 10s. 


I just realized one more thing `MIN_PERIODIC_INTERVAL_MILLIS` is set to 15 minutes (at least on Android 9, 10). We are cheating by returning `Result.retry()` from `doWork()`. I believe this is not how the "Result.retry()" was intended to be used. 

```Returns an instance of {@link Result} that can be used to indicate that the work encountered a transient failure and should be retried with backoff specified in {@link WorkRequest.Builder#setBackoffCriteria(BackoffPolicy, long, TimeUnit)}```


So even this is a huge indicator that we are running the task way too often. We didn't encounter any error and we are just misusing this technique so we can run the worker more often than `MIN_PERIODIC_INTERVAL_MILLIS`. I think running the task every 15 minutes is still too often for our use case.

PR #1143, opened 2020-09-08
Add Tracking Recent

theck13:
### Fix
Add analytics to track when recent searches is used.

### Test
There are no user-facing changes in this pull request.  In order to check if the analytics tracking is triggered in the right spot, set a breakpoint [here](https://github.com/Automattic/simplenote-android/compare/add-tracking-recent?expand=1#diff-291124c3836fdaed5786d9e1f5aab559R1272) and follow the steps below.
1. Tap ***Search*** action in top app bar.
2. Tap any recent search item.

### Review

### Release


AmandaRiu (approved):
Works! :shipit:

PR #1154, opened 2020-09-21
Merge release 2.11 into develop

mokagio:


Simplenote/metadata/release_notes.txt
```diff
@@ -1,3 +1,3 @@
-* Added improvements for devices with hardware keyboards
-* Added support for screen readers when adding a new note
-* Added condensed list option for widget
+* Added linking between notes
+* Fixed a bug with toggling tab-indented checkboxes
+
```
mokagio, line 3: The release toolkit automation copied over the `#` numbers of the PRs from `RELEASE-NOTES.txt`, see [this commit](https://github.com/Automattic/simplenote-android/pull/1154/commits/90f4d4dfb500b9bceb18089b420d9416672fb417)

I manually removed, my thinking being that they add noise to the release notes that would be displayed on Google Play. Do you agree?

If we want to expose the PR numbers to the users, which might actually be cool, gives them a way to see us working in the open, then I think we should make them links to GitHub, not just numerical references

theck13: I don't think we should have pull request numbers or links to GitHub in the release notes published in the Play Store listing.  It won't be helpful to **_all_** users and it will further limit the amount of information we are able to include in release notes.  The addition of the pull request numbers was for internal testing posts rather than external release notes.

theck13 (approved):

Simplenote/metadata/release_notes.txt
mokagio: Agreed.



We might need to tweak the code that generates that diff to filter out links in the form of `\[\#\d+\]` or something like that.

theck13: Yeah, we should automate removing those links if possible.  I'm not sure if the internal testing posts prefer the #123 format or the full GitHub link though.

PR #1157, opened 2020-09-23
Crash Encode Charset

theck13:
### Fix
Update the URL encoder character encoding parameter is the `TagUtils.hashTagValid` and `TagUtils.hashTagValid` methods to avoid an `IllegalCharsetNameException` crash on Android 6.0.x.  The static constant `StandardCharsets.UTF_8`'s `toString()` method throws the exception since `java.nio.charset.CharsetICU[UTF-8]` is returned, which is not a proper character encoding value.  `StandardCharsets.UTF_8.toString()` returns `UTF-8` for all other Android versions supported.  Replacing `toStrong()` with `name()` returns the canonical name, which is `UTF-8` for all versions including Android 6.0.x.

### Test
Since the `IllegalCharsetNameException` crash occurs only on Android 6.0.x (API 23) devices, it's best to test on at least one Android 6.0.x (API 23) device and one Android 7.0 (API 24) or higher device.
1. Tap any note in list.
2. Enter any text in ***Add tag...*** field at bottom of note.
3. Tap ***Spacev or ***Enter***/***Return*** key.
4. Notice text from Step 2 is added as tag to note.
5. Notice app does not crash.
6. Tap back arrow in app bar.
7. Open navigation drawer.
8. Tap ***Edit*** button in ***Tags*** header.
9. Tap tag from Step 4.
10. Notice ***Rename Tag*** dialog is shown.
11. Notice app does not crash.

### Review

### Release
`RELEASE-NOTES.txt` was updated in 19909313 with:
> Fixed adding or editing a tag on Android 6 [https://github.com/Automattic/simplenote-android/pull/1157]

#### Note
@mokagio, these changes will require a hotfix and new release candidate build once they are merged.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,7 @@
 2.11
 -----
-* Added linking between notes [#1145]
-* Fixed a bug with toggling tab-indented checkboxes [#1148]
+* Added linking between notes [https://github.com/Automattic/simplenote-android/pull/1145]
+* Fixed toggling tab-indented checkboxes [https://github.com/Automattic/simplenote-android/pull/1148]
 * Fixed adding or editing a tag on Android 6 [https://github.com/Automattic/simplenote-android/pull/1157]
 
 2.10
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,8 @@
 2.11
 -----
+* Added linking between notes [#1145]
+* Fixed a bug with toggling tab-indented checkboxes [#1148]
+* Fixed adding or editing a tag on Android 6 [https://github.com/Automattic/simplenote-android/pull/1157]
 
 2.10
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,7 @@
 2.11
 -----
-* Added linking between notes [https://github.com/Automattic/simplenote-android/pull/1145]
-* Fixed toggling tab-indented checkboxes [https://github.com/Automattic/simplenote-android/pull/1148]
+* Added linking between notes [#1145]
+* Fixed a bug with toggling tab-indented checkboxes [#1148]
 * Fixed adding or editing a tag on Android 6 [https://github.com/Automattic/simplenote-android/pull/1157]
 
 2.10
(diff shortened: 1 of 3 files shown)


RELEASE-NOTES.txt
```diff
@@ -1,5 +1,8 @@
 2.11
 -----
+* Added linking between notes [https://github.com/Automattic/simplenote-android/pull/1145]
+* Fixed toggling tab-indented checkboxes [https://github.com/Automattic/simplenote-android/pull/1148]
+* Fixed adding or editing a tag on Android 6 [https://github.com/Automattic/simplenote-android/pull/1157]
```
mokagio: Unfortunately, we haven't put automation on the Fastlane lane that makes the code freeze to add a new release note entry.

To avoid conflicts with the `release/2.11` branch and to keep a "correct" release notes log, I suggest doing this. Really, I should have though of that when I pushed the branch, but I didn't notice it 😞 

```suggestion

2.10.1
-----

* Fixed adding or editing a tag on Android 6 [https://github.com/Automattic/simplenote-android/pull/1157]
```

Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
```diff
@@ -62,7 +62,7 @@ public static String hashTag(String name) {
         try {
             String normalized = Normalizer.normalize(name, Normalizer.Form.NFC);
             String lowercased = normalized.toLowerCase(Locale.US);
-            String encoded = URLEncoder.encode(lowercased, StandardCharsets.UTF_8.toString());
+            String encoded = URLEncoder.encode(lowercased, StandardCharsets.UTF_8.name());
```
dmsnell, line 65: yikes. thanks for fixing this

RELEASE-NOTES.txt
theck13: I reverted the commit changing the descriptions in 691c5243 and I moved the tag fix description from 2.11 to 2.10.1 in 4b2fba6a.

mokagio (approved):
I tested this on a Pixel 2 API 23 Emulator. I was able to reproduce the crash when adding a new tag and when loading the edit tag dialog on `develop`. On this branch, it worked.

PR #1175, opened 2020-10-12
Merge hotfix 2.11.1 into release 2.12 with conflicts resolution

mokagio:
As one would expect, the conflicts where on:

- `RELEASE-NOTES.txt`
- `Simplenote/build.gradle`

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -30,9 +30,9 @@ android {
         if (project.hasProperty("versionName")) {
             versionName project.property("versionName")
         } else {
-            versionName "2.11"
+            versionName "2.11.1"
         }
-        versionCode 119
+        versionCode 121
         minSdkVersion 23
         targetSdkVersion 29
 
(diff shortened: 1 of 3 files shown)


Simplenote/build.gradle
```diff
@@ -30,9 +30,9 @@ android {
         if (project.hasProperty("versionName")) {
             versionName project.property("versionName")
         } else {
-            versionName "2.12-rc-1"
+            versionName "2.12-rc-2"
```
mokagio, line 33: Bumped to `rc-2` just in case we want to ship a new internal version

Simplenote/build.gradle
```diff
@@ -30,9 +30,9 @@ android {
...
         }
-        versionCode 120
+        versionCode 122
```
mokagio, line 35: `2.11.` was `121` so now we need to bump this again to make sure the build differs.

theck13 (approved):

PR #1184, opened 2020-10-23
Update Bucket Lifecycle

theck13:
### Fix
Update the bucket lifecycle by moving the `start()` and `stop()` statements from `Activity` and `Fragment` classes to the `Application` class.  These updates are meant to avoid sync issues caused by starting and stopping the buckets on screen changes.  By starting the buckets in the `onActivityResumed` method, the buckets are started when any screen is shown whether the app is started from the launcher, widget, or elsewhere.  Starting the buckets when they are already started does not affect syncing as shown in the [`Channel.start` method within Simperium](https://github.com/Simperium/simperium-android/blob/098bece0f6fe0f41bd13359b08f28482faf5a76d/Simperium/src/main/java/com/simperium/client/Channel.java#L610-L613).

These changes also remove the `mNotesBucket = noteBucket;` statement from the `onNetworkChange` method in the `NoteEditorFragment` class.  That statement was added in https://github.com/Automattic/simplenote-android/pull/1060 to fix tags being reverted when added/removed remotely.  Since the `mNotesBucket` variable was being used to retrieve the `Note` object of the updated note, assigning the `mNotesBucket` variable to the `noteBucket` parameter was required.  Removing that assigning statement and using the `noteBucket` parameter to retrieve the `Note` object maintains the fix from https://github.com/Automattic/simplenote-android/pull/1060 in my testing.  Also, it removes the only assignment of a `Bucket<Note>` variable that is not from the `Application` class.  Now, all instances of `Bucket<Note>` are from the `Simplenote` class.

### Test
Since these changes have no user-facing aspect, smoke testing is all that is required.  The changes affect the bucket starting and stopping so it's best to test creating and editing notes and tags to ensure syncing is working as expected across devices. Using multiple emulators or platforms side-by-side is a good way to see if syncing is working.

Since these changes also modify the changes from https://github.com/Automattic/simplenote-android/pull/1060, I'm including testing steps for that fix to make sure there isn't a regression.
1. ***Android***: Tap any note in list.
2. ***Other***: Select note from Step 1 in list.
3. ***Android***: Add `android` tag to note.
4. ***Other***: Notice `android` tag in note.
5. ***Other***: Add `other` tag to note.
6. ***Android***: Notice `other` tag in note.
7. ***Android***: Remove `other` tag from note.
8. ***Other***: Notice `other` tag removed from note.
9. ***Other***: Remove `android` tag from note.
10. ***Android***: Notice `android` tag removed from note.

### Review

#### Note
Contact me for an installable build.

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -1535,8 +1535,8 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
         if (changeType == Bucket.ChangeType.MODIFY) {
             if (getNote() != null && getNote().getSimperiumKey().equals(key)) {
                 try {
-                    mNotesBucket = noteBucket;
-                    final Note updatedNote = mNotesBucket.get(key);
+                    final Note updatedNote = noteBucket.get(key);
+
                     if (getActivity() != null) {
                         getActivity().runOnUiThread(new Runnable() {
                             @Override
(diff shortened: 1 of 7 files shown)


0nko (approved):
Nice work! Works exactly as described.

:shipit:

PR #1192, opened 2020-11-06
Update Tag Canonical

theck13:
### Fix
Update adding a tag to a note to use the canonical representation if the tag already exists.  If the canonical tag doesn't exist, a new tag is created.  Canonical refers to the hashed value of a tag added in https://github.com/Automattic/simplenote-android/pull/1092.  Lexical refers to variations of a tag that resolve to the same hashed value (e.g. tag, Tag, TAG, etc.).  When a lexical variation of an existing tag is entered into the **_Add tag..._** field at the bottom of a note, the tag is autocorrected to the canonical representation.

### Test
1. Tap any note in list.
2. Enter any value in **_Add tag..._** field not matching an existing tag.
3. Tap **_Space_** key to submit tag.
4. Notice tag is added to note as entered.
5. Enter any variation of value from Step 4 with different lowercase or uppercase characters.
6. Tap **_Space_** key to submit tag.
7. Notice tag is not added to note as entered.
8. Tap tag chip from Step 4.
9. Tap _**x**_ in tag chip.
10. Notice tag is removed from note.
11. Enter any variation of value from Step 2 with different lowercase or uppercase characters.
12. Tap **_Space_** key to submit tag.
13. Notice tag from Step 4 is added to note

### Review
Only one developer is required to review these changes, but anyone can perform the review.  @dmsnell, feel free to verify these changes are using the correct terminology and/or code, but don't feel obligated.

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -431,6 +431,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
         mContentEditText.setDropDownBackgroundResource(R.drawable.bg_list_popup);
         mContentEditText.setAdapter(mLinkAutocompleteAdapter);
         mTagInput = mRootView.findViewById(R.id.tag_input);
+        mTagInput.setBucketTag(((Simplenote) requireActivity().getApplication()).getTagsBucket());
         mTagInput.setDropDownBackgroundResource(R.drawable.bg_list_popup);
         mTagInput.setTokenizer(new SpaceTokenizer());
         mTagInput.setAdapter(mTagAutocompleteAdapter);
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
@@ -52,6 +52,27 @@ public class TagUtils {
     }
 
     /**
+     * Get the canonical representation of a tag from the hashed value of the lexical variation.
+     *
+     * @param bucket    {@link Bucket<Tag>} in which to get tag.
+     * @param lexical   {@link String} lexical variation of tag.
+     *
+     * @return          {@link String} canonical of tag if exists; lexical variation otherwise.
+     */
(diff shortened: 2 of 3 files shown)


mchowning (approved):
Looks good and works well @theck13 ! 👍 

Since you just opened this, I won't go ahead and merge this now so that @dmsnell has an opportunity to look at it if he wants. I'll plan to circle back on this after the weekend and merge this if there haven't been any comments.

Also, I did notice a very minor, very-much-an-edge case that doesn't seem ideal.
1. Add two tags to a note: `aa1` and `aa2`
2. Exit the note and open the edit tags screen
3. Edit the `aa1` tag to be `Aa`
4. Observe that the `aa1` tag is removed from the tag list and the `Aa` tag is added as you would expect
5. Edit the `aa2` tag to be `aA` (i.e., so that it is a lexical variation of the recently updated `Aa` tag)
6. Observe that the `aa2` and `Aa` tags are removed, and the quantity for the new `aA` tag is 1. I was a little surprised to see the quantity of 1 instead of 2 at first, but once I realized this was probably because both tags had been attached to the same note, it made sense.
7. Go back to the original note and observe that it has 2 tags: `Aa` and `aA`. I did not expect to see the `Aa` tag anymore since it was no longer included in the list of tags on the Edit Tags screen.
8. Add a third tag to that note that is _not_ a lexical variation of the note's current tags (i.e., `b`).
9. Observe that **when you add the third `b` tag, the `Aa` tag disappears**, leaving only the `aA` and `b` tags. The tag that seems to disappear in Step 9 is what seems slightly concerning to me because it gives the impression that the user has done something to remove a tag. Seems like it might be nice to clean any duplicates out of the tag list when the note is opened instead of only after a tag is added. 

I certainly don't consider this a blocker for this PR, just sharing so you're aware.

theck13:
Thanks for the thorough testing and explanation!

That sounds like a known issue with tag renaming, which would be fixed with https://github.com/Automattic/simplenote-android/pull/1112.  That pull request was put on hold until we decided exactly what we're going to do about conflicting tags.  The first step is using the canonical version only as the changes in this pull request address.  The second step is updating https://github.com/Automattic/simplenote-android/pull/1112 to prompt the user when renaming conflicts occur.  That will avoid the situation when renaming a tag to an existing tag as you found in your testing.  Let me know if that doesn't make sense or I misunderstood your test.

PR #1194, opened 2020-11-10
Update Checkbox Metrics

theck13:
### Fix
Update the color and size of checkboxes in the editor to indicate they are tappable and easier to tap, respectively.  All checkboxes in lists including the **_Select Note_** dialog of the **_Note (Light)_** and **_Note (Dark)_** widgets remain unchanged in both color and size.  See the screenshots below for illustration.

update_checkbox_metrics

The translucent square behind the checkbox shows the tappable area.  Notice the small space between lines in both **_Before_** and **_After_** columns.  That is the line height spacing value, which applies to all lines in the editor and not only those with checkboxes.

### Test
1. Notice notes with checkboxes remain unchanged in note list.
2. Tap any note with checkboxes in list.
3. Notice checkboxes match accent color of theme (i.e. `simplenote_blue_50`/`#3361cc` for **_Light_** and `simplenote_blue_20`/`#84a4f0` for **_Dark_**).
4. Tap any checkbox in note.
5. Notice checkbox toggles from checked to unchecked or unchecked to checked.
6. Notice area approximately half size of checkbox around checkbox is tappable.

### Review

### Release


SylvesterWilmott:
These are much easier to press than what we currently have!

I'm curious how much we can tweak these before they start to be difficult to touch again. Although they are easier to touch the spacing between each item in a checklist really breaks the visual rhythm of a note. I think this is because of the additional space between each item. Can we try reducing the built-in padding on the check asset by the total number of pixels in each space?

theck13:
> Can we try reducing the built-in padding on the check asset by the total number of pixels in each space?

I'm not completely sure I understand what you mean.  The [current line spacing extra value is 4dp](https://github.com/Automattic/simplenote-android/blob/fbcec822f5a75646849a441a8d6bc9a496c718c9/Simplenote/src/main/res/values/dimens.xml#L49).  Are you saying to reduce the internal padding of the checkbox drawables by 2dp on the top and bottom?  That would look something like this.

update_checkbox_metrics_

theck13:
After chatting with @SylvesterWilmott, we decided to reduce the internal padding of the checkbox drawables by 2dp on all sides, only color the checked state blue, and reduce the checkbox extra size from 16 to 12.  See the screenshots below for illustration.

update_checkbox_metrics_final

SylvesterWilmott (approved):
Still much easier to tap each one and the rhythm fits more nicely with the paragraph text. Thanks!

ThomazFB (approved):
Looking good, Tyler! Code reviewed and checkboxes tested! :shipit:

PR #1218, opened 2020-12-01
Update Tag Hash

theck13:
### Fix
Update the tag hash methods to include the `-`, `.`, and `_` characters in encoding.  That ensures the encoded tag hash value is consistent across platforms.

### Test
The best way to test whether all the proper characters are encoded is to add a tag with the `:/?#[]@!$&'()*+,;=-._~` text and verify that all characters are encoded in the database.  Contact me for database access.
1. Tap any note in list.
2. Enter `:/?#[]@!$&'()*+,;=-._~` text in ***Add tag...*** field.
3. Tap **_Space_** key to submit tag.
4. Notice `:/?#[]@!$&'()*+,;=-._~` tag is added to note.
5. Notice all characters in tag key are encoded in database.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
@@ -118,6 +118,11 @@ public class TagUtils {
      * @return          {@link String} replaced characters with encoded values.
      */
     private static String replaceEncoded(String encoded) {
-        return encoded.replace("+", "%20").replace("*", "%2A").replace("-", "%2D").replace(".", "%2E").replace("_", "%5F");
+        return encoded
+            .replace("+", "%20")
+            .replace("*", "%2A")
+            .replace("-", "%2D")
+            .replace(".", "%2E")
+            .replace("_", "%5F");
     }
 }
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
@@ -84,7 +84,7 @@ public class TagUtils {
             String normalized = Normalizer.normalize(name, Normalizer.Form.NFC);
             String lowercased = normalized.toLowerCase(Locale.US);
             String encoded = URLEncoder.encode(lowercased, StandardCharsets.UTF_8.name());
-            return encoded.replace("*", "%2A").replace("+", "%20");
+            return encoded.replace("*", "%2A").replace("+", "%20").replace("-", "%2D").replace(".", "%2E").replace("_", "%5F");
         } catch (UnsupportedEncodingException e) {
             // TODO: Handle encoding exception with a custom UTF-8 encoder.
             return name;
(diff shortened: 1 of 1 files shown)


ParaskP7:
> Contact me for database access.
👋 @theck13 ! How can I get access to the database to verify correctness?

Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
```diff
@@ -102,11 +102,22 @@ public static boolean hashTagValid(String name) {
...
+
+    /**
+     * Replace certain characters in @param encoded that were not encoded with encoded value.
+     *
+     * @param encoded   {@link String} to replace certain characters with encoded value.
+     *
+     * @return          {@link String} replaced characters with encoded values.
+     */
+    private static String replaceEncoded(String encoded) {
+        return encoded.replace("+", "%20").replace("*", "%2A").replace("-", "%2D").replace(".", "%2E").replace("_", "%5F");
```
ParaskP7: 1. Minor (🔍 ): Consider formatting this line to multilines for better readability and for not having this line exceed a certain limit of character (ie. 120 character). Something like the below:
```
return encoded
                .replace("+", "%20")
                .replace("*", "%2A")
                .replace("-", "%2D")
                .replace(".", "%2E")
                .replace("_", "%5F");
```
2. Warning (⚠️ ): I understand that this is not part of this PR's change, but after looking it up I found that the encoded character for plus is the `%2B` instead of `%20`. Wdyt?

ParaskP7 (commented):
Thanks for including me in this PR @theck13 ! 🌟

I just have a couple of  comments for you.

Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
theck13: 1. There are no code style requirements like 120 characters limits in Simplenote.  There are a ton of lines longer than 120 characters.  Making the single line more readable by breaking it into multiple lines is subjective.  I updated the code style in 1525f768.

2. Actually, the `encoded` parameter has already been encoded upstream, which means plus characters are already `%2B` and space characters are `+`.  That's why we replace the `+` character in the `encoded` parameter with `%20` since it was a space character originally.  You can verify that by putting a breakpoint or log statement at the [`return encoded...` statement](https://github.com/Automattic/simplenote-android/pull/1218/files#diff-95498cc2ef506f4683acbb12c4a8e1af8cb36a0b6bd63656dec1a744a4b54507R121) and [adding `name = " +";` to the `try` statement in the `hashTagValid` method](https://github.com/Automattic/simplenote-android/pull/1218/files#diff-95498cc2ef506f4683acbb12c4a8e1af8cb36a0b6bd63656dec1a744a4b54507R102).  You'll see the `encoded` parameter will be `+%2B`.  I added details to the comment of the `replaceEncoded` method in 588dd4b8.

theck13 (commented):
> 👋 @theck13 ! How can I get access to the database to verify correctness?

We discussed this over direct message.  Let me know if you need help accessing the database.

Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
ParaskP7: 1) Thanks for the change Tyler! I didn't know that there is no such requirement like 120 characters limits in Simplenote, good to know. Indeed all style related comments can be subjective, thus I am trying to use `Minor` with an emoji when commenting on those meaning that you can ignore as minor. Feel free to do so next time.
2) Thanks for the explanation Tyler! Also, thanks for actually included the reasoning behind it as a Javadoc.

ParaskP7 (approved):
Tested the solution and it works as expected, verified the result in the database as well! 🌟

PR #1224, opened 2020-12-03
Add Delete Action

theck13:
### Fix
Add the ***Delete*** action to the overflow/ellipsis menu to permanently delete notes in the ***Trash*** individually.  Currently, notes are only able to be permanently deleted in bulk using the ***Empty Trash*** action.  ***Delete*** will only be shown once the note has been trashed and will be above the ***Restore*** action.  See the screenshots below for illustration.

#### Phone
add_delete_action_phone

#### Tablet
add_delete_action_tablet

Also included in these changes is an update to the ***Empty Trash*** dialog message.  Since emptying the trash is a destructive action that permanently deletes all notes in ***Trash*** and it cannot be undone like trashing a note, a more verbose message explaining exactly that seems necessary.  See the screenshots below for illustration.

add_delete_action_empty

### Test
Since the actions are different on phones and tablets, it's best to test both device form factors to verify the correct actions are visible/invisible as shown in the screenshots above.
1. Tap any note in list.
2. Tap overflow menu action in app bar.
3. Notice ***Trash*** action is shown.
4. Notice ***Delete*** action is hidden.
5. Tap outside overflow menu.
6. Open navigation drawer.
7. Tap ***Trash*** item in navigation drawer.
8. Tap ***Empty Trash*** action in app bar (phone) or overflow menu (tablet).
9. Notice ***Empty Trash*** dialog is shown with messag shown above.
10. Tap ***No*** button in dialog.
11. Notice ***Trash*** is not emptied.
12. Tap any note in list.
13. Tap overflow menu action in app bar.
14. Notice ***Restore*** action is shown.
15. Notice ***Delete*** action is shown.
16. Tap ***Delete*** action in overflow menu.
17. Notice ***Delete Permanently*** dialog is shown.
18. Tap ***No*** button in dialog.
19. Notice ***Delete Permanently*** dialog is hidden.
20. Notice note is not deleted.
21. Tap overflow menu action in app bar.
22. Tap ***Delete*** action in overflow menu.
23. Notice ***Delete Permanently*** dialog is shown.
24. Tap ***Yes*** button in dialog.
25. Notice note is deleted.

### Review

#### Note
Contact me for an installable build.

### Release
`RELEASE-NOTES.txt` was updated in 259e51e with:
> Added permanently deleting notes from trash individually [https://github.com/Automattic/simplenote-android/pull/1224]

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
@@ -1,14 +1,25 @@
 package com.automattic.simplenote.utils;
 
 import android.app.Activity;
+import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
+import android.text.method.LinkMovementMethod;
+import android.widget.TextView;
 
+import androidx.appcompat.app.AlertDialog;
+import androidx.appcompat.view.ContextThemeWrapper;
+
+import com.automattic.simplenote.NoteEditorActivity;
+import com.automattic.simplenote.NotesActivity;
+import com.automattic.simplenote.R;
 import com.automattic.simplenote.Simplenote;
 import com.automattic.simplenote.analytics.AnalyticsTracker;
 import com.automattic.simplenote.models.Note;
 
 import java.text.NumberFormat;
 import java.util.Calendar;
+import java.util.Objects;
 
 public class NoteUtils {
     public static void setNotePin(Note note, boolean isPinned) {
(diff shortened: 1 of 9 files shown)


Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
```diff
@@ -53,4 +64,48 @@ public static String getWordCount(String content) {
...
+                    public void onClick(DialogInterface dialog, int whichButton) {
+                        if (note != null) {
+                            note.delete();
+
+                            // Show empty placeholder for large devices in landscape.
+                            if (activity instanceof NotesActivity) {
+                                ((NotesActivity) activity).showDetailPlaceholder();
+                            // Close editor for small devices and large devices in portrait.
+                            } else if (activity instanceof NoteEditorActivity) {
+                                ((NoteEditorActivity) activity).finish();
```
ParaskP7, line 85: Minor (🔍 ): Consider removing the cast to `NoteEditorActivity` for this case, it is not needed, isn't it?

Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
```diff
@@ -53,4 +64,48 @@ public static String getWordCount(String content) {
...
+                        } else {
+                            showDialogErrorDelete(activity);
+                        }
+                    }
+                }
+            )
+            .show();
+    }
+
+    private static void showDialogErrorDelete(Context context) {
```
ParaskP7, line 96: Idea (💡): I am seeing that this is a duplicate of the `TagDialogFragment.showDialogErrorRename()' method, isn't it? Maybe consider removing the duplication by reusing this one. Wdyt?

ParaskP7 (commented):
Great job @theck13 ! 🌟 

Only minor and idea comments from my side.

SylvesterWilmott:
Looking good!

Thanks for writing the copy for those dialogs, looks clean and concise. I am wondering if we should state the action in the button text.

Delete Confirm:
[Cancel] [Delete]

Empty Trash:
[Cancel] [Empty]

Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
theck13: That's correct.  Since `finish` is available to all `Activity` classes, it's not required to cast `NoteEditorActivity` here.  I try to be a little extra verbose with cases like this when we're doing different things based on layout.  It's not always apparent that Simplenote uses `NotesActivity` in some cases (i.e. large devices in landscape orientation) and `NoteEditorActivity` in some cases (i.e. small devices and large devices in portrait orientation).  Having the `NoteEditorActivity` cast may help someone realize the difference even if the code is refactored and the comments are removed.

Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
theck13: I noticed that too.  The [`delete_dialog_error_message_email`](https://github.com/Automattic/simplenote-android/blob/259e51e5ac495ed4bd0fcd313620197c4c48ce76/Simplenote/src/main/res/values/strings.xml#L45) string resource is also a duplicate of [`rename_tag_message_email`](https://github.com/Automattic/simplenote-android/blob/259e51e5ac495ed4bd0fcd313620197c4c48ce76/Simplenote/src/main/res/values/strings.xml#L21).  I considered consolidating the methods and resources, but I didn't want to creep too far out of scope of this pull request.  I already crept a little by changing the **_Empty Trash_** dialog message.  I'll combine the methods and resources in a subsequent pull request.

theck13:
> I am wondering if we should state the action in the button text.

That's a good idea!  I updated the dialog buttons in 7d8a1207 and 5266af6a.  See the screenshots below for illustration.

add_delete_action_buttons

SylvesterWilmott (approved):
Thanks! Looks good.

Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
ParaskP7: Great, thanks for explaining the reasoning behind that, it makes sense!

However, bear in mind that static analysis tools will not like that and that it might create confusion for other devs, which like me, might try to verify that everything is okay within a class, traversing it by error/warnings/spelling mistakes and correct anything suspicious. For example, if I where to touch this class at some point in time I would have definitely refactored this code, since it is being greyed out on my IDE and appears as warning.

Another idea to be extra verbose, which also allows getting rid of the comments altogether, is extracting the logic in private methods. I would have done that instead and then cover both cases:
```
if (activity instanceof NotesActivity) {
    showPlaceholderInLandscapeLargeDevices((NotesActivity) activity);
} else if (activity instanceof NoteEditorActivity) {
    closeEditorInPortrait((NoteEditorActivity) activity);
}
```

ParaskP7 (approved):

PR #1225, opened 2020-12-03
Crash Device Language Change

### LINKED ISSUE #959: Crash: Changing device language with a note open crashes the app
The app crashes when I switch between languages with a note open. I observed this while switching between UK English and either Spanish or German.

With a markdown note open, I got this NullPointerException: `Attempt to invoke virtual method 'void com.automattic.simplenote.NoteMarkdownFragment.updateMarkdown(java.lang.String)' on a null object reference`. In Sentry: [SIMPLENOTE-ANDROID-11B](https://sentry.io/share/issue/09db54c3cf3a49528b72eef209d823b8/)

With a non-markdown note open, I got this NullPointerException: `Attempt to invoke virtual method 'boolean androidx.fragment.app.Fragment.isAdded()' on a null object reference`. In Sentry: [SIMPLENOTE-ANDROID-11C](https://sentry.io/share/issue/57caf1efe44245ccb63ed8bf2aa0b97e/)

### Reproduced

1. Open the app and select a markdown note in the notes list to open the editor.
2. Change your device language (e.g. from UK English to Spanish).
3. Return to the app. Result: Crash.
4. Open the app again and select a non-markdown note.
5. Change your device language (e.g. from Spanish back to UK English).
6. Return to the app. Result: Crash.

-|-|-|-
Motorola|moto e5 play|8.1.0|2.4-rc-3

There are no other reports of either of these exact crashes, and I wouldn't expect users to commonly switch languages at this point, but I wanted to report it in case resolving this case could help with resolving other, more common crashes.

ParaskP7:
Fixes #959 

### Fix
This PR fixes a crash on changing device language with a note open.

The crash was caused due to the fact that, when saving a tab during 'onSaveInstanceState(...)' a key of a specific language was used (ie. 'Edit' if the device was on English at that point), but when restoring a tab during 'onCreate(...)' a key of another language was used (ie. 'Editar' if the device was switched to Spanish). Thus, no restore actually occurred on the tab layer, causing the crash.

Using a constant as a key for saving/restoring state fixes this issue.

### Test
1) Open the app and select a markdown note in the notes list to open the editor.
1) Change your device language (e.g. from UK English to Spanish).
1) Return to the app. Result: App **NOT** crashing and language change is visible on tabs.
1) Open the app again and select a non-markdown note.
1) Change your device language (e.g. from Spanish back to UK English).
1) Return to the app. Result: App **NOT** crashing and language change is visible in app.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
@@ -50,6 +50,8 @@ import static com.automattic.simplenote.utils.WidgetUtils.KEY_LIST_WIDGET_CLICK;
 import static com.automattic.simplenote.utils.WidgetUtils.KEY_WIDGET_CLICK;
 
 public class NoteEditorActivity extends ThemedAppCompatActivity {
+    private static final String STATE_TAB_EDIT = "TAB_EDIT";
+    private static final String STATE_TAB_PREVIEW = "TAB_PREVIEW";
     private static final String STATE_MATCHES_INDEX = "MATCHES_INDEX";
     private static final String STATE_MATCHES_LOCATIONS = "MATCHES_LOCATIONS";
     private static final int INDEX_TAB_EDIT = 0;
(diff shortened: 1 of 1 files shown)


theck13 (approved):
The code looks good to me and the tests work as expected.  Great fix!

PR #1226, opened 2020-12-04
Add Dialog Utils

theck13:
### Fix
Add a dialog utility class to consolidate showing a dialog with a link to email support.  These changes deduplicate code and creates a class for combining reused code with dialogs throughout the codebase.

### Test
There are no interface changes in this pull request.  All existing dialogs containing a link to email support should work as they did previously.  The three cases to test are below.
#### Add Tag
1. Tap any note in list.
2. Enter four England flag emojis (`🏴󠁧󠁢󠁥󠁮󠁧󠁿🏴󠁧󠁢󠁥󠁮󠁧󠁿🏴󠁧󠁢󠁥󠁮󠁧󠁿🏴󠁧󠁢󠁥󠁮󠁧󠁿`) in ***Add tag...*** field.
3. Tap ***Space*** key.
4. Notice ***Error*** dialog is shown with link to email support.

#### Delete Note
1. Tap any note in list.
2. Tap overflow menu action in app bar.
3. Tap ***Trash*** item in overflow menu.
4. Open navigation drawer.
5. Tap ***Trash*** item in navigation drawer.
6. Tap overflow menu action in app bar.
7. Tap ***Delete*** item in overflow menu.
8. Notice ***Delete Permanently*** dialog is shown with link to email support.

#### Rename Tag
While going through this test case, I noticed a crash occurs if the **_Error_** dialog is shown when renaming a tag encounters an error.  I developed a fix for the crash, but I reverted the commits and I'll submit the crash fix in a separate pull request in order to avoid scope creep in this pull request.

### Review

### Release


Simplenote/src/main/java/com/automattic/simplenote/utils/DialogUtils.java
```diff
@@ -0,0 +1,38 @@
...
+public class DialogUtils {
+    /**
+     * Show an alert dialog with a link to the support@simplenote.com email address,
+     * which can be tapped to launch the device's email app.
+     *
+     * @param context   {@link Context} from which to determine theme and resources.
+     * @param title     {@link String} for the dialog title.
+     * @param message   {@link String} for the dialog message.
+     */
+    public static void showDialogWithEmail(Context context, String title, String message) {
```
ParaskP7: Star (🌟) Great job on extracting this duplicate logic from everywhere!

Idea (💡 ): Since the `title` is alway this `R.string.error` string, and based on the fact that you already decided to use the `R.string.support_email` within, which is currently only used for such errors, consider removing the `title` from the method arguments and hardcode it, just like you did for `R.string.support_email`. Then, the method can change to something like `showErrorDialogWithEmail(...)` instead.

ParaskP7 (commented):
Great job! 🌟 

I have only one suggestion to make.

Question (❓ ): Also, I am seeing a lot of additional commits that was added and then reverted. Why not dropping the commits altogether and instead reverting them?

Simplenote/src/main/java/com/automattic/simplenote/utils/DialogUtils.java
theck13: I had the `title` parameter so that future uses of the `showErrorDialogWithEmail` method could choose the title.  We could move `R.string.support_email` to a parameter too so that any email address could be used.  Since we're only using `R.string.error` for `title` at the moment, I removed the parameter in febd7eb4.

theck13 (commented):
> Why not dropping the commits altogether and instead reverting them?

Since those commits are needed to fix a crash in a subsequent pull request, I branched off `add-dialog-utils` before reverting those commits so that I would have to redo all of them again.

ParaskP7 (approved):

PR #1228, opened 2020-12-07
Crash Rename Tag

theck13:
### Fix
Update the `TagDialogFragment` class to avoid an `IllegalStateException` crash.  While going through the tests in https://github.com/Automattic/simplenote-android/pull/1226, I noticed a crash can occur when renaming a tag encounters an error.  When the **_Tag Conflict_** `AlertDialog` was shown and the `DialogFragment` was dismissed, the fragment is detached from the activity.  Therefore, when the **_Error_** dialog was shown when an error occurred renaming a tag, the context used was no longer valid and an `IllegalStateException` was thrown with the "Fragment TagDialogFragment not attached to a context" message.  This was introduced in https://github.com/Automattic/simplenote-android/pull/1112, which has not been released yet.  To fix the crash, I modified `TagDialogFragment` to show and hide views based on renaming a tag or a tag conflict.  Now, the `DialogFragment` will only be dismissed/detached after determining there is no tag conflict or the user has chosen to merge conflicting tags, which means the context used for the **_Error_** dialog will be valid.

#### Note
Branching off `add-dialog-utils` didn't do what I expected.  It effectively made the changes in the original `crash-rename-tag` branch null.  I had to branch off the latest `develop` and revert the revert, which is not what I wanted.  That's why the commits in this pull request are a little confusing.

### Test
With the changes to the **_Rename Tag_** dialog, an invalid tag should not be able to be entered and the ***Error*** dialog should never be shown.  In order to test the ***Error*** dialog, the `try` part of the [`try`/`catch`/`finally` statement in the `TagDialogFragment` class](https://github.com/Automattic/simplenote-android/blob/ae3f699ba9b659febb819f68a40dbac52a38463c/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java#L209-L226) can be commented out to force the ***Error*** dialog to be shown.
1. Open navigation drawer.
3. Tap any tag in list.
4. Edit tag name to any non-existing tag.
5. Tap ***Save*** button in dialog.
6. Notice ***Error*** dialog is shown with link to email support.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
@@ -36,14 +36,15 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
     private Bucket<Note> mBucketNote;
     private Bucket<Tag> mBucketTag;
     private Button mButtonNegative;
+    private Button mButtonNeutral;
     private Button mButtonPositive;
     private String mTagOld;
     private Tag mTag;
     private TextInputEditText mEditTextTag;
     private TextInputLayout mEditTextLayout;
     private TextView mMessage;
-    private View.OnClickListener mClickListenerNegativeConflict;
     private View.OnClickListener mClickListenerNegativeRename;
+    private View.OnClickListener mClickListenerNeutralConflict;
     private View.OnClickListener mClickListenerPositiveConflict;
     private View.OnClickListener mClickListenerPositiveRename;
 
@@ -110,7 +111,7 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
             }
         };
 
-        mClickListenerNegativeConflict = new View.OnClickListener() {
+        mClickListenerNeutralConflict = new View.OnClickListener() {
             @Override
             public void onClick(View v) {
                 showDialogRenameTag();
@@ -132,6 +133,7 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
     @Override
     public void onShow(DialogInterface dialog) {
         mButtonNegative = mDialogEditTag.getButton(DialogInterface.BUTTON_NEGATIVE);
+        mButtonNeutral = mDialogEditTag.getButton(DialogInterface.BUTTON_NEUTRAL);
         mButtonPositive = mDialogEditTag.getButton(DialogInterface.BUTTON_POSITIVE);
         showDialogRenameTag();
         mEditTextTag.setText(mTag.getName());
@@ -184,22 +186,32 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
     }
 
     private void showDialogErrorConflict(String canonical, String tagOld) {
+        mDialogEditTag.setTitle(R.string.dialog_tag_conflict_title);
+
         mMessage.setText(getString(R.string.dialog_tag_conflict_message, canonical, tagOld, canonical));
+        mButtonNeutral.setText(R.string.back);
+        mButtonPositive.setText(R.string.dialog_tag_conflict_button_positive);
+
         mMessage.setVisibility(View.VISIBLE);
         mEditTextLayout.setVisibility(View.GONE);
-        mDialogEditTag.setTitle(R.string.dialog_tag_conflict_title);
-        mButtonNegative.setText(R.string.cancel);
-        mButtonPositive.setText(R.string.dialog_tag_conflict_button_positive);
-        mButtonNegative.setOnClickListener(mClickListenerNegativeConflict);
+        mButtonNegative.setVisibility(View.GONE);
+        mButtonNeutral.setVisibility(View.VISIBLE);
+
+        mButtonNeutral.setOnClickListener(mClickListenerNeutralConflict);
         mButtonPositive.setOnClickListener(mClickListenerPositiveConflict);
     }
 
     private void showDialogRenameTag() {
-        mMessage.setVisibility(View.GONE);
-        mEditTextLayout.setVisibility(View.VISIBLE);
         mDialogEditTag.setTitle(R.string.rename_tag);
+
         mButtonNegative.setText(R.string.cancel);
         mButtonPositive.setText(R.string.save);
+
+        mMessage.setVisibility(View.GONE);
+        mEditTextLayout.setVisibility(View.VISIBLE);
+        mButtonNegative.setVisibility(View.VISIBLE);
+        mButtonNeutral.setVisibility(View.GONE);
+
         mButtonNegative.setOnClickListener(mClickListenerNegativeRename);
         mButtonPositive.setOnClickListener(mClickListenerPositiveRename);
     }
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
@@ -4,6 +4,7 @@ import android.app.Dialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.DialogInterface.OnShowListener;
+import android.content.res.Configuration;
 import android.os.Bundle;
 import android.text.Editable;
 import android.text.TextWatcher;
@@ -50,6 +51,12 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
         mTag = tag;
         mBucketNote = bucketNote;
         mBucketTag = bucketTag;
+    }
+
+    @Override
+    public void onConfigurationChanged(@NonNull Configuration newConfig) {
+        super.onConfigurationChanged(newConfig);
+        dismiss();
     }
 
     @NonNull
@@ -209,6 +216,8 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
                 context,
                 context.getString(R.string.rename_tag_message)
             );
+        } finally {
+            dismiss();
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
```diff
@@ -33,52 +35,42 @@
...
     private Button mButtonPositive;
     private String mTagOld;
     private Tag mTag;
     private TextInputEditText mEditTextTag;
     private TextInputLayout mEditTextLayout;
+    private TextView mMessage;
+    private View.OnClickListener mClickListenerNegativeConflict;
+    private View.OnClickListener mClickListenerNegativeRename;
+    private View.OnClickListener mClickListenerPositiveConflict;
+    private View.OnClickListener mClickListenerPositiveRename;
```
ParaskP7, line 49: Question (❓  ): Is the Hungarian notation still used in this repo? I have seen elsewhere that it is a mixture of using it so I am not sure. If it has been decided to not use that anymore then consider dropping the `m` prefix from all those new fields.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
```diff
@@ -87,12 +79,62 @@ public void onClick(DialogInterface dialog, int whichButton) {
...
+                }
+
+                tryToRenameTag(tagNew, index);
+            }
+        };
+
+        mClickListenerNegativeConflict = new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                showDialogRenameTag();
```
ParaskP7, line 117: Question (❓ ): This is a new behavior, isn't it?

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
```diff
@@ -87,12 +79,62 @@ public void onClick(DialogInterface dialog, int whichButton) {
...
+            }
+        };
+
+        mClickListenerPositiveConflict = new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                String tagNew = getTagNew();
+                int index = mTag.hasIndex() ? mTag.getIndex() : mBucketTag.count();
+                tryToRenameTag(tagNew, index);
+                dismiss();
```
ParaskP7: Question (❓ ): Why adding the extra `dismiss()` here? This can be probably removed since `dismiss()` would anyway be triggered as part of the finally clause within `tryToRenameTag(...)`, isn't that right?

ParaskP7 (commented):
Cool crash fix @theck13 ! 🌟 

I have added some individual comments/questions for you. In addition to those, I have the below:
1) I was only able to reproduce this issue by replacing the `if (hasCanonicalTag && !isRenamingToLexicalTag) { ... }` with `if (true) { ... }` to force both dialogs to show, one after the other, otherwise, the crash never occurs, even in `develop`. Maybe the test instructions should be updated. Wdyt?
2) When I try to rename a tag, and it succeeds, I am not seeing it being renamed. I even tried that on `develop`. Is this broken or am I not understanding the feature?

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
theck13: Yes, Hungarian notation is still used in this codebase.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
theck13: Yes, tapping the negative button in the dialog, which is currently ***Cancel***, reverts to the ***Rename Tag*** view.  Since the dialog was updated to show/hide views between ***Rename Tag*** and ***Tag Conflict*** rather than using separate dialogs, I chose to add that behavior.  @SylvesterWilmott, what do you think of that interaction (i.e. returning to the previous dialog view allowing the user to rename the tag again rather than dismissing the dialog completely)?  If we keep that interaction, should we change the ***Cancel*** button to ***Back*** or something similar so be a little clearer to the user what will happen?  See the animation below for illustration.

<kbd><a href="https://user-images.githubusercontent.com/3827611/101591324-89421e00-39a9-11eb-80ce-1da9ebdee29d.gif"></a></kbd>

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
theck13: I forgot to delete that `dismiss();` statement after adding it to the `finally` clause in `tryToRename`.  I removed the extra statement in 6de2d25a.

theck13 (commented):
> 1. I was only able to reproduce this issue by replacing the `if (hasCanonicalTag && !isRenamingToLexicalTag) { ... }` with `if (true) { ... }` to force both dialogs to show, one after the other, otherwise, the crash never occurs, even in `develop`. Maybe the test instructions should be updated. Wdyt?

As mentioned in the ***Test*** section, the ***Error*** dialog should never be shown in this branch or `develop` if the code isn't modified since an invalid tag should not be able to be entered.  However, I forgot that the dialog is dismissed `if (tagNew.equals(mTagOld))`.  So, the initial test steps wouldn't work exactly.  I updated Step 4.  Now, commenting out the `try` part of the [`try`/`catch`/`finally` statement in the `TagDialogFragment` class](https://github.com/Automattic/simplenote-android/blob/ae3f699ba9b659febb819f68a40dbac52a38463c/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java#L209-L226) will show the ***Error*** dialog.

> 2. When I try to rename a tag, and it succeeds, I am not seeing it being renamed. I even tried that on `develop`. Is this broken or am I not understanding the feature?

I'm seeing renamed tags being renamed in this branch and `develop`.  See the animation below for illustration.

<kbd><a href="https://user-images.githubusercontent.com/3827611/101592282-38332980-39ab-11eb-8645-e8f162414b48.gif"></a></kbd>

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
ParaskP7: So, for new fields should we use the Hungarian notation or not? I am asking that because I assumed we shouldn't and I haven't included it on the fields I created so far.

ParaskP7:
> As mentioned in the Test section, the Error dialog should never be shown in this branch or develop if the code isn't modified since an invalid tag should not be able to be entered. However, I forgot that the dialog is dismissed if (tagNew.equals(mTagOld)). So, the initial test steps wouldn't work exactly. I updated Step 4. Now, commenting out the try part of the try/catch/finally statement in the TagDialogFragment class will show the Error dialog.

Thank you! 🌟 

> I'm seeing renamed tags being renamed in this branch and develop. See the animation below for illustration.

You are right, I was able to rename some tags are well. But, I have other tags that I can't rename. For example, I have a `Thing` tag that I can rename to `Thingsss` but I also have a `So` tag that I can't rename to anything. Probably a sync issue? Wdyt?

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
theck13: Since the codebase is 89% Java and 0% Kotlin, I would use Hungarian notation even though it's not a strictly required code style in Simplenote Android.  Hungarian is the accepted notation for Java and camel case is the accepted notation for Kotlin.  In this file, all variables use Hungarian notation.  So, keeping that notation makes sense to me.

theck13:
> You are right, I was able to rename some tags are well. But, I have other tags that I can't rename. For example, I have a `Thing` tag that I can rename to `Thingsss` but I also have a `So` tag that I can't rename to anything. Probably a sync issue? Wdyt?

That may be an issue with your account or those tags specifically.  We can troubleshoot that with the database if you would like.  The changes in this pull request do not change the way tags are renamed.  That occurs in [`Tag.renameTo`](https://github.com/Automattic/simplenote-android/blob/efc5c5ba96be717bee02d5baabc311acc406c2a9/Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java#L80-L130).  This pull request only changes the crash caused by the ***Error*** dialog.

ParaskP7:
> That may be an issue with your account or those tags specifically. We can troubleshoot that with the database if you would like. The changes in this pull request do not change the way tags are renamed. That occurs in Tag.renameTo. This pull request only changes the crash caused by the Error dialog.

Definitely, thank you! I'll approve the PR now. 🌟

ParaskP7 (approved):

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
ParaskP7: I see how you are thinking about it, thanks for sharing Tyler! 🌟 

But, Hungarian notation is language-independent, it doesn't have to be Java vs Kotlin.

When I argue about that, I found this [article](https://jakewharton.com/just-say-no-to-hungarian-notation/) very enlighting. I can add very little to that.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
SylvesterWilmott: The interaction of returning to the renaming dialog from the confirmation dialog is ideal. Ideally the button would describe the action such as "Choose another name" but this label is far too long. I think in this case "Back" works just fine.

theck13:
After chatting with @SylvesterWilmott, we decided to use the "Back" text in the **_Tag Conflict_** dialog and move that button to the neutral position.  That may be more familiar for users going back since the back arrow in the app and system navigation is on the start/left side.  Also, moving to that location helps the action differentiate from the negative button action, which typically dismisses a dialog.  See the animation below for illustration.

<kbd><a href="https://user-images.githubusercontent.com/3827611/101664112-b1149e80-3a08-11eb-8dd9-2d2325f923a9.gif"></a></kbd>

Changes were made after your approval, @ParaskP7.  So, I re-requested your review.

ParaskP7 (approved):
Looks good, thanks for grinding on it! 🌟

PR #1229, opened 2020-12-08
Right Arrow Hardware Keyboard Key Blanks Out Note Editor

### LINKED ISSUE #472: Pressing 'Right Arrow' Key Blanks Out Note Editor
User report:

https://cloudup.com/cnp8V942ABL

I wonder if that is switching to the markdown view without triggering the code that renders the markdown content?

ParaskP7:
Fixes #472 

### Fix
This PR fixed the bug which causes the note editor to blank out when markdown is disabled.

The problem was that, when markdown was disabled, the note editor will still switch to the preview tab when the `right arrow` hardware keyboard key was pressed, when being at the end of the text editor's selection. This behavior makes sense when the markdown is enabled, but should not be allowed when markdown is disabled.

The bug is effectively fixed by disabling the `right arrow` hardware keyboard key when being at the end of the text editor's selection. Also, the same bug occurs on the tags section, and as such, this was fixed as well.

For the tags section, it was consider to not allow switching tabs even when markdown was enabled, but this was dropped because when markdown is enabled, clicking on the `right arrow` hardware keyboard key, when being at the end of the text editor's selection, the cursor jumps down to the tags sections instead, and from there if the user clicks the `right arrow` hardware keyboard key again, they can switch to the markdown tab. Since this might be intended, I decided to keep this functionality untampered.

### Note
I considered RTL support as well, but giving the fact that the text editor is not to a 100% RTL ready, adding the fix for RTL will break current functionality. This is due to the fact that when adding a text while on RTL, the text starts from left to right anyway, and as such the `right arrow` hardware keyboard key solution continues to apply. In order to improve that, the text editor needs to first start supporting RLT for that scenario. @theck13 please correct me if I am wrong. 🙏 

### Test
1) Use an emulator with keyboard input enabled.
1) Tap `New Note` floating action button.
1) Make sure that the `Markdown` is disabled on this note.
1) Enter `Test` in new note.
1) Notice `Test` is displayed.
1) Notice software keyboard is shown.
1) Press `right arrow` key on hardware keyboard.
1) Notice software keyboard is **NOT** dismissed.
1) Notice `Test` **DOESN'T** slide to left.
1) Notice `Test` is **STILL** displayed.
1) Finally, follow the above same steps on the tags sections and verify correctness.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/widgets/NoteEditorViewPager.java
@@ -37,6 +37,11 @@ public class NoteEditorViewPager extends ViewPager {
         return this.mIsEnabled && super.performClick();
     }
 
+    @Override
+    public boolean arrowScroll(int direction) {
+        return this.mIsEnabled && super.arrowScroll(direction);
+    }
+
     public void setPagingEnabled(boolean enabled) {
         this.mIsEnabled = enabled;
     }


Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -440,6 +443,20 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
+                    EditText editText = (EditText) view;
+                    if (editText.getSelectionStart() == editText.getText().length()) {
+                        if (keyCode == KeyEvent.KEYCODE_DPAD_RIGHT) {
+                            return true;
+                        }
+                    }
+                }
+                return false;
+            }
+        });
```
theck13:
```suggestion
        mTagInput.setOnKeyListener(new OnKeyListener() {
            @Override
            public boolean onKey(View view, int keyCode, KeyEvent event) {
                return !mIsMarkdownEnabled && mTagInput.getSelectionStart() == mTagInput.getText().length() && keyCode == KeyEvent.KEYCODE_DPAD_RIGHT;
            }
        });
```

Can we use the `mTagInput` variable instead of creating the `editText` variable from the `view` itself?  If we remove the `editText` variable, we can simplify this cascade of `if` statements into a single `if` statement like this.

```java
if (!mIsMarkdownEnabled && mTagInput.getSelectionStart() == mTagInput.getText().length() && keyCode == KeyEvent.KEYCODE_DPAD_RIGHT) {
    return true;
}
```

Since that `if` statement is just returning `true` if true and `false` otherwise, we can simplify the entire block to something like this.

```java
mTagInput.setOnKeyListener(new OnKeyListener() {
    @Override
    public boolean onKey(View view, int keyCode, KeyEvent event) {
        return !mIsMarkdownEnabled && mTagInput.getSelectionStart() == mTagInput.getText().length() && keyCode == KeyEvent.KEYCODE_DPAD_RIGHT;
    }
});
```

theck13 (commented):
I left on comment about a possible simplification to the `mTagInput.setOnKeyListener` code block.  After leaving that comment and reading through the rest of the changes, I had an idea for an alternate solution.  Rather than adding the `isMarkdownEnabled` variable to `SimplenoteEditText` and setting that value for all changes to the note in `NoteEditorFragment`, why don't we set an `OnKeyListener` for `mContentEditText` and override `onKey` like we did for `mTagInput`?  That way we have all the key disabling logic in `NoteEditorFragment` and we don't have to keep track of the markdown value across multiple classes.  Since it would be the same logic for both views, we could create a method and keep the logic in a single place.  Also, I would recommend putting some details about the logic to disable the arrow key in a comment for future reference since it's not immediately clear when and why we want to disable that.  See the method below for example.

```java
/**
 * Disable the arrow key when markdown is not enabled and the cursor is at the end of the note.
 * This will avoid going from the Edit tab to the Preview tab unexpectedly since the ViewPager
 * will page to the NoteMarkdownFragment even is the Edit and Preview tabs are not visible
 * (i.e. markdown is not enabled).
 */
private boolean shouldDisableArrowKey(EditText editText, int keyCode) {
    return !mIsMarkdownEnabled && editText.getSelectionStart() == editText.getText().length() && keyCode == KeyEvent.KEYCODE_DPAD_RIGHT;
}
```

If we did that, we could make the key listener code blocks something like this.

```java
mContentEditText.setOnKeyListener(new OnKeyListener() {
    @Override
    public boolean onKey(View view, int keyCode, KeyEvent event) {
        return shouldDisableArrowKey(mContentEditText, keyCode);
    }
});
```

```java
mTagInput.setOnKeyListener(new OnKeyListener() {
    @Override
    public boolean onKey(View view, int keyCode, KeyEvent event) {
        return shouldDisableArrowKey(mTagInput, keyCode);
    }
});
```

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
ParaskP7: Great suggestion @theck13, many thanks, indeed the code gets much more simlified! 🌟 🤗 

This is now done, you can check the change here: 8714b3881874cfb9f5f6735eb89b5f8b84c7df3f

ParaskP7:
Once more, great suggestions @theck13, the end result is much cleaner! 🌟 🤗 

> Rather than adding the isMarkdownEnabled variable to SimplenoteEditText and setting that value for all changes to the note in NoteEditorFragment, why don't we set an OnKeyListener for mContentEditText and override onKey like we did for mTagInput? That way we have all the key disabling logic in NoteEditorFragment and we don't have to keep track of the markdown value across multiple classes.

This is now done, you can check the change here: e6c6f443af8fc7dc02651b879a391d4e875f172f

> Since it would be the same logic for both views, we could create a method and keep the logic in a single place. Also, I would recommend putting some details about the logic to disable the arrow key in a comment for future reference since it's not immediately clear when and why we want to disable that.

This is now done, you can check the change here: 95cde4330828d90ae32424d6e035dfd833528776

🙏

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -547,6 +562,17 @@ public void onGlobalLayout() {
         return mRootView;
     }
 
+    /**
+     * Disable the right arrow key when markdown is disabled and the cursor is at the end of the note.
+     * This will avoid going from the edit tab to the preview tab unexpectedly since the view pager
+     * will page to the {@link NoteMarkdownFragment} even is the edit and preview tabs are not visible.
```
theck13:
```suggestion
     * will page to the {@link NoteMarkdownFragment} even if the edit and preview tabs are not visible.
```

I think the "is" should be "if" here.  I just noticed this typographical error in my comment suggestion.  Sorry about that.

theck13 (commented):
> I considered RTL support as well, but giving the fact that the text editor is not to a 100% RTL ready, adding the fix for RTL will break current functionality.

What do you mean by this?  It seems like Simplenote Android supports left-to-right and right-to-left languages with both the editor content and device setting.

#### Content
The app shows left-to-right, right-to-left, and multilingual text as expected on the ***Edit*** and ***Preview*** tabs regardless of the device's language setting.  See the screenshots below for illustration.

1229_right_arrow_blank_content

#### Setting
When a device is set to a left-to-right language like English, the interface is left-to-right.  When a device is set to a right-to-left language like Hebrew, the interface is right-to-left.  See the screenshots below for illustration.

1229_right_arrow_blank_setting

I added the `[status] needs design review` label so that @SylvesterWilmott can take a look at these changes and ensure the user interaction is as expected.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
ParaskP7: Thanks for noticing that! 🌟 

This is now done, you can check the change here: f106a2f722d6e678138e417527f3a8ffd54114d2

ParaskP7:
> What do you mean by this? It seems like Simplenote Android supports left-to-right and right-to-left languages with both the editor content and device setting.

@theck13 you are right, I was thinking about it wrong. Thanks for pointing that out! 🌟 

I was changing the device setting to RTL and not the keyboard language (like Hebrew). I am actually not the best RTL person so I totally missed that. 😊 

After testing with an RTL keyboard language, I can verify that the solution is incomplete and needs to be updated in order to support an RTL keyboard language as well. So, what needs to be done is:
- For LTR keyboard languages, disable `right arrow` keyboard key, when the cursor is at the end. ✅ 
- For RTL keyboard languages, disable `left arrow` keyboard key, when the cursor is at the end. ⚙️ 

I am currently working on it.

ParaskP7:
> For RTL keyboard languages, disable left arrow keyboard key, when the cursor is at the end. ⚙️

@theck13 , this is now done, you can check the change here: f5bf3e1d7420bb41e734913081e065561db69a27 and 860721c4a41578c440c23de1dd4ddea616d09f4e

However, and as discussed and agreed upon (in Slack), please note that this overall PR solution works explicitly when a device language configuration matches the selected keyboard language. Both should be LTR or RTL for it to work. If however, the device language configuration is on LTR and the keyboard language on RTL, or vice-versa, then the user would be unable to move when the edit text selection is at the very end, if they try to go back a character, that is, by using the respective directional arrow keyboard key.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -547,6 +562,17 @@ public void onGlobalLayout() {
         return mRootView;
     }
 
+    /**
+     * Disable the right arrow key when markdown is disabled and the cursor is at the end of the note.
```
theck13:
```suggestion
     * Disable the arrow key when markdown is disabled and the cursor is at the end of the note.
```

Now that we are disabling the left or right arrow key depending on the device configuration, let's remove "right" from the comment to avoid confusion.

theck13 (commented):
I left one comment about the `shouldDisableRightArrowKey` method comment.  Everything else looks good to me.

> However, and as discussed and agreed upon (in Slack), please note that this overall PR solution works explicitly when a device language configuration matches the selected keyboard language. Both should be LTR or RTL for it to work. If however, the device language configuration is on LTR and the keyboard language on RTL, or vice-versa, then the user would be unable to move when the edit text selection is at the very end, if they try to go back a character, that is, by using the respective directional arrow keyboard key.

I think that's ideal actually.  The interface supports left-to-right and right-to-left languages based on the device's selected language and the ***Edit*** and ***Preview*** tabs switch positions based on that.  Those tabs are the root cause of the right or left arrows paging inadvertently.  If the user changes their keyboard language without changing their device language, the interface and tabs do not change.  Therefore, the disabled arrow key should not change either.  That allows the user to switch keyboard language, enter text in left-to-right and/or right-to-left languages in the same note, and the arrow key will work as expected.  In other words, I think these changes are the correct solution.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
ParaskP7: @theck13 you are right, I was thinking about this having LTR being the default, but it is better to be more inclusive and as such more descriptive. I updated the method name and Javadoc with `Directional` instead. Wdyt?

This is now done, you can check the change here: 52f71c7f12f3db99342b2d967b382654a4bfc3d9

ParaskP7:
> I think that's ideal actually. The interface supports left-to-right and right-to-left languages based on the device's selected language and the Edit and Preview tabs switch positions based on that. Those tabs are the root cause of the right or left arrows paging inadvertently. If the user changes their keyboard language without changing their device language, the interface and tabs do not change. Therefore, the disabled arrow key should not change either. That allows the user to switch keyboard language, enter text in left-to-right and/or right-to-left languages in the same note, and the arrow key will work as expected. In other words, I think these changes are the correct solution.

I am still a bit skeptical about this being the ideal, since this, in specific situations, will wrongly disable a specific directional arrow key, and the user will no longer be able to move left or right with the directional arrows. For example, consider a user that have LTR device configuration, but writes notes in Hebrew. Or a user that have RTL device configuration, but writes notes in English. Both of them will have this specific issue. I understand that this sub-case of users might be just a handful, since they will have to meet all those criteria, using a physical keyboard and all, but it is still a valid use-case. We can merge this solution to `develop`, but be aware of this limitation and if we get reports from users on having such problems, then revisit it.

SylvesterWilmott:
@theck13 was able to show me the behaviour over on slack and it looks sound to me!

Thanks @ParaskP7

theck13:
After doing more testing, I found some cases that are buggy and we may have to think of another solution rather than disabling arrow keys based on device language.

#### Device English / Note Hebrew
1. Set device language to English.
2. Tap note in list with only Hebrew text.
3. Tap end of note content.
4. Tap right arrow.
5. Notice cursor does not move.

#### Device Hebrew / Note English
1. Set device language to Hebrew.
2. Tap note in list with only English text.
3. Tap end of note content.
4. Tap left arrow.
5. Notice cursor does not move.
6. Tap right arrow.
7. Notice blank screen is shown.

I think I found a much simpler and stable solution.  I can't believe I didn't find this earlier or include it when I created the class.  [`ViewPager` has the `arrowScroll` method](https://developer.android.com/reference/kotlin/androidx/viewpager/widget/ViewPager#arrowScroll(kotlin.Int)), which handles scrolling with left or right arrows.  I think that's exactly what we want to override.  Since the `NoteEditorViewPager` class already has the `mIsEnabled` variable, which is being set to `true`/`false` when markdown is enabled/disabled, we can simply add this to the `NoteEditorViewPager` class.

```java
@Override
public boolean arrowScroll(int direction) {
    return this.mIsEnabled && super.arrowScroll(direction);
}
```

It seems to work as expected with my testing.  Please, verify that with your own testing as well.

ParaskP7:
👋 @theck13 !

> After doing more testing, I found some cases that are buggy and we may have to think of another solution rather than disabling arrow keys based on device language.

This is exactly what I was trying to describe above, thank you so much for putting that into clear test instructions and actually testing it itself. I should have done that myself, mental note to my future self for next time. 📜

> I think I found a much simpler and stable solution. I can't believe I didn't find this earlier or include it when I created the class. ViewPager has the arrowScroll method, which handles scrolling with left or right arrows. I think that's exactly what we want to override. Since the NoteEditorViewPager class already has the mIsEnabled variable, which is being set to true/false when markdown is enabled/disabled, we can simply add this to the NoteEditorViewPager class.

Wow, you are totally right, this is an amazing finding Tyler! 🌟

> It seems to work as expected with my testing. Please, verify that with your own testing as well.

I just tested it and it works as expected, this is great! I am now going to drop all the previous commits and just add this as a single-commit solution. 🙏

ParaskP7:
> I am now going to drop all the previous commits and just add this as a single-commit solution.

This is now done, please review again! 🙏

theck13 (approved):

PR #1234, opened 2020-12-14
Update Simperium References

theck13:
### Fix
Update references from Simperium to Simplenote dimension resources.  The `fragment_about` layout file was referencing the `width_layout` dimension resource, which comes from the Simperium library.  This pull request changes that reference to `style_width` and renames `style_width` to `content_width` since the dimension is no longer used for the **_Style_** screen only.  These changes update the layout width on large devices since the Simperium dimension resource `width_layout` has different values for `values-large` and `values-large-land`.  That is expected since we have updated the content width to mimic the **_Style_** screen.

### Test
1. Open navigation drawer.
2. Tap ***Settings*** item in navigation drawer.
3. Tap ***About*** item under ***More*** section.
4. Notice ***About*** layout is full-width on phones and not full-width on tablets.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/res/layout/fragment_about.xml
+++ b/Simplenote/src/main/res/layout/fragment_about.xml
@@ -12,7 +12,7 @@
     <LinearLayout
         android:layout_gravity="center_horizontal"
         android:layout_height="wrap_content"
-        android:layout_width="@dimen/width_layout"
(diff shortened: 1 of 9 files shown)


AmandaRiu (approved):
Tested on a Pixel3 and a Pixel C. All looks good! 👍 :shipit:

PR #1237, opened 2020-12-15
Fix the issue of adding checkboxes to long lines

### LINKED ISSUE #1221: Can't add checkboxes to long lines
If a line is so long that it wraps, the "checkbox" button fails to work properly.

### Expected
The checkbox button should toggle a checkbox at the beginning of a line, whether or not the line is wrapped.

### Observed
The first tap of the checkbox button adds a Markdown checkbox instead of a graphical checkbox in the middle of the line. Tapping it again adds further Markdown checkboxes.

### Reproduced
1. Type a line so long that it wraps
2. Tap the checkbox button

Screen Shot 2020-12-01 at 6 31 58 PM

Device: Google Pixel 3, Android 10.0.0 (Android Emulator)

hichamboushaba:
Fixes #1221 

### Fix
Currently when trying to add checkboxes for long lines (lines that wrapped) it gets added at the wrong position, the cause is the start and end of line used to create the span use the `layout` methods, which are different from the content positions if the line is wrapped.
I added two utility methods to calculate the start of end of line(s) of selection.

#### 1. Confirm the issue is solved

1. Type a line so long that it wraps
2. Put the cursor inside the line, or select a portion
2. Tap the checkbox button
3. Confirm that a checkbox is added at the beginning of the line

#### 2. Non regression
Smoke test the checkbox feature, and confirm that it works as expected.

### Review

### Release
RELEASE-NOTES.txt` was updated in 4eda620591492b26121ec90077004a085200a736 with:
> Fixed a bug with adding checkboxes to long lines

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,6 @@
 2.15
 -----
 * Added permanently deleting notes from trash individually [https://github.com/Automattic/simplenote-android/pull/1224]
-* Fixed a bug with adding checkboxes to long lines [https://github.com/Automattic/simplenote-android/pull/1237]
  
 2.14
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 2.15
 -----
 * Added permanently deleting notes from trash individually [https://github.com/Automattic/simplenote-android/pull/1224]
+* Fixed a bug with adding checkboxes to long lines [https://github.com/Automattic/simplenote-android/pull/1237]
  
 2.14
 -----
--- a/Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
@@ -191,13 +191,11 @@ public class SimplenoteEditText extends AppCompatMultiAutoCompleteTextView imple
         }
     }
 
-    /**
-     *  Finds the start of line containing the start of selection
-     */
     private int findStartOfLineOfSelection() {
         int position = getSelectionStart();
-        if(position == -1) {
-            position = 0;
+        // getSelectionStart may return -1 if there is no selection nor cursor
+        if (position == -1) {
+            return 0;
         }
         Editable editable = getText();
         for (int i = position - 1; i >= 0; i--) {
@@ -208,16 +206,12 @@ public class SimplenoteEditText extends AppCompatMultiAutoCompleteTextView imple
         return 0;
     }
 
-    /**
-     *  Finds the end of line containing the end of selection
-     */
     private int findEndOfLineOfSelection() {
-        int position = getSelectionEnd();
-        if(position == -1) {
-            position = 0;
-        }
+        // getSelectionEnd may return -1 if there is no selection nor cursor
+        // and as the minimum position is 0, use the max value between 0 and getSelectionEnd()
+        int position = Math.max(0, getSelectionEnd());
         Editable editable = getText();
-        for (int i = position; i < editable.length() ; i++) {
+        for (int i = position; i < editable.length(); i++) {
             if (editable.charAt(i) == '\n') {
                 // We return the max here, because when the cursor is at an empty line,
                 // i-1 would point to the start of line
--- a/Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
@@ -191,28 +191,45 @@ public class SimplenoteEditText extends AppCompatMultiAutoCompleteTextView imple
         }
     }
 
-    // Returns the line position of the text cursor
-    private int getCurrentCursorLine() {
-        int selectionStart = getSelectionStart();
-        Layout layout = getLayout();
-
-        if (!(selectionStart == -1)) {
-            return layout.getLineForOffset(selectionStart);
+    /**
+     *  Finds the start of line containing the start of selection
+     */
+    private int findStartOfLineOfSelection() {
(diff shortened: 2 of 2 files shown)


RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 2.15
 -----
 * Added permanently deleting notes from trash individually [https://github.com/Automattic/simplenote-android/pull/1224]
+* Fixed a bug with adding checkboxes to long lines [https://github.com/Automattic/simplenote-android/pull/1237]
```
theck13:
```suggestion
```

I don't think we need a release note entry for this change.  It isn't a crash and we haven't received reports from users about it even though it's been there since checkboxes were added years ago.

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
```diff
@@ -191,28 +191,45 @@ public void run() {
...
-    // Returns the line position of the text cursor
-    private int getCurrentCursorLine() {
-        int selectionStart = getSelectionStart();
-        Layout layout = getLayout();
-
-        if (!(selectionStart == -1)) {
-            return layout.getLineForOffset(selectionStart);
+    /**
+     *  Finds the start of line containing the start of selection
+     */
```
theck13:
```suggestion
```

I don't think this comment adds more information about the method.  The name of the method is a slightly shorter version of the comment.  So, we can remove it.

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
```diff
@@ -191,28 +191,45 @@ public void run() {
...
-        int lineNumber = getCurrentCursorLine();
-        start = getLayout().getLineStart(lineNumber);
-
-        if (getSelectionEnd() > getSelectionStart() && !selectionIsOnSameLine()) {
-            end = getSelectionEnd();
-        } else {
-            end = getLayout().getLineEnd(lineNumber);
+    /**
+     *  Finds the end of line containing the end of selection
+     */
```
theck13:
```suggestion
```

Similar to the previous suggestion, the comment doesn't seem to add more information about the method since the name of the method is a slightly shorter version of the comment.  So, we can remove it.

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
```diff
@@ -191,28 +191,45 @@ public void run() {
...
-        if (!(selectionStart == -1)) {
-            return layout.getLineForOffset(selectionStart);
+    /**
+     *  Finds the start of line containing the start of selection
+     */
+    private int findStartOfLineOfSelection() {
+        int position = getSelectionStart();
+        if(position == -1) {
+            position = 0;
+        }
```
theck13:
```suggestion
         // Since the minimum position index is 0 and the selection start can return -1 for no
        // selection, get the maximum value between 0 and getSelectionStart().
        int position = Math.max(0, getSelectionStart());
```

We can use `Math.max` as a trick similar to the one in `findEndOfLineOfSelection()`.  It may not be obvious why we're checking for `-1` if someone isn't familiar with `getSelectionStart()`.  So, we may want to add a comment explaining what's happening.

It's debatable whether using `Math.max` or `if (position == -1)` is better.  Changing this to `Math.max` is not necessary.  It's just a suggestion.  I wanted to point out a possible alternative in case you want to use it.  Either way, we should probably add some comments so that the logic is clear.

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
```diff
@@ -191,28 +191,45 @@ public void run() {
...
-        } else {
-            end = getLayout().getLineEnd(lineNumber);
+    /**
+     *  Finds the end of line containing the end of selection
+     */
+    private int findEndOfLineOfSelection() {
+        int position = getSelectionEnd();
+        if(position == -1) {
+            position = 0;
+        }
```
theck13:
```suggestion
        // Since the minimum position index is 0 and the selection end can return -1 for no
        // selection, get the maximum value between 0 and getSelectionEnd().
        int position = Math.max(0, getSelectionEnd());
```

We can use `Math.max` as a trick similar to the `return` statement below.  It may not be obvious why we're checking for `-1` if someone isn't familiar with `getSelectionEnd()`.  So, we may want to add a comment explaining what's happening.

It's debatable whether using `Math.max` or `if (position == -1)` is better.  Changing this to `Math.max` is not necessary.  It's just a suggestion.  I wanted to point out a possible alternative in case you want to use it.  Either way, we should probably add some comments so that the logic is clear.

theck13 (commented):
The fix works as described.  Great work diagnosing the issue!

Even though we are looping through the length of the note looking for the start and end of the line, I didn't see any new performance issues on a note with 77,000 lines and 516,000 characters on a single line.

I left some suggestions on the code.  No changes are required to merge this though.  Let me know if you would like to make any changes or leave the code as is.

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
hichamboushaba: I adjusted the logic to return 0 directly without looping `if (position == -1)`, IMO it's better, and follows the return early pattern.

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
hichamboushaba: Done 8ece2a016764255da072bfa2a10d088c478be3cd

theck13 (approved):

PR #1240, opened 2020-12-16
Fix search of terms containing dash character

### LINKED ISSUE #1190: Search for items containing dash does not return result.
### Expected
If I have a note that includes a dash, for example 16-3, I should be able to search for this and see the results returned. 

### Observed
When searching 16-3, the search does not return any result even though it is included in one of my notes. 

### Reproduced

1. Go to a new or existing note and add the text `16-3`
2. Tap on search and search for `16-3`
3. See no results are returned. 

When searching for this in the web-app, electron app, iOS or macOS app, the search returns the result.

iOS|Android
-|-
<kbd><a href="https://user-images.githubusercontent.com/10325627/97867293-7d59a100-1d05-11eb-8855-2323e0167ff6.jpeg"></a></kbd>|<kbd><a href="https://user-images.githubusercontent.com/10325627/97867369-a0845080-1d05-11eb-8894-6529dc4da247.jpg"></a></kbd>

User report in 3434240-zen 

Device Make: Samsung
Device Model: S9+
Android Version: 10.0.0
Simplenote Version: 2.11.1

hichamboushaba:
Fixes #1190 

### Fix
Searching for terms that include the dash character are not working currently, while they work on other platforms.
The cause is that Simperium uses FTS3 for the full text search, and the dash character is a special character that translates to NOT:
> The NOT operator (or, if using the standard syntax, a unary "-" operator)
https://www.sqlite.org/fts3.html#section_3_2

To fix this, the term including the dash character has to be converted to a Phrase query, which means enclosed in double quotation.
The fix impacts only the situation where the dash is between two words, but we are keeping the logic to exclude terms preceded by a dash.

#### Issue fix:

1. Go to a new or existing note and add the text 16-3
2. Tap on search and search for 16-3
3. Make sure that the note is part of the search results.

#### Exclude the words preceded by a dash

1. Go to a new or existing note and add the text 16-3
2. Tap on search and search for 16 -3
3. Make sure that the note is not part of the search results.

#### Non regression:
Smoke test the search feature

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
@@ -97,7 +97,10 @@ public class SearchTokenizer {
                 continue;
             }
 
-            if (current == HYPHEN && (!inStrictTerm || quoteChar == SINGLE_QUOTE)) {
+            if (current == HYPHEN
+                    && last != SPACE
+                    && position != 0
+                    && (!inStrictTerm || quoteChar == SINGLE_QUOTE)) {
                 String tempQuery = query.toString();
                 int tokenStartIndex = tempQuery.lastIndexOf(SPACE, tempQuery.indexOf(last));
                 query.insert(Math.max(0, tokenStartIndex + 1), DOUBLE_QUOTE);
@@ -123,6 +126,7 @@ public class SearchTokenizer {
         // close the strict term
         if (inStrictTerm) query.append(quoteChar);
         if (inTerm && !inStrictTerm && !isLiteral) query.append(GLOB);
+        if (hasHyphen) query.append(DOUBLE_QUOTE);
 
         return query.toString();
     }
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
@@ -11,6 +11,7 @@ public class SearchTokenizer {
     static public final char GLOB = '*';
     static public final char COLON = ':';
     static public final char ESCAPE = '\\';
+    static public final char HYPHEN = '-';
 
     private String mRawQuery;
 
@@ -42,6 +43,8 @@ public class SearchTokenizer {
 
         // if the previous char was an escape char
         boolean isEscaped = false;
+
+        boolean hasHyphen = false;
 
         // the current character
         char current = '\0', last, quoteChar = '\0';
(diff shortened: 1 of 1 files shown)


Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
```diff
@@ -94,10 +97,26 @@ private String parseQuery() {
                 continue;
             }
 
+            if (current == HYPHEN
+                    && last != SPACE
+                    && position != 0
+                    && (!inStrictTerm || quoteChar == SINGLE_QUOTE)) {
```
theck13:
```suggestion
            if (current == HYPHEN && last != SPACE && position != 0) {
```

Since there is a `continue` statement above this which occurs when `inStrictTerm` is `true`, `!inStrictTerm` will always be `true` at this point.  Therefore, `!inStrictTerm || quoteChar == SINGLE_QUOTE` will always be `true`.  So, we can remove that condition.

theck13 (commented):
I left one comment about the code.  Other than that, I noticed the things below while testing.  Let me know if they are out of the scope of these changes, they are a limitation of full-text search, or they are something we would want to fix as well.

I created a note with numbers before and after a hyphen plus various spaces around the hyphen.  See the ***Note*** screenshot below for illustration.  It seems like the hyphen character is ignored unless it's preceded by a space.  Searching for `12 -34` worked as expected though.  Also, hyphens don't appear to be included in the search match highlights.  Searching for `12-34`, `12- 34`, and `12 34` returned different notes in the results while the highlight matches are the same.  See the screenshots below for illustration.

1190_search_hyphen_multiple

There doesn't seem to be a way to search for hyphens.  If you search for just a `-`, no results are shown even if notes include hyphens.  See the screenshot below for illustration.

<a href="https://user-images.githubusercontent.com/3827611/102531825-3844c080-4060-11eb-93bc-d29a98817c98.png"></a>

hichamboushaba:
> I created a note with numbers before and after a hyphen plus various spaces around the hyphen. See the Note screenshot below for illustration. It seems like the hyphen character is ignored unless it's preceded by a space. Searching for 12 -34 worked as expected though. Also, hyphens don't appear to be included in the search match highlights. Searching for 12-34, 12- 34, and 12 34 returned different notes in the results while the highlight matches are the same. See the screenshots below for illustration.

That's because according to the Simperium code, the app uses the tokenizer `unicode61` which according to the documentation treats all punctuation characters as token separators: 
> Unicode61 works very much like "simple" except that it does simple unicode case folding according to rules in Unicode Version 6.1 and it recognizes unicode space and punctuation characters and uses those to separate tokens. 
https://www.sqlite.org/fts3.html#tokenizer

Which means that all the above search terms are equivalent when it comes to the highlighted terms, but differs in the results as with the hyphen between the terms, we convert the query to a Phrase query, which requires the terms to be concatenated. 
And this explains too why searching for hyphen only fails.

We can customize the tokenizer by asking it to treat the hyphen as a `tokenchar` (it has to be done on Simperium), which may solve the above issues (I didn't test it, and I don't know how it impacts treating '-' as NOT)
But IMO this should wait until a decision is made about the search parity between platforms, this PR at least solves the issue of finding composite words containing hyphens, which fails currently.

theck13 (approved):

PR #1241, opened 2020-12-17
Add Heartbeat Listener

theck13:
### Fix
Add a heartbeat listener to the Simplenote application class for the Simperium WebSocket.  When a heartbeat is received or stopped, an item is added to the application log.  The frequency of the heartbeat is `WebSocketManager.HEARTBEAT_INTERVAL`, which is ten seconds.  A stopped heartbeat is determined by the heartbeat timeout, which is double `WebSocketManager.HEARTBEAT_INTERVAL` or twenty seconds.

### Test
1. Open app.
2. Wait twenty or thirty seconds.
3. Enable airplane mode.
4. Wait forty to fifty seconds.
5. Open navigation drawer.
6. Tap ***Settings*** item in navigation drawer.
7. Tap ***Send logs*** item under ***More*** section.
8. Choose option in ***Share*** sheet to view application log.
9. Notice ***NETWORK: Heartbeat received*** and ***NETWORK: Heartbeat stopped*** items in application log.

### Review

### Release

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -81,7 +81,7 @@ dependencies {
     // Fastlane screengrab for screenshots automation
     androidTestImplementation 'tools.fastlane:screengrab:2.0.0'
     // Automattic and WordPress dependencies
-    implementation 'com.simperium.android:simperium:0.9.14'
+    implementation 'com.simperium.android:simperium:0.9.16'
     implementation 'com.automattic:tracks:1.2'
     implementation 'org.wordpress:passcodelock:2.0.2'
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -35,24 +35,23 @@ import org.wordpress.passcodelock.AppLockManager;
 import static com.automattic.simplenote.models.Preferences.PREFERENCES_OBJECT_KEY;
 
 public class Simplenote extends Application implements HeartbeatListener {
-    private static final long HEARTBEAT_TIMEOUT =  WebSocketManager.HEARTBEAT_INTERVAL * 2;
-    private static final int TEN_SECONDS_MILLIS = 10000;
-
-    // log tag
-    public static final String TAG = "Simplenote";
-
-    // intent IDs
-    public static final int INTENT_PREFERENCES = 1;
-    public static final int INTENT_EDIT_NOTE = 2;
     public static final String DELETED_NOTE_ID = "deletedNoteId";
     public static final String SELECTED_NOTE_ID = "selectedNoteId";
+    public static final String TAG = "Simplenote";
+    public static final int INTENT_EDIT_NOTE = 2;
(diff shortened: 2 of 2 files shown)


hichamboushaba:
The functionality is working well, I have just two remarks, but they are not mandatory to merge the code:

1. IMO it would be better to create a separate class that implements `HeartbeatListener` instead of having the application class do it, as it's not its role, and would make the code easier to read by finding the responsible class and what's is doing.
2. This is not related to the current PR, but just wanted to suggest it: To make testing and debugging the app easier, why not log the messages sent to AppLog in Logcat for debug builds? something like this:
```kotlin
    public static void add(Type type, String message) {
        String log;

        if (type == Type.ACCOUNT || type == Type.DEVICE) {
            log = message;
        } else {
            String timestamp = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS", Locale.US).format(new Date());
            log = timestamp + " - " + type.toString() + ": " + message + "\n";
        }

        if(BuildConfig.DEBUG) Log.d(type.toString(), message);

        mQueue.put(mQueue.size(), log);
    }
```

hichamboushaba (commented):
The functionality is working well, I have just two remarks, but they are not mandatory to merge the code:

1. IMO it would be better to create a separate class that implements `HeartbeatListener` instead of having the application class do it, as it's not its role, and would make the code easier to read by finding the responsible class and what's is doing.
2. This is not related to the current PR, but just wanted to suggest it: To make testing and debugging the app easier, why not log the messages sent to AppLog in Logcat for debug builds? something like this:
```kotlin
    public static void add(Type type, String message) {
        String log;

        if (type == Type.ACCOUNT || type == Type.DEVICE) {
            log = message;
        } else {
            String timestamp = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS", Locale.US).format(new Date());
            log = timestamp + " - " + type.toString() + ": " + message + "\n";
        }

        if(BuildConfig.DEBUG) Log.d(type.toString(), message);

        mQueue.put(mQueue.size(), log);
    }
```

theck13:
<blockquote>
<ol start="1">
<li>IMO it would be better to create a separate class that implements `HeartbeatListener` instead of having the application class do it, as it's not its role, and would make the code easier to read by finding the responsible class and what's is doing.</li>
</ol>
</blockquote>

Thanks for the feedback.  I think it's fine as it is.  It doesn't clutter the `Simplenote` class incredibly.  Since the heartbeat is intended to last the lifetime of the application, it makes sense to have it in the application class.

<blockquote>
<ol start="2">
<li>This is not related to the current PR, but just wanted to suggest it: To make testing and debugging the app easier, why not log the messages sent to AppLog in Logcat for debug builds? something like this:</li>
</ol>
</blockquote>

We can add that to `AppLog` if we want.  Since the application log is meant for troubleshooting customer support issues rather than debugging development issues, logging to the development console wasn't included.  The only time logging to the development console would be used is while reviewing a pull request related to the application log, which is a one-time thing and very infrequent.

hichamboushaba (approved):

PR #1242, opened 2020-12-17
Select next note after trashing/restoring a note in tablet landscape

### LINKED ISSUE #1021: Select next note after trashing/restoring a note in tablet landscape
### What

Two suggested changes to the tablet landscape behavior:

* Select the note below a recently trashed note.
* Select the note below a recently restored note in trash view.

Currently, the Android app shows a split view (note list and note editor) in tablet landscape view. When a note is trashed, or a trashed note is restored, the note editor is blank until another note is manually selected:

android-notes-loading

(As observed on HTC Nexus 9, Android 7.1.1, Simplenote Android 2.6-rc-1.)

The current Electron behavior is to select the next note in the list when a note is trashed/restored, so the note editor is never blank:

electron-notes-loading

(As observed on a Macbook Pro, macOS 10.15.4, Simplenote Electron 1.16.0 in Chrome.)

### Why

This enhancement would bring the Android tablet landscape behavior in line with Electron behavior.

hichamboushaba:
Fixes #1021 

### Fix
This PR adds implementation of the bellow suggestions:
- Select the note below a recently trashed note.
- Select the note below a recently restored note in trash view.

The goal is to offer feature parity with other platforms.

trash_select

#### Selecting the note below the trashed note:
1. Open the notes list using a tablet in landscape mode
2. Open a note that's not the latest in the list.
3. Trash the note.
4. Make sure that the note below is selected.

#### Selecting the note above the trashed note:
1. Open the notes list using a tablet in landscape mode
2. Open a latest note in the list.
3. Trash the note.
4. Make sure that the note above is selected.

#### Trashing the latest note:
1. Open the notes list using a tablet in landscape mode
2. Trash all note except one, and open it.
3. Trash the note.
4. Make sure that the note placeholder is shown.

**Important** Please try to run the tests on a Chromebook too, as I don't have one to test.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -266,12 +266,9 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
     }
 
     public void updateSelectionAfterTrashAction() {
-        updateSelectionAfterTrashAction(getSelectedNotesPositions());
-    }
-
-    private void updateSelectionAfterTrashAction(List<Integer> deletedNotesPositions) {
         if (DisplayUtils.isLargeScreenLandscape(getActivity())) {
             // Try to find the nearest note to the first deleted item
+            List<Integer> deletedNotesPositions = getSelectedNotesPositions();
             int firstDeletedNote = deletedNotesPositions.get(0);
             int positionToSelect = -1;
             // Loop through the notes below
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -205,7 +205,6 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
         inflater.inflate(R.menu.bulk_edit, menu);
         DrawableUtils.tintMenuWithAttribute(getActivity(), menu, R.attr.actionModeTextColor);
         mActionMode = actionMode;
-        requireActivity().getWindow().setStatusBarColor(ThemeUtils.getColorFromAttribute(requireContext(), R.attr.mainBackgroundColor));
         return true;
     }
 
@@ -266,29 +265,55 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
         return positions;
     }
 
-    public Note getItemAtPosition(int position) {
-        return mNotesAdapter.getItem(position + mList.getHeaderViewsCount());
+    public void updateSelectionAfterTrashAction() {
+        updateSelectionAfterTrashAction(getSelectedNotesPositions());
+    }
+
+    private void updateSelectionAfterTrashAction(List<Integer> deletedNotesPositions) {
+        if (DisplayUtils.isLargeScreenLandscape(getActivity())) {
+            // Try to find the nearest note to the first deleted item
+            int firstDeletedNote = deletedNotesPositions.get(0);
+            int positionToSelect = -1;
+            // Loop through the notes below
+            for (int i = firstDeletedNote + 1; i < mNotesAdapter.getCount(); i++) {
+                if (!deletedNotesPositions.contains(i)) {
+                    positionToSelect = i;
+                    break;
+                }
+            }
+            if (positionToSelect == -1) {
+                // Loop through the above notes
+                for (int i = firstDeletedNote - 1; i >= 0; i--) {
+                    if (!deletedNotesPositions.contains(i)) {
+                        positionToSelect = i;
+                        break;
+                    }
+                }
+            }
+
+            if (positionToSelect != -1) {
+                Note noteToSelect = mNotesAdapter.getItem(positionToSelect + mList.getHeaderViewsCount());
+                mCallbacks.onNoteSelected(noteToSelect.getSimperiumKey(), null, noteToSelect.isMarkdownEnabled(), noteToSelect.isPreviewEnabled());
+                // As we will trigger a list refresh later, save the selectedNoteId
+                mSelectedNoteId = noteToSelect.getSimperiumKey();
+            } else {
+                // The list of notes is empty
+                if (getActivity() != null) {
+                    ((NotesActivity) requireActivity()).showDetailPlaceholder();
+                }
+            }
+        }
     }
 
     @Override
     public void onDestroyActionMode(ActionMode mode) {
         mCallbacks.onActionModeDestroyed();
         mActionMode = null;
-        new Handler().postDelayed(
-            new Runnable() {
-                @Override
-                public void run() {
-                    if (getActivity() != null) {
-                        NotesActivity notesActivity = (NotesActivity) getActivity();
-                        setActivateOnItemClick(DisplayUtils.isLargeScreenLandscape(notesActivity));
-                        notesActivity.showDetailPlaceholder();
-                    }
-
-                    requireActivity().getWindow().setStatusBarColor(getResources().getColor(android.R.color.transparent, requireActivity().getTheme()));
-                }
-            },
-            requireContext().getResources().getInteger(android.R.integer.config_mediumAnimTime)
-        );
+        if (getActivity() != null) {
+            NotesActivity notesActivity = (NotesActivity) getActivity();
+            setActivateOnItemClick(DisplayUtils.isLargeScreenLandscape(notesActivity));
+            if(mSelectedNoteId == null) notesActivity.showDetailPlaceholder();
+        }
     }
 
     @Override
@@ -1473,11 +1498,8 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
                         refreshList();
                         return true;
                     case R.id.menu_trash:
-                        note.setDeleted(!note.isDeleted());
-                        note.setModificationDate(Calendar.getInstance());
-                        note.save();
                         if (getActivity() != null) {
-                            ((NotesActivity) getActivity()).updateViewsAfterTrashAction(note);
+                            ((NotesActivity) getActivity()).trashNote(note);
                         }
                         return true;
                     default:
@@ -1678,6 +1700,8 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
                 notesActivity.showUndoBarWithNoteIds(mDeletedNoteIds);
             }
 
+            fragment.updateSelectionAfterTrashAction();
+            fragment.mActionMode.finish();
             fragment.refreshList();
         }
     }
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -205,6 +205,7 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
         inflater.inflate(R.menu.bulk_edit, menu);
         DrawableUtils.tintMenuWithAttribute(getActivity(), menu, R.attr.actionModeTextColor);
         mActionMode = actionMode;
+        requireActivity().getWindow().setStatusBarColor(ThemeUtils.getColorFromAttribute(requireContext(), R.attr.mainBackgroundColor));
         return true;
     }
 
@@ -313,6 +314,15 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
                 notesActivity.showDetailPlaceholder();
             }
         }
+        new Handler().postDelayed(
+            new Runnable() {
+                @Override
+                public void run() {
+                    requireActivity().getWindow().setStatusBarColor(getResources().getColor(android.R.color.transparent, requireActivity().getTheme()));
+                }
+            },
+            requireContext().getResources().getInteger(android.R.integer.config_longAnimTime)
+        );
     }
 
     @Override
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
@@ -309,7 +309,7 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
         if (getActivity() != null) {
             NotesActivity notesActivity = (NotesActivity) getActivity();
             setActivateOnItemClick(DisplayUtils.isLargeScreenLandscape(notesActivity));
-            if (mSelectedNoteId == null){
+            if (mSelectedNoteId == null) {
                 notesActivity.showDetailPlaceholder();
             }
         }
@@ -356,7 +356,7 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
         NotesActivity notesActivity = (NotesActivity) requireActivity();
 
         if (ACTION_NEW_NOTE.equals(notesActivity.getIntent().getAction()) &&
-                !notesActivity.userIsUnauthorized()){
+            !notesActivity.userIsUnauthorized()) {
             //if user tap on "app shortcut", create a new note
             createNewNote("", "new_note_shortcut");
         }
@@ -411,8 +411,8 @@ public class NoteListFragment extends ListFragment implements AdapterView.OnItem
             @Override
             public void onClick(View v) {
                 Context context = ThemeUtils.getStyle(requireContext()) == R.style.Style_Sepia ?
-                        new ContextThemeWrapper(requireContext(), R.style.ToolbarTheme_Popup_Sepia) :
-                        mSortOrder.getContext();
+                    new ContextThemeWrapper(requireContext(), R.style.ToolbarTheme_Popup_Sepia) :
+                    mSortOrder.getContext();
                 PopupMenu popup = new PopupMenu(context, mSortOrder, Gravity.START);
                 MenuInflater inflater = popup.getMenuInflater();
                 inflater.inflate(R.menu.search_sort, popup.getMenu());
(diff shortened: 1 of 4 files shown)


hichamboushaba:
@theck13 I didn't add any release notes for this, please let me know if I should mention it.

theck13 (commented):
I tested these changes on a Nexus 10 emulator and a Pixel Slate device.  Both trashing and restoring a single note select the next or previous note as expected.  Both trashing multiple notes and deleting a single note show the placeholder whether a note exists before the selection or not.  See the animations below for illustration.

#### Trash [Single]
<kbd><a href="https://user-images.githubusercontent.com/3827611/102572201-add47f00-40a8-11eb-9f6e-0ddc4389def8.gif"></a></kbd>

#### Restore [Single]
<kbd><a href="https://user-images.githubusercontent.com/3827611/102572208-af9e4280-40a8-11eb-9e53-a0825663d24d.gif"></a></kbd>

#### Trash [Multiple]
<kbd><a href="https://user-images.githubusercontent.com/3827611/102572215-b2993300-40a8-11eb-880f-9802aa44e431.gif"></a></kbd>

#### Delete [Single]
<kbd><a href="https://user-images.githubusercontent.com/3827611/102572218-b4fb8d00-40a8-11eb-9857-611b33f14578.gif"></a></kbd>

hichamboushaba:
@theck13 thanks for your test, I didn't expect the multiple trash to be in the scope 😄.
I added it:
trash_multiple

And the delete permanently:
delete

Simplenote/src/main/res/values/colors.xml
```diff
@@ -235,4 +235,7 @@
     <color name="yellow_90">#332200</color>
     <color name="yellow_100">#1c1300</color>
 
+    <!-- Update the status guard color to avoid a black status bar on ActionMode -->
+    <color name="abc_input_method_navigation_guard" tools:override="true">@android:color/transparent</color>
```
hichamboushaba: Using the postDelayed to restore the transparent status bar color when action mode was destroyed was causing some flickering after my changes, as I had to run `setActivateOnItemClick` before refreshing the list to avoid resetting the selection.
Overriding this color fixes the issue without having to change the status bar color.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -253,25 +252,65 @@ private String getSelectedNoteLinks() {
...
-
-                    requireActivity().getWindow().setStatusBarColor(getResources().getColor(android.R.color.transparent, requireActivity().getTheme()));
-                }
-            },
-            requireContext().getResources().getInteger(android.R.integer.config_mediumAnimTime)
-        );
+        if (getActivity() != null) {
+            NotesActivity notesActivity = (NotesActivity) getActivity();
+            setActivateOnItemClick(DisplayUtils.isLargeScreenLandscape(notesActivity));
+            if(mSelectedNoteId == null) notesActivity.showDetailPlaceholder();
```
theck13:
```suggestion
            if (mSelectedNoteId == null) {
                notesActivity.showDetailPlaceholder();
            }
```

Some parts of the codebase have inline `if` statements.  So, this may be seen as following the code style for some files or modules.  We're trying to make all `if` statements consistent by using the multiline syntax.  Let's update this to multiline too.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -1061,45 +1058,47 @@ private void updateActionsForLargeLandscape(Menu menu) {
...
         }
 
-        // If we just deleted/restored the active note, show the placeholder
-        if (mCurrentNote != null && mCurrentNote.getSimperiumKey().equals(note.getSimperiumKey())) {
-            showDetailPlaceholder();
-        }
-
-        NoteListFragment fragment = getNoteListFragment();
-        if (fragment != null) {
+        if(getNoteListFragment() != null) {
```
theck13:
```suggestion
        if (getNoteListFragment() != null) {
```

This is missing a space between the `if` and `(`.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -1061,45 +1058,47 @@ private void updateActionsForLargeLandscape(Menu menu) {
...
             fragment.refreshList();
         }
 
         if (mInvalidateOptionsMenuHandler != null) {
             mInvalidateOptionsMenuHandler.removeCallbacks(mInvalidateOptionsMenuRunnable);
             mInvalidateOptionsMenuHandler.postDelayed(
-                mInvalidateOptionsMenuRunnable,
-                getResources().getInteger(android.R.integer.config_shortAnimTime)
+                    mInvalidateOptionsMenuRunnable,
+                    getResources().getInteger(android.R.integer.config_shortAnimTime)
```
theck13:
```suggestion
                mInvalidateOptionsMenuRunnable,
                getResources().getInteger(android.R.integer.config_shortAnimTime)
```

This might have been automatically edited by Android Studio.  The edit and extra indention is not necessary.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -1061,45 +1058,47 @@ private void updateActionsForLargeLandscape(Menu menu) {
...
+                    "overflow_menu"
             );
         } else {
             AnalyticsTracker.track(
-                EDITOR_NOTE_RESTORED,
-                CATEGORY_NOTE,
-                "overflow_menu"
+                    EDITOR_NOTE_RESTORED,
+                    CATEGORY_NOTE,
+                    "overflow_menu"
```
theck13:
```suggestion
                EDITOR_NOTE_RESTORED,
                CATEGORY_NOTE,
                "overflow_menu"
```

This might have been automatically edited by Android Studio.  The edit and extra indention is not necessary.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -1061,45 +1058,47 @@ private void updateActionsForLargeLandscape(Menu menu) {
...
             deletedNoteIds.add(note.getSimperiumKey());
             mUndoBarController.setDeletedNoteIds(deletedNoteIds);
             mUndoBarController.showUndoBar(getUndoView(), getString(R.string.note_deleted));
             AnalyticsTracker.track(
-                LIST_NOTE_DELETED,
-                CATEGORY_NOTE,
-                "overflow_menu"
+                    LIST_NOTE_DELETED,
+                    CATEGORY_NOTE,
+                    "overflow_menu"
```
theck13:
```suggestion
                LIST_NOTE_DELETED,
                CATEGORY_NOTE,
                "overflow_menu"
```

This might have been automatically edited by Android Studio.  The edit and extra indention is not necessary.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
```diff
@@ -253,25 +252,65 @@ private String getSelectedNoteLinks() {
...
+            if (positionToSelect != -1) {
+                Note noteToSelect = mNotesAdapter.getItem(positionToSelect + mList.getHeaderViewsCount());
+                mCallbacks.onNoteSelected(noteToSelect.getSimperiumKey(), null, noteToSelect.isMarkdownEnabled(), noteToSelect.isPreviewEnabled());
+                // As we will trigger a list refresh later, save the selectedNoteId
+                mSelectedNoteId = noteToSelect.getSimperiumKey();
+            } else {
+                // The list of notes is empty
+                if (getActivity() != null) {
+                    ((NotesActivity) requireActivity()).showDetailPlaceholder();
+                }
```
theck13:
```suggestion
                ((NotesActivity) requireActivity()).showDetailPlaceholder();
```

Do we need the `if (getActivity() != null)` check since `requireActivity()` throws an `IllegalStateException` if `getActivity() == null`?  I guess the question is do we want to do nothing or throw an `IllegalStateException` if `getActivity() == null`?  If `getActivity() == null`, that indicates something is wrong with `NoteListFragment`.  If we do nothing, it will leave the last note content in the editor and hide that something is wrong with `NoteListFragment`.  If we throw an `IllegalStateException`, it will crash and show something is wrong with `NoteListFragment`.

Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
```diff
@@ -75,7 +75,12 @@ public void onClick(DialogInterface dialog, int whichButton) {
 
                             // Show empty placeholder for large devices in landscape.
                             if (activity instanceof NotesActivity) {
-                                ((NotesActivity) activity).showDetailPlaceholder();
+                                NotesActivity notesActivity = (NotesActivity) activity;
+                                if(notesActivity.getNoteListFragment() != null) {
```
theck13:
```suggestion
                                if (notesActivity.getNoteListFragment() != null) {
```

This is missing a space between the `if` and `(`.

Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
```diff
@@ -75,7 +75,12 @@ public void onClick(DialogInterface dialog, int whichButton) {
...
+                                    notesActivity.getNoteListFragment().updateSelectionAfterTrashAction();
+                                } else {
+                                    ((NotesActivity) activity).showDetailPlaceholder();
```
theck13:
```suggestion
                                    notesActivity.showDetailPlaceholder();
```

Since the `notesActivity` variable is created above, we can use it in this statement, right?

Simplenote/src/main/res/values/colors.xml
theck13: I'm not a fan of overriding third-party library resources or using reflection since we don't have complete control over those resources.  Can we avoid overriding this resource if we leave the `postDelayed` code in `onDestroyActionMode`, only update the status bar color in that  `Runnable`, move the `setActivateOnItemClick` statement with the associated code outside the `Runnable`, and add the status bar setting statement back to `onCreateActionMode`?  I think we can replace `getActivity()` with `requireActivity()` too.  It would look something like this.

```java
@Override
public boolean onCreateActionMode(ActionMode actionMode, Menu menu) {
    mCallbacks.onActionModeCreated();
    MenuInflater inflater = actionMode.getMenuInflater();
    inflater.inflate(R.menu.bulk_edit, menu);
    DrawableUtils.tintMenuWithAttribute(getActivity(), menu, R.attr.actionModeTextColor);
    mActionMode = actionMode;
    requireActivity().getWindow().setStatusBarColor(ThemeUtils.getColorFromAttribute(requireContext(), R.attr.mainBackgroundColor));
    return true;
}
```

```java
@Override
public void onDestroyActionMode(ActionMode mode) {
    mCallbacks.onActionModeDestroyed();
    mActionMode = null;
    new Handler().postDelayed(
        new Runnable() {
            @Override
            public void run() {
                requireActivity().getWindow().setStatusBarColor(getResources().getColor(android.R.color.transparent, requireActivity().getTheme()));
            }
        },
        requireContext().getResources().getInteger(android.R.integer.config_mediumAnimTime)
    );

    NotesActivity notesActivity = (NotesActivity) requireActivity();
    setActivateOnItemClick(DisplayUtils.isLargeScreenLandscape(notesActivity));

    if (mSelectedNoteId == null) {
        notesActivity.showDetailPlaceholder();
    }
}
```

theck13 (changes requested):

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
hichamboushaba: As this code is called from the `TrashNotesTask`, it's wise to keep the check if the activity is not null, as it may occur after the destruction, and it won't mean that there is something wrong about `NoteListFragment`, it's just that the callback occurred after detachment, which means that we don't need to update the selected note.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
hichamboushaba: If we want to avoid those, we need to clarify that the code style to use (preferably in the contributing guide), and why not add a checkstyle file like FluxC 🙂.
For the current PR, I created a codestyle config from the checkstyle file in our FluxC repo, and removed the all the wrapping options (as they will create a lot changes), then reformatted the files: NotesActivity.java, NoteListFragment.java and NoteUtils.java e1fa30d2d524d614e51951f42fd32721b1dc87eb, please let me know what you think?

Simplenote/src/main/res/values/colors.xml
hichamboushaba: This was the first thing I tried, but as `setActivateOnItemClick` will cause clearing the selection, I noticed that there is a blink of the black status bar.
But now when I think about it, the status guard is added by the compat library only if the status bar is transparent, and we don't need transparent status bars in the app I think, I changed the statusBarColor to match the background color in this commit: 1d264aec5b255deaabf586ea2ef4f82c2b7654f5, which solves the issue without impacting the UI, unless I'm missing something, or an edge case.

hichamboushaba: That was the first thing that I tried, but I noticed some blinking in the status bar.
Personally I prefer the color override, as both solutions are kind of hackish, the fact that we need to post reverting to the transparent color depends on the internals of the library too, and could break with a library update 🙂, I wish they offered a way to disable this behavior for transparent status bar..
Anyway, I was able to solve the blinking issue by using a longer duration for the delay 🙂 : f4ff27ffc4eeb5c1d74f3c7cfc403638a3a75823

Simplenote/src/main/res/values/colors.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
```
theck13:
```suggestion
<resources>
```

The `tools` namespace is no longer user.  We can remove it.

Simplenote/src/main/res/values/colors.xml
```diff
@@ -234,5 +234,4 @@
     <color name="yellow_80">#4f3500</color>
     <color name="yellow_90">#332200</color>
     <color name="yellow_100">#1c1300</color>
-
```
theck13: This line was accidentally deleted with the overridden resource removal.  We can add it back to keep the file as it was before this pull request.

Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
```diff
@@ -75,8 +75,13 @@ public void onClick(DialogInterface dialog, int whichButton) {
...
                             if (activity instanceof NotesActivity) {
-                                ((NotesActivity) activity).showDetailPlaceholder();
-                            // Close editor for small devices and large devices in portrait.
+                                NotesActivity notesActivity = (NotesActivity) activity;
+                                if (notesActivity.getNoteListFragment() != null) {
+                                    notesActivity.getNoteListFragment().updateSelectionAfterTrashAction();
+                                } else {
+                                    notesActivity.showDetailPlaceholder();
+                                }
+                                // Close editor for small devices and large devices in portrait.
```
theck13:
```suggestion
                            // Close editor for small devices and large devices in portrait.
```

This comment was indented with the checkstyle automation.  Let's move it back one indention so that it is the way it was before this pull request and it is clearer the comment is referring to the `else if` below.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -1411,7 +1415,7 @@ public void onConfigurationChanged(@NonNull Configuration newConfig) {
                 }
 
                 invalidateOptionsMenu();
-            // Go to NoteEditorActivity if note editing was fullscreen and orientation was switched to portrait
+                // Go to NoteEditorActivity if note editing was fullscreen and orientation was switched to portrait
```
theck13:
```suggestion
            // Go to NoteEditorActivity if note editing was fullscreen and orientation was switched to portrait
```

This comment was indented with the checkstyle automation. Let's move it back one indention so that it is the way it was before this pull request and it is clearer the comment is referring to the `else if` below.

Simplenote/src/main/java/com/automattic/simplenote/NoteListFragment.java
theck13: The purpose of `requireActivity()` is to throw an `IllegalStateException` if the activity is `null`.  If we're checking `getActivity() != null` just before calling `requireActivity()`, then `requireActivity()` loses its purpose.

If we're worried about `NoteListFragment` being detached when `onPostExecute` is called in `TrashNotesTask`, we can check if the fragment is detached before calling the fragment methods like this:

```java
if (!fragment.isDetached()) {
    fragment.updateSelectionAfterTrashAction();
    fragment.mActionMode.finish();
    fragment.refreshList();
}
```

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
theck13: There is a ton of legacy code with various styles due to different maintainers over the years.  While code style consistency is good in general, there are too many inconsistencies in the codebase to enforce something like checkstyle without having alert fatigue constantly and modifying code that is unrelated to the changes for the bug/enhancement/etc.  Plus, there are always exceptions to the rules.

Simplenote/src/main/java/com/automattic/simplenote/utils/NoteUtils.java
hichamboushaba: Done 6b925da8d3374e1775ea5f803589f51670aa9860, but IMO those comments should go after the curly brace of the `else if` not before, and this would've avoided this indentation issue 🙂

theck13 (approved):

PR #1243, opened 2020-12-17
Update Information Sheet

theck13:
### Fix
Update color, style, and weight of the header and title text as well as the minimum row height in the ***Information*** bottom sheet.  See the screenshots below for illustration.

update_information_sheet

### Test
1. Tap any note in list.
2. Tap ***Information*** action in app bar.
3. Notice dimensions as shown in screenshots above.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/res/layout/bottom_sheet_info.xml
+++ b/Simplenote/src/main/res/layout/bottom_sheet_info.xml
@@ -122,6 +122,7 @@
                 android:layout_height="wrap_content"
                 android:layout_marginTop="@dimen/padding_small"
                 android:layout_width="match_parent"
+                android:minHeight="@dimen/bottom_sheet_row_height"
                 android:padding="@dimen/padding_large"
                 android:text="@string/references"
                 android:textStyle="bold"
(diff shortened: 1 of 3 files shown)


hichamboushaba (approved):
LGTM, I'll wait to merge once design review has been completed.

SylvesterWilmott:
Looks good!

What is the dp height of the reference rows? I am measuring 58dp but it's hard to tell. Can we aim for 64dp total height since that is the recommended height for the condensed list rows [here](https://material.io/components/lists#specs).

SylvesterWilmott:
Maybe this is for another PR but looking at this again, it might make sense to hide the reference count unless there are 2 or more references of that note.

theck13:
> What is the dp height of the reference rows? I am measuring 58dp but it's hard to tell. Can we aim for 64dp total height since that is the recommended height for the condensed list rows [here](https://material.io/components/lists#specs).

I tried making the minimum height 64dp, but it didn't change anything.  According to the dimension values, the height is only 58dp (12dp + 12dp + 16sp + 14sp + 2dp + 2dp), but the height doesn't change unless I use a minimum height of 66dp or more.

> Maybe this is for another PR but looking at this again, it might make sense to hide the reference count unless there are 2 or more references of that note.

I would keep it as is just for consistency and ease of scanning visually.  It might not be obvious why there is a reference number for some items and not others.

theck13:
I removed the line spacing extra (2dp) and updated the top and bottom padding from 12dp to 16dp.  Is this closer to what we want?

update_information_sheet_row

SylvesterWilmott (approved):
The measurements on the screenshot look good, however on the device the measurements are off. It looks like the information rows didn't change on the actual device. I'm not too concerned about it though, the main goal of the PR was met. Thanks!

PR #1248, opened 2020-12-22
Attempt to fix the duplication of shared notes

### LINKED ISSUE #917: Sharing Broken to Simplenote from Google now + Firefox
### Expected
When I share content with the app I expect a note to be created with that content. Just one note should be made, with the title of the site as the title of the note and the URL as the first piece of content.

### Observed
When an article from Google Now is shared to Simplenote multiple copies of a note with that URL are created. When a link is shared from Firefox no note is created. The app simply opens.

The sharing seems to work as expected from other apps tested like Chrome, Airbnb, Google Maps, Pinterest, or the WordPress app reader.

### Reproduced
1. Open up Google now's recommended articles. I do this on my Pixel by swiping from left to right on the homescreen
2. Select an article
3. Select the share icon
4. Select Simplenote
5. Exit the note just created in Simplenote. You'll see multiple copies of that same note

1. Open a URL in Firefox
2. Select the share icon
3. Select the Simplenote icon
4. Simplenote will open but no note will be made

Android 10|Simplenote 2.2
Android 10|Simplenote 2.3

Ticket: #2585883-zen

hichamboushaba:
### Fix
This PR attempts to fix #917, from the video in the issue, it seems that Android is recreating the NotesActivity after going back, and by re-reading the intent we add it another time.
I wasn't able to replicate the same behavior, but by enabling `don't keep activities` option in my device, and sharing a note from the browser, I ended up in a loop where every time I leave the note editor, the note gets duplicated.

@ParaskP7 identified another bug in the same issue, in some cases (the identified one in the issue is quick share from Chrome), the browser attempts to re-use the same activity instance, instead of creating a new one, and then the shared note will be delivered via the callback `onNewIntent`, which wasn't handled.

So the changes here are:
1. Don't add the note if the activity is being recreated by the platform 4c5611c
2. Handle the `onNewIntent` callback f613d80

#### Activity recreation
1. Enable "Don't keep activities" in the developer options.
2. Open the browser, then share a website.
3. Pick Simplenote.
4. The note will be added.
5. Go back.
6. Confirm that the note won't get duplicated.

#### Quick share from Chrome
1. Make sure that "don't keep activities" is disabled, and that Simplenote is closed.
2. Open chrome, then share a website.
3. Pick Simplenote.
4. Click on the home button.
5. Open Chrome again.
6. Share another website but using the quick share button.
7. Confirm that the note was added.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -237,11 +237,17 @@ public class NotesActivity extends ThemedAppCompatActivity implements NoteListFr
         checkForFirstLaunch();
 
         if (savedInstanceState == null) {
-            checkForSharedContent();
+            checkForSharedContent(getIntent());
         }
 
         currentApp.getSimperium().setOnUserCreatedListener(this);
         currentApp.getSimperium().setUserStatusChangeListener(this);
+    }
+
+    @Override
+    protected void onNewIntent(Intent intent) {
+        super.onNewIntent(intent);
+        checkForSharedContent(intent);
     }
 
     @Override
@@ -555,10 +561,9 @@ public class NotesActivity extends ThemedAppCompatActivity implements NoteListFr
         }
     }
 
-    private void checkForSharedContent() {
-        if (getIntent().hasExtra(Intent.EXTRA_TEXT)) {
+    private void checkForSharedContent(Intent intent) {
+        if (intent.hasExtra(Intent.EXTRA_TEXT)) {
             // Check share action
-            Intent intent = getIntent();
             String subject = intent.getStringExtra(Intent.EXTRA_SUBJECT);
             String text = intent.getStringExtra(Intent.EXTRA_TEXT);
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -236,7 +236,9 @@ public class NotesActivity extends ThemedAppCompatActivity implements NoteListFr
         // Creates 'Welcome' note
         checkForFirstLaunch();
 
-        checkForSharedContent();
+        if (savedInstanceState == null) {
+            checkForSharedContent();
+        }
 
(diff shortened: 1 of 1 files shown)


theck13 (commented):
The code looks good to me.  I tested with Chrome 87.0.4280.101, DuckDuckGo 5.71.0, Firefox 84.1.1, and Opera 61.2.3076.56749.  None of those apps created duplicate notes.  Firefox did create a new instance of Simplenote in the overview/recents list every time I shared a note from that browser.  Opera did not create a separate instance of Simplenote in the overview/recents list, which was a little weird since it seemed like Simplenote was part of Opera rather a separate app.  See the animation and screenshot below for illustration.

<kbd><a href="https://user-images.githubusercontent.com/3827611/102969240-97686200-44b2-11eb-81e1-ddca1542aa22.gif"></a></kbd>

<kbd><a href="https://user-images.githubusercontent.com/3827611/102967945-39d31600-44b0-11eb-971c-51d4f17a7143.png"></a></kbd>

I don't think we need to address those odd things with Firefox and Opera necessarily.  I just wanted to mention them in case you knew why those happen or if you wanted to fix them in this pull request.

hichamboushaba:
I did a small re-search to understand the behavior, and it seems Opera's behavior is the default one by the system, as we don't specify any custom launchMode for our activity, it will be launched in the same task of the calling activity.
Firefox (and Chrome too I think) specify some flags to enforce the activity to be launched in a separate task, Firefox uses [FLAG_ACTIVITY_NEW_DOCUMENT](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_NEW_DOCUMENT) combined with [FLAG_ACTIVITY_MULTIPLE_TASK](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_MULTIPLE_TASK) to start a new task, but clear it from recents when it's finished: https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/share/ShareController.kt#L101

I don't think we need to do anything about it, as it doesn't impact the user experience much, and it's the same behavior for other apps too (I tested sharing from Opera to WhatsApp and Twitter, and got the same experience).

theck13 (approved):

PR #1227, opened 2020-12-04
Feature/add app styles

theck13:
### Fix
Merge the `feature/add-app-styles` branch into `develop`.   This pull request is just to merge the feature branch into the base branch to avoid merge conflicts in the future once the feature is ready to be released.  All associated changes have already been reviewed in preceding pull requests.

### Test
1. Open navigation drawer.
2. Tap **_Settings_** item in drawer.
3. Notice **_Membership_** setting under **_Account_** section.
4. Notice **_Free_** in **_Membership_** setting.
5. Notice **_Upgrade_** button in **_Membership_** setting.
6. Notice **_Style_** setting under **_Appearance_** section.
7. Notice **_Default_** in **_Style_** setting.
8. Tap **_Style_** setting under **_Appearance_** section.
9. Notice **_Default_** style is selected.
10. Notice other styles are locked.
11. Tap any style other than **_Default_**.
12. Notice **_Style Locked_** dialog is shown.
13. Tap **Cancel_** button in dialog.
14. Notice dialog is dismissed.
15. Tap any style other than **_Default_**.
16. Notice **_Style Locked_** dialog is shown.
17. Tap **_Upgrade_** button in dialog.
18. Notice https://app.simplenote.com is shown in browser (if installed).

### Review
Only one developer is required to review these changes, but anyone can perform the review.

### Release
These changes do not require release notes.  `RELEASE-NOTES.txt` will be updated once the styles feature is complete.

PR #1252, opened 2020-12-24
Separate search sort order from the default sort order

### LINKED ISSUE #907: Search: Sort order changes affect note list sorting when app is closed during search
### Expected

### Observed

### Reproduced

2. Start a new search.
7. Start another new search.
10. Force close the app.

-|-|-|-

hichamboushaba:
### Fix

Fixes two issues:
1. Fixes #907, as we save temporarily the selected search sort order using the same key of the default sort order in the shared preferences, so if the fragment doesn't get a chance to call `onDetach` or `clearSearch`, the search sort order will replace the default sort order.
2. The sort order is totally broken in the develop branch, I think since merging #1227. As when we leave PreferenceActivity, we allways call `NavUtils.navigateUpFromSameTask`, the previous NotesActivity gets destroyed, and we create a new one, which means that `NoteListFragment` will call `detach` and override the saved sort order.

The fix I did is adding a new field that saves the search sort order locally in the fragment, and use it for everything related to the search, which avoids modifying the default sort order, and fixes both the above issues. 

#### Test sort order
1. Open preferences, and change the sort order.
2. Go back to the notes list.
3. Confirm that the sort order is applied.

#### Fix for overriding the sort order
2. Open notes list.
4. Change the sort order and/or direction for the search.
5. Force close the app.
6. Reopen the app and confirm the search order wasn't impacted by the selected search order.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -402,7 +402,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
                 query.include(Note.TITLE_INDEX_NAME);
                 query.where(Note.DELETED_PROPERTY, Query.ComparisonType.NOT_EQUAL_TO, true);
                 query.where(Note.TITLE_INDEX_NAME, Query.ComparisonType.LIKE, String.format("%%%s%%", filter));
-                PrefUtils.sortNoteQuery(query, requireContext(), true);
+                PrefUtils.sortNoteQuery(query, requireContext());
                 Cursor cursor = query.execute();
 
                 final int heightAutocomplete = DisplayUtils.dpToPx(requireContext(), cursor.getCount() * 48);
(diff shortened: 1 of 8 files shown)


theck13 (commented):
We have discussed making the search sort bar always visible (i.e. not just while searching).  In that case, the changes related to #907 wouldn't have been needed.  @SylvesterWilmott, do you think we'll make the search sort bar always visible?

I confirmed the sort order preference is not being saved when changed.  If we can separate those changes into another pull request, that might be a better solution since we may not get an answer about the sort order bar until after the holidays.

hichamboushaba:
> I confirmed the sort order preference is not being saved when changed. If we can separate those changes into another pull request, that might be a better solution since we may not get an answer about the sort order bar until after the holidays.

@theck13 the PR #1253 should fix the issue.

SylvesterWilmott:
>We have discussed making the search sort bar always visible (i.e. not just while searching)

Yeah, this has been in discussion for a while but nothing has materialised yet. I'll propose the idea to the team and see where we all stand and come back to this PR with the answer. Thanks @hichamboushaba

theck13:
We've decided to keep the search sort bar always visible in the note list.  @hichamboushaba, would you like to update this pull request, create a new pull request, or have me handle those changes?

hichamboushaba:
@theck13, I created a new PR for the new change: #1270, I'll close this one.

PR #1253, opened 2020-12-25
Improve theme and style changes handling

hichamboushaba:
### Fix
This improves theme/style changes handling, by centralizing the logic in the `ThemedAppCompatActivity`, and responding to changes in the same place, instead of relying `NavUtils.navigateUpTo`.
One of the advantages of this way too, is that it decouples it from the navigation logic, so if the navigation order gets changed later, or we add ability to open the settings from somewhere else, the theme changes will continue to be applied without having to change anything.

This also fixes the current bug of not saving changes of the sort order preference (see #1252 for details), as the `NotesActivity` won't get recreated until there is some theme or style changes.

#### Sort order fix
1. Open settings.
2. Change the sort order value.
3. Go back to notes list.
4. Confirm the new sort order value is applied.

#### Change theme
1. Open settings.
2. Change theme.
3. Confirm the theme is applied to the current screen.
4. Go back to the notes list.
5. Confirm the theme is applied here too.

#### Change style
1. Change return value of `PrefUtils.isPremium()` to true.
2. Open settings, then open style settings.
3. Change style.
4. Confirm the style is applied to the current screen.
5. Go back to the previous screens.
6. Confirm the theme is applied here too.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/AboutActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/AboutActivity.java
@@ -8,7 +8,6 @@ import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.Toolbar;
 
 import com.automattic.simplenote.utils.DrawableUtils;
-import com.automattic.simplenote.utils.ThemeUtils;
 
 public class AboutActivity extends AppCompatActivity {
     @Override
--- a/Simplenote/src/main/java/com/automattic/simplenote/AboutActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/AboutActivity.java
@@ -13,7 +13,6 @@ import com.automattic.simplenote.utils.ThemeUtils;
 public class AboutActivity extends AppCompatActivity {
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
-        ThemeUtils.setTheme(this);
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_about);
         Toolbar toolbar = findViewById(R.id.toolbar);
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
@@ -25,7 +25,6 @@ import com.automattic.simplenote.utils.AppLog;
 import com.automattic.simplenote.utils.AppLog.Type;
 import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.NetworkUtils;
-import com.automattic.simplenote.utils.ThemeUtils;
 import com.automattic.simplenote.widgets.NoteEditorViewPager;
 import com.automattic.simplenote.widgets.RobotoMediumTextView;
 import com.google.android.material.tabs.TabLayout;
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
@@ -76,7 +76,6 @@ public class NoteEditorActivity extends ThemedAppCompatActivity {
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
-        ThemeUtils.setTheme(this);
         super.onCreate(savedInstanceState);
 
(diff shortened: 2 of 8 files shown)


Simplenote/src/main/java/com/automattic/simplenote/StyleActivity.java
```diff
@@ -86,17 +80,20 @@ public void onCreate(Bundle savedInstanceState) {
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
         if (item.getItemId() == android.R.id.home) {
-            NavUtils.navigateUpFromSameTask(StyleActivity.this);
+            finish();
         }
 
         return super.onOptionsItemSelected(item);
     }
```
theck13:
```suggestion
```

Since we are no longer doing something special when tapping the back arrow in the app bar, we can remove `onOptionsItemSelected` entirely.

Simplenote/src/main/java/com/automattic/simplenote/ThemedAppCompatActivity.java
```diff
@@ -12,10 +16,49 @@
...
+        if(mThemeChanged) {
+            recreate();
+            mThemeChanged = false;
+        }
+    }
+
+    @Override
+    public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
+        if(key.equals(PrefUtils.PREF_THEME) || key.equals(PrefUtils.PREF_STYLE_INDEX)) {
+            if(getLifecycle().getCurrentState().isAtLeast(Lifecycle.State.RESUMED)) {
```
theck13:
```suggestion
        if (mThemeChanged) {
            recreate();
            mThemeChanged = false;
        }
    }

    @Override
    public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
        if (key.equals(PrefUtils.PREF_THEME) || key.equals(PrefUtils.PREF_STYLE_INDEX)) {
            if (getLifecycle().getCurrentState().isAtLeast(Lifecycle.State.RESUMED)) {
```

There is a space missing between the `if` and the `(` in these three statements.

Simplenote/src/main/java/com/automattic/simplenote/StyleActivity.java
```diff
@@ -47,14 +47,8 @@
     private LinearLayoutManager mLayoutManager;
```
theck13, line 45: We can remove the `import androidx.core.app.NavUtils;` statement since it is no longer used.

Simplenote/src/main/java/com/automattic/simplenote/AboutActivity.java
```diff
@@ -13,7 +13,6 @@
 public class AboutActivity extends AppCompatActivity {
```
theck13, line 12: We can remove the `import com.automattic.simplenote.utils.ThemeUtils;` statement since it is no longer used.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
```diff
@@ -76,7 +76,6 @@
```
theck13, line 75: We can remove the `import com.automattic.simplenote.utils.ThemeUtils;` statement since it is no longer used.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesActivity.java
```diff
@@ -24,14 +24,8 @@
     private PasscodePreferenceFragmentCompat mPasscodePreferenceFragment;
```
theck13, line 20: We can remove the `import android.content.Intent;`, `import androidx.core.app.NavUtils;`, and `import com.automattic.simplenote.utils.ThemeUtils;` statements since they are no longer used.

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
```diff
@@ -62,7 +62,6 @@
```
theck13, line 61: We can remove the `import com.automattic.simplenote.utils.ThemeUtils;` statement since it is no longer used.

theck13 (changes requested):
Good idea to consolidate the style and theme changes into `ThemedAppCompatActivity`.

I left some comments about code style and import statements.  Android Studio can automatically remove unused import statements by checking the **_Optimize imports_** option for commits.

hichamboushaba:
@theck13 I amended the last commit to address your remarks, ready for another review 🙏

theck13 (approved):

PR #1260, opened 2020-12-31
Reset the buckets if the user is unauthenticated

### LINKED ISSUE #1258: Logout due to token invalidation doesn't reset the local database
### Expected
If the token of my account get's invalidated or expired, the app will log the user out, the expected behavior is that the buckets will be reset, to reset local storage.

### Observed
When the token is expired or invalidated, the logout action doesn't trigger buckets reset, which results in keeping and displaying the notes from the local storage if the user signs in using a different account.

### Reproduced
1. Log into the app.
2. Close the app.
2. Go to the account settings in the web app, and change the password.
3. Open the app again.
4. Notice that the login screen has been shown.
5. Log in using a different account.
6. Notice that the notes of the first account are still displayed.

hichamboushaba:
### Fix
Fixes #1258 by making sure to reset the buckets if the user is unauthenticated (token expired, or password changed...), and avoids keeping the old notes if the user signs in using a different account.

### Test
1. Sign in to the app.
2. Force close the app.
3. Trigger the unauthentication scenario (personally I deleted the simperium.xml shared pref file using the device explorer in Android Studio, a password change on the web app triggers it too)
4. Open the app.
5. Notice that the login screen is displayed
5. Sign in using a different account.
6. Confirm that notes from the old account have been deleted.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -1278,6 +1278,13 @@ public class NotesActivity extends ThemedAppCompatActivity implements NoteListFr
         // Remove Passcode Lock password
         AppLockManager.getInstance().getAppLock().setPassword("");
 
+        // Reset Buckets
+        Simplenote application = (Simplenote) getApplication();
+
+        application.getNotesBucket().reset();
+        application.getTagsBucket().reset();
+        application.getPreferencesBucket().reset();
+
         Intent intent = new Intent(NotesActivity.this, SimplenoteAuthenticationActivity.class);
         startActivityForResult(intent, Simperium.SIGNUP_SIGNIN_REQUEST);
     }
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -43,6 +43,7 @@ import com.automattic.simplenote.models.Note;
 import com.automattic.simplenote.models.Tag;
 import com.automattic.simplenote.utils.AppLog;
 import com.automattic.simplenote.utils.AppLog.Type;
+import com.automattic.simplenote.utils.AuthUtils;
 import com.automattic.simplenote.utils.CrashUtils;
 import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.DrawableUtils;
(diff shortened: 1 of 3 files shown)


theck13 (commented):
The code looks good and these changes _do_ reset the buckets as described.  The `NotesActivity.startLoginActivity()` method also clears the WordPress login data and resets the passcode lock password.  That's almost everything the [`PreferencesFragment.logOut()`](https://github.com/Automattic/simplenote-android/blob/cb0660765cab6a8caa9d2872f79db429d29c7946/Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java#L309-L348) method does.  Should we add the other items too (i.e. reset analytics and update widgets)?  If we wanted, we could abstract all those items to a separate `AuthUtils` class so that both `NotesActivity.startLoginActivity()` and `PreferencesFragment.logOut()` can use that class for unauthenticated users and logging out.  Let me know if you would like to include any of that in these changes or if you would like to keep this pull request as is.

hichamboushaba:
Thanks @theck13 for the review, good point 👍, it's done in 590de405015b7997fef31d439cc54c0913a313bf, I also added some log entries to make analyzing the log easier, and clarify the cause of stopping the buckets, if you have any remarks regarding those, please let me know 🙂

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -1251,6 +1259,7 @@ public void run() {
                 runOnUiThread(new Runnable() {
                     @Override
                     public void run() {
+                        AppLog.add(Type.ACCOUNT, "Token not authorized");
```
theck13:
```suggestion
                        AppLog.add(Type.ACCOUNT, "Access token not authorized");
```

Should we make this log message the same as the other statement when an access token is not authorized just for consistency?  If we want to differentiate this statement from the other, we could make it something like "Access token not authorized (NotesActivity.onUserStatusChange)" for this one and "Access token not authorized (NotesActivity.userAuthenticationIsInvalid)" for the previous one.

theck13 (commented):
Originally, the application logs were meant to diagnose syncing issues, but they're grow a little to cover other actions for troubleshooting.  Even though we don't get many users reporting issues with authentication being invalidated and the log statements may not be used much, I don't think they will hurt anything.  One thing to keep in mind is that the application log is a queue with a maximum of 100 entries.  So, if we add too many log statements, the logs may not be as helpful diagnosing something that has been pushed out of the queue.  I don't think the authentication log statements added in the latest commit will cause a problem though.

The new `AuthUtils` class and associated changes look good to me.  I left one comment about a log statement message, but I'm fine with merging these changes as is if you would like.

hichamboushaba:
Thanks @theck13, I applied your suggestion 6fd2b4f

theck13 (approved):

PR #1261, opened 2020-12-31
Fix for having a wrong them in the webviews when the app is launched from the widget

### LINKED ISSUE #1259: Dark them not applied to the markdown preview when launched from the widget
### Expected
The dark them is fully applied regardless of how the app is launched.

### Observed
When launching the editor from the widget, the markdown preview is using the light theme, where the rest is using the correct dark theme.

### Reproduced
1. Add the notes list widget to the home screen.
2. Enable dark theme from the app's settings.
3. Close the app.
4. Click on one markdown note in the widget.
5.  Notice that the markdown preview is white.

hichamboushaba:
### Fix
Fixes #1259.
When inflating the first Webview in the app, the `uiMode` value of the configuration may be reset to the system's mode, the issue: https://issuetracker.google.com/issues/37124582 explains the issue, and though it says it's fixed since `appcompat 1.1.0-alpha03`, it's happening in 1.1.0 and other commenters noticed too.
The fix is by re-applying the nightMode in NoteMarkdownFragment's `onResume`, the call will either recreate the activity if the uiMode is different or do nothing if it's already correct.

### Test
1. Add the notes list widget to the home screen.
2. Enable dark theme from the app's settings.
3. Close the app (to confirm the fix, please force close it).
4. Click on one markdown note in the widget, and open the preview tab.
5. Notice that the markdown preview is dark.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/ThemedAppCompatActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/ThemedAppCompatActivity.java
@@ -51,6 +51,9 @@ abstract public class ThemedAppCompatActivity extends AppCompatActivity implemen
     public void recreate() {
         Intent intent = new Intent(this, getClass());
         intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+        if (getIntent().getExtras() != null) {
+            intent.putExtras(getIntent().getExtras());
+        }
         startActivity(intent);
         overridePendingTransition(android.R.anim.fade_in, android.R.anim.fade_out);
         finish();
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.java
@@ -13,6 +13,7 @@ import android.webkit.WebView;
 import android.webkit.WebViewClient;
 
 import androidx.annotation.NonNull;
+import androidx.appcompat.app.AppCompatActivity;
 import androidx.core.view.MenuCompat;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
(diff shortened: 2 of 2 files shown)


Simplenote/src/main/java/com/automattic/simplenote/ThemedAppCompatActivity.java
```diff
@@ -51,6 +51,9 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
     public void recreate() {
         Intent intent = new Intent(this, getClass());
         intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+        if (getIntent().getExtras() != null) {
+            intent.putExtras(getIntent().getExtras());
```
hichamboushaba, line 55: We need to pass the bundle of current intent before recreating the activity, I missed this on the theme changes PR, and it caused a crash when the NoteEditorActivity was recreated to fix the theme.

theck13: How do you reproduce the crash before these changes?  I wasn't able to get the app to crash after opening a note from the list or from a widget and toggling the theme on a small device.

hichamboushaba: Sorry I wasn't clear enough 🙂, so first to be clear, I should've included this code by default, as when recreating the activity we need to pass the extras that we had before to avoid any unexpected behavior, and Android does this by default when we call the default `recreate` implementation: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/app/ActivityThread.java;l=5419

In case we don't do this, when recreating `NoteEditorActivity` for example, we will have a null noteId here: https://github.com/Automattic/simplenote-android/blob/d43aeb4710c41a5f943430034b6cbcb0fe6cd8f5/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java#L101.
To reproduce the crash, pull the changes of this PR, then remove this change on `ThemedAppCompatActivity`, and try the exact steps described in the PR description, the app should crash after clicking the note in the widget.

theck13 (approved):

PR #1262, opened 2020-12-31
Note Info Sheet: Update with updates to note

dmsnell:
Previously the note info sheet would render the information about a note on open
and then statically hold those values until it's closed and opened again.

In this patch we are adding listenwers to the application to watch for changes
to the note and then re-displaying the info so that it stays current.

There are a couple caveats:
 - Because I wasn't sure how the UI is being arranged the references don't update.
   When I blindly called `getReferences(note)` on updates it blanked out the
   list of references.
 - Similarly I tried setting the visibility of the sync label to `VISIBLE` when
   updating the sync value but the label doesn't appear until re-opening the sheet.
   I think this is related to the references display issue.

Therefore in order to see the sync time if the sheet was opened when none was available
you need to re-open the sheet. If you open the sheet and a sync time is already avialable
it will work as expected. If you want to see the updated references that reflect changes
in other notes you will also need to re-open the sheet.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/InfoBottomSheetDialog.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/InfoBottomSheetDialog.java
@@ -26,6 +26,7 @@ import com.automattic.simplenote.utils.SimplenoteLinkify;
 import com.google.android.material.bottomsheet.BottomSheetBehavior;
 import com.google.android.material.bottomsheet.BottomSheetDialog;
 import com.simperium.client.Bucket;
+import com.simperium.client.BucketObjectMissingException;
 
 import java.util.ArrayList;
 import java.util.Calendar;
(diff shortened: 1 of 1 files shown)


theck13 (commented):
I like the idea of these changes, but the caveats make me think we shouldn't merge this.  I'm not sure why, but the bottom sheet dialog does not like layout changes.  That's why the ***Synced*** field was never shown and the ***Reference In*** list was blanked out when a network change was received.  I tried multiple things with the `BottomSheetDialogBase` and `InfoBottomSheetDialog` classes, but nothing worked.  Even hiding a previously visible view does not update the layout bounds as expected.  That's why I think showing the ***Synced*** view has no effect.  The bottom sheet just doesn't update.  I tried removing the `NestedScrollView` from the `bottom_sheet_info` layout as well, but nothing changed.

Since the ***Synced*** date/time is probably the view that would benefit the most from a network change listener and that field is not working as expected, I don't think we should merge this until we have a better solution.

dmsnell:
Thanks for looking into it @theck13 - strange that it's so hard to get it to obey.

PR #1140, opened 2020-09-04
WIP: Add new last-sync time indicator

dmsnell:
Following Simperium/simperium-android#229 this adds a new listener,
the `LastSyncTimeCache`, to track the sync status of notes. It tracks:

 - are the local changes that have not received confirmation from the server?
 - when was the last time changes from the server arrived?

These two pieces of information can be presented to indicate:

 - is it safe for me to quit out of the app? (`isSynced`)
 - do I have the edits I made on my other device? (`lastSyncTime`)

Components in the app may poll the cache for individual notes or subscribe
as a listener and receive updates.

## Todo

 - [ ] persist `LastSyncTimeCache` so it's available across app sessions
 - [x] listen during indexing operations so we don't have to wait for changes to occur before populating the values
 - [x] add UI in the note info panel showing last sync time
 - [ ] stop background sync process once all local changes have been acknowledged
 - [x] allow multiple listeners to the `LastSyncTimeCache` (currently only one is allowed)
 - [x] remove extra log statements

PR #1264, opened 2021-01-05
Add Sync Persistence

theck13:
### Fix
Add persistence to the note last sync time cache added in https://github.com/Automattic/simplenote-android/pull/1140.  Since that pull request has not been merged, I'm targeting the associated branch, `add/last-sync-time-indicator`.  These changes save the last sync time for each note when a remote change occurs with the `onNetworkChange` callback and when a local change has been confirmed as received by the server with the `onSyncObject` callback.  The last sync times are saved as preferences.  The preferences option was chosen for storage since preferences are meant to store private, primitive data in key-value pairs and all that is needed for the last sync time data is a note ID as a `String` (i.e. the key) and a time as a `long` (i.e. the value).  Originally, a [Room database](https://developer.android.com/training/data-storage/room) was used for storage.  The storage was changed to preferences since databases are meant for structured data, only one item will be retrieved at a time using the key, the note ID and last sync time are not sensitive data, and using Room would have required the [`allowMainThreadQueries()` method](https://developer.android.com/reference/androidx/room/RoomDatabase.Builder#allowMainThreadQueries()) for executing queries in order to avoid using an [`AsyncTask`, which has been deprecated](https://developer.android.com/reference/android/os/AsyncTask).

#### Note
For some unknown reason, all notes do not trigger the `onNetworkChange` callback when first loading notes after login.  When testing various accounts, approximately 70% of notes triggered `onNetworkChange` and caused a last sync time to be saved for those notes.  All other notes that did not trigger `onNetworkChange` would not have a last sync time until those notes were edited.  Similarly, if a user is already logged in when installing the app with these changes, notes will not have a last sync time value cached or saved until the note is edited remotely or locally and the device has an active network connection.  Since the last sync time is most important after an edit has been made and these changes will update the last sync time after an edit, solving the problems with first loading notes after login and determining a last sync time for users already logged in can be addressed in a future enhancement.

### Test
Since the last sync time cache can be affected by local and remote edits, using two devices to test is best.  I used an Android emulator and the web app.
1. **Android**: Tap any note in list.
2. **Web**: Tap note in list from Step 1.
3. **Android**: Tap ***Information*** action in app bar.
4. **Android**: Notice ***Synced*** field is hidden.
5. **Android**: Tap anywhere outside ***Information*** sheet to dismiss.
6. **Web**: Edit note from Step 1.
7. Wait ten seconds.
8. **Android**: Tap ***Information*** action in app bar.
9. **Android**: Notice ***Synced*** field is shown.
10. **Android**: Enable airplane mode.
11. **Android**: Edit note from Step 1.
12. **Android**: Tap ***Information*** action in app bar.
13. **Android**: Notice ***Synced*** field is shown and unchanged.
14. **Android**: Notice ***Modified*** field is shown and changed.
15. **Android**: Tap anywhere outside ***Information*** sheet to dismiss.
16. Wait one minute.
17. **Android**: Disable airplane mode.
18. Wait ten seconds.
19. **Android**: Tap ***Information*** action in app bar.
20. **Android**: Notice ***Synced*** field is shown and changed.
21. **Android**: Notice ***Modified*** field is shown and unchanged.
22. **Android**: Tap recents/overview system navigation button.
23. **Android**: Swipe app away from recents/overview list.
24. **Android**: Launch app.
25. **Android**: Tap note from Step 1.
26. **Android**: Tap ***Information*** action in app bar.
27. **Android**: Notice ***Synced*** field is shown with value from Step 20.

### Review
Only one developer is required to review these changes, but anyone can perform the review.  @dmsnell, I requested you as a reviewer since you created the last sync time cache.  Don't feel obligated to review this pull request, but you can if you want.  I just wanted to notify you of this enhancement to your last sync time cache feature.

### Release


Simplenote/src/main/java/com/automattic/simplenote/LastSyncTimeCache.java
```diff
@@ -1,23 +1,33 @@
...
+    private final Set<SyncTimeListener> mListeners = new HashSet<>();
+    private final SharedPreferences mPreferences;
+
+    public LastSyncTimeCache(Context context) {
+        mPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+    }
+
+    public long getLastSyncTime(String key) {
+        return mPreferences.getLong(key, DEFAULT_LAST_SYNC_TIME);
     }
```
hichamboushaba: IMO returning a Calendar here instead of making the conversion in InfoBottomSheetDialog will make the code easier to read.

Simplenote/src/main/java/com/automattic/simplenote/LastSyncTimeCache.java
```diff
@@ -38,33 +48,28 @@ private void notifyListeners(String entityId) {
...
 
+    public void removeSyncTime(String entityId) {
+        mPreferences.edit().remove(entityId).apply();
+    }
+
     private void updateSyncTime(String entityId) {
-        Calendar now = Calendar.getInstance();
-        mSyncTimes.put(entityId, now);
+        Calendar calendar = Calendar.getInstance();
+        long millis = calendar.getTimeInMillis();
```
hichamboushaba: System.currentTimeMillis() is more efficient, and will return the same result.

Simplenote/src/main/java/com/automattic/simplenote/LastSyncTimeCache.java
```diff
@@ -1,23 +1,33 @@
...
     private static final String TAG = LastSyncTimeCache.class.getSimpleName();
 
-    public Calendar getLastSyncTime(String key) {
-        return mSyncTimes.get(key);
+    private final HashSet<String> mUnsyncedKeys = new HashSet<>();
+    private final Set<SyncTimeListener> mListeners = new HashSet<>();
+    private final SharedPreferences mPreferences;
+
+    public LastSyncTimeCache(Context context) {
+        mPreferences = PreferenceManager.getDefaultSharedPreferences(context);
```
hichamboushaba: WDYT about moving those to a dedicated preferences file, instead of the default one? IMO think it will be cleaner, and keep using the default preferences file for "user preferences"

hichamboushaba (commented):
Looks good, I just left some comments, but nothing blocking.

And a question before approving, shouldn't we clear the cache on logout?

Simplenote/src/main/java/com/automattic/simplenote/LastSyncTimeCache.java
theck13: I updated the return type from `long` to `Calendar` in 5febef06.

theck13: I updated the return type from `long` to `Calendar` in 5febef06.

Simplenote/src/main/java/com/automattic/simplenote/LastSyncTimeCache.java
theck13: I added an exclusive preferences file for last sync times in 577db428.

Simplenote/src/main/java/com/automattic/simplenote/LastSyncTimeCache.java
theck13: Yeah, we could use `System.currentTimeMillis()`, but then we would need to do some calculation for the log statement below to convert the milliseconds value into a human-reading format.  That's why a `Calendar` was used.

theck13 (commented):
I added clearing the last sync time preferences file in 67e45ab0.

hichamboushaba (approved):
Looks good.
Do we want to merge this, or wait for @dmsnell to take a look as well.

dmsnell:
Going to share a more general comment here instead of poking at individual lines. First of all thanks for spearheading this. I'm delighted at how easy we can persist these values; I saw a note in the Android docs about the preferences being inefficient but I doubt that it will have any meaningful impact on us with how infrequently we update them.

My first thought when seeing this was to think of a way we could persist without changing `LastSyncTimeCache`, which now seems like a poor name for that thing. At least if we were to move the coupling directly into `Simplenote.java` then the "cache" is little more than a listener we could erase, at which point we'd be jumbling it all into `Simplenote`.

```java
// Simplenote.java
mNoteSyncTimes.addListener(new SyncTimeListener() {
	void onUpdate(String entityId, Calendar lastSyncTime, boolean isSynced) {
		mPreferences.edit().putLong(entityId, lastSyncTime.getTimeInMillis()).apply();
	}
});
```

At this point the `LastSyncTimeCache` remains untouched and the Info Sheet can read the preference directly or through a helper.

```java
mPreferences.getLong(noteId, null);
```

I noted that the preferences includes its own listener API that gets called when the preferences change, which could be similarly useful.

---

My other thought was to pass in a more generic `get/put` interface in the constructor of `LastSyncTimeCache` to avoid coupling the main app and context but I wasn't thrilled with stewings on that either.

```java
public Calendar getLastSyncTime(String noteId) {
	try {
		Calendar instant = Calendar.getInstance();
		instant.setTimeInMillis(mStore.get(noteId));
		return instant;
	} catch {
		return null;
	}
}
```

This seems a bit too jumbled inside the sync cache.

---

Then I thought about it more and it seems like the only time we actually need the persistence is on boot. The persistent store should be updated when we get new updates but only at load time do we need to read from it. The new `null` checks and smashing of `-1` into the time type as a sentinel for `null` makes me think there might be a cleaner API to preserve for in-app consumers.

Maybe we could consider loading the persisted data at boot as a more generic set of data and then listening in the main app to persist it while preserving the existing API for accessing sync info through the cache.

```java
// LastSyncTimeCache.java
public LastSyncTimeCache(Map<String, Calendar> storedSyncTimes) {
    mSyncTimes.putAll(storedSyncTimes);
}
```

```java
// Simplenote.java
mSyncTimes = context.getSharedPreferences(PREFERENCES_LAST_SYNC_TIME, Context.MODE_PRIVATE);

// update: I just realized that we'd need to convert these from ms to Calendar
mNoteSyncTimes = new LastSyncTimeCache(mSyncTimes.getAll());

mNoteSyncTimes.addListener(new SyncTimeListener() {
	void onUpdate(String noteId, Calendar lastSyncTime, boolean isSynced) {
		mSyncTimes.edit().putLong(entityId, lastSyncTime.getTimeInMillis()).apply();
	}

	void onRemove(String noteId) {
		mSyncTimes.edit().remove(noteId).apply();
	}
});
```

With this we still get a fully-testable sync-time class that pretty specifically does little beyond mangling the logic between "note bucket operations" and "note sync times" then all the persistence sits in one little spot in the top-level app (or in some subsidiary class should we desire).

What are your thoughts on this approach instead? Main points are:
 - with exception of the `onRemove` method the existing API for sync changes remains the same, always presents valid data to listeners (no nulls, no `-1`, no missing data)
 - `LastSyncTimeCache` doesn't need to couple itself to the app or to the means of persistence
 - The full cycle of persistence is visible in one spot and it should be obvious how to change it

Also it seems clear by now that `LastSyncTimeCache` could have a better name. Since writing it it's become clearer that we're using it more specifically for notes and less specifically as a cache. Here are a couple ideas:
 - `SyncTimeAdapter` if kept as a generic bucket thing
 - `NoteSyncTimes` if specified to the note bucket

It seems valuable still on its own at least to keep it around since the Simperium API doesn't present this particular need very well: when has an entity been synchronized in either direction with the server? Has the server confirmed my latest local copy?

---

Sorry for the long review; I hope you find it useful feedback. Would like to hear your thoughts on the coupling and the idea of separating out persistence from the `LastSyncTimeCache` as proposed above.

theck13:
I'm fine with separating the persistence from `LastSyncTimeCache` and renaming that class to `NoteSyncTimes`.  It might be best if you make those changes in https://github.com/Automattic/simplenote-android/pull/1140 once this pull request is merged and I can review them once you hit the "Ready for review" button in that pull request.  Does that sound good?

dmsnell:
> It might be best if you make those changes in #1140 once this pull request is merged and I can review them once you hit the "Ready for review" button in that pull request. Does that sound good?

Which changes? for the rename that seems fine but if we split out the logic it doesn't make as much sense to me to merge this as-is, which rips apart `LastSyncTimeCache`, and then immediately rip apart that in another PR.

I could propose or directly commit a change on this branch if that's what you are talking about?

theck13:
> I could propose or directly commit a change on this branch if that's what you are talking about?

That works too.

dmsnell:
up for your review @theck13 - I hope I didn't overstep with the changes I pushed here.

in summary:
 - separated the persistence and moved it into `Simplenote` where the context and connection with the main app and its preferences is more contained
 - created a helper class for persistence
 - generalized the `SyncTimes` class since there's really nothing note-specific in it. as far as I can tell the only thing specific to notes is the name of the preferences key

theck13 (commented):
@dmsnell, thanks for pushing those changes.  I made some minor optimizations and reorganized a couple things.  None of the functionality changed.  Basically, I just moved things around.  I tested the latest with the steps from the ***Test*** section above and everything works as expected.  If the latest changes look good to you, let me know.  If something doesn't looks right, let me know.  I'll wait for your approval before merging.

dmsnell (approved):
Thanks again @theck13 - this looks great

PR #1267, opened 2021-01-06
Fix the issue of not showing the widget configuration screen when applock is enabled

### LINKED ISSUE #725: Widget Configuration is Not Shown When PIN Lock is On and Simplenote Was Recently Opened
### Expected
The note selection dialog is always shown after placing the ***Note*** widget on the home screen even when the PIN lock is on.  If the app was recently opened and PIN lock is on, the PIN lock is shown first and the note selection dialog is shown after entering the correct passcode.

### Observed
The note selection dialog is sometimes shown after placing the ***Note*** widget on the home screen even when the PIN lock is on.  If the app was recently opened and PIN lock is on, the PIN lock is shown first and the last screen is shown after entering the correct passcode.

### Reproduced
These steps assume the AOSP launcher (i.e. Nexus/Pixel launcher). The steps to add the widget to the home screen may be slightly different based on the launcher used. Other steps should be the same regardless of launcher.
1. Enable PIN lock.
2. Exit app and remove it from recent apps view.
3. Long-press home screen.
4. Tap ***Widgets*** option.
5. Scroll to ***Simplenote*** widget.
6. Notice ***To-Do List*** preview image.
7. Long-press ***Note*** widget.
8. Place ***Note*** widget on home screen.
9. Notice PIN lock is shown.
10. Enter passcode.
11. Notice ***Select Note*** dialog with note list.
12. Select note from list.
13. Notice widget with selected note is added to home screen.
14. Open app.
15. Open navigation drawer.
16. Tap ***Settings*** item.
17. Notice ***Settings*** is shown.
18. Tap home system navigation button.
19. Long-press home screen.
20. Tap ***Widgets*** option.
21. Scroll to ***Simplenote*** widget.
22. Notice ***To-Do List*** preview image.
23. Long-press ***Note*** widget.
24. Place ***Note*** widget on home screen.
25. Notice PIN lock is shown.
26. Enter passcode.
27. Notice ***Settings*** is shown.

-|-|-|-
Google|Pixel 3|9.0.0|1.8.0
Google|Pixel 2|7.1.1|1.8.0

hichamboushaba:
### Issue description
As described in the issue #725, the widget configuration screen is not displayed if the applock is enabled, and the app was already open.
The cause is that the `DefaultAppLock` of the `Passcode-Lock` library starts the `PasscodeUnlockActivity` using the application context, and the flag: `FLAG_ACTIVITY_NEW_TASK`: https://github.com/wordpress-mobile/PasscodeLock-Android/blob/804db0580e7719a36784b8a7c9751614808246ab/library/src/org/wordpress/passcodelock/DefaultAppLock.java#L53-L55. But as we don't customize the [taskAffinity](https://developer.android.com/guide/components/activities/tasks-and-back-stack#Affinities), if the app has an existing task, it will be brought to foreground, and the `PasscodeUnlockActivity` will be added to it, for example:
`Running activities (most recent first):
      TaskRecord{36c9e37 #884 A=com.automattic.simplenote.debug U=0 StackId=382 sz=3}
        Run #2: ActivityRecord{daf4dd u0 com.automattic.simplenote.debug/org.wordpress.passcodelock.PasscodeUnlockActivity t884}
        Run #1: ActivityRecord{ce9c436 u0 com.automattic.simplenote.debug/com.automattic.simplenote.PreferencesActivity t884}
        Run #0: ActivityRecord{33d7dcc u0 com.automattic.simplenote.debug/com.automattic.simplenote.NotesActivity t884}`

But as the widget configuration activity is started by the launcher, it will be added to its task, and not the app's task, so when we finish `PasscodeUnlockActivity` we will go back to the previous screen in the app's task, and not the widget configuration activity.

To solve this there is two ways:
1. The solution suggested in this PR, call `startActivity` from the resumed activity, and not the applicationContext, and don't provider `FLAG_ACTIVITY_NEW_TASK`, this will give us the following task when trying to add the widget:
` Running activities (most recent first):
      TaskRecord{39ed39c #891 A=com.google.android.apps.nexuslauncher U=0 StackId=389 sz=3}
        Run #2: ActivityRecord{a27d305 u0 com.automattic.simplenote.debug/org.wordpress.passcodelock.PasscodeUnlockActivity t891}
        Run #1: ActivityRecord{1d736e4 u0 com.automattic.simplenote.debug/com.automattic.simplenote.NoteWidgetDarkConfigureActivity t891}
        Run #0: ActivityRecord{6c89f9b u0 com.google.android.apps.nexuslauncher/com.android.launcher3.proxy.ProxyActivityStarter t891}
`

2. use a custom `taskAffinity` for the `PasscodeUnlockActivity` to make sure starting it will always create a new task, and then when we go back, we will go to the launcher's task, which contains the widget configuration activity. But it will cause the appearance of the default animation of navigating to a different task, which is similar to navigating to another app:
passcode

Personally, I prefer solution 1, but it should be implemented in the passcode library, and not as I did here, I wanted your opinion before creating a PR in the library code, as I don't know the history, and if the use of the `applicationContext` was for some reason, wdyt @theck13 ?

### Fix
The solution suggested by this PR is to extend the `DefaultAppLock` and make small adjustments just for the widget configuration activities while keeping the default behavior for the rest.

### Test

1. Enable PIN lock.
2. Open app.
3. Open navigation drawer.
4. Tap Settings item.
5. Notice Settings is shown.
6. Tap home system navigation button.
7. Long-press home screen.
8. Tap Widgets option.
9. Scroll to Simplenote widget.
10. Notice To-Do List preview image.
11. Long-press Note widget.
12. Place Note widget on home screen.
13. Notice PIN lock is shown.
14. Enter passcode.
15. Confirm that the Select Dialog is shown.

### Review

### Release


Simplenote/src/main/java/com/automattic/simplenote/CustomAppLock.java
```diff
@@ -0,0 +1,233 @@
...
+    @Override
+    public void onActivityPaused(Activity activity) {
+        if (!isExemptActivity(activity.getClass().getName())) mLostFocusDate = new Date();
+    }
+
+    @Override
+    public void onActivityResumed(Activity activity) {
+        if (!isExemptActivity(activity.getClass().getName()) && shouldShowUnlockScreen()) {
+            Intent i = new Intent(activity.getApplicationContext(), PasscodeUnlockActivity.class);
+            activity.startActivity(i);
```
hichamboushaba: The only change from the `DefaultAppLock` is the removal of the `FLAG_ACTIVITY_NEW_TASK` flag, and using the activity instead of the applicationContext.

theck13 (commented):
The passcode lock library is very old and I doubt it was developed with widgets in mind.  It wasn't meant to be a full security feature like the operating system locking mechanism.  It was designed to safeguard against inadvertent modification or deletion rather than data protection.  The passcode lock library is in maintenance mode and the only changes we want to make are ones that are absolutely necessary.  That's pretty much all I know about the library history.

I wouldn't be surprised if Simplenote is the only app using the passcode lock library to be honest.  So, we don't have to worry about developing the library with any apps in mind other than Simplenote.  Even if other apps are using the library, they don't have to update if they don't want to.

I'm not sure why the application context was used rather than the activity.  I don't think the consequence in Solution 2 of the default navigation animation is terrible.  It might be okay with the widget dialog interface.  With either solution, I would choose the simplest option possible just so we don't introduce any potential issues in the future since the library is in maintenance mode.

As for these changes and the issue they are trying to solve, I wouldn't spend too much time on it.  The issue occurs in a very specific situation and we haven't received many complaints or reports from users about it.  Even if we do, there is a simple workaround by removing Simplenote from the recents/overview list before adding a widget.

hichamboushaba:
Thanks @theck13 for the insight, to keep the changes minimal, I refactored the solution to don't touch the passcode library for now, and just overwrite its behavior for the widget configuration screens: 6d85091, please let me know what you think.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -72,7 +73,11 @@ public void onCreate() {
         super.onCreate();
 
         CrashUtils.initWithContext(this);
-        AppLockManager.getInstance().enableDefaultAppLockIfAvailable(this);
+        if (DefaultAppLock.isSupportedApi()) {
+            SimplenoteAppLock appLock = new SimplenoteAppLock(this);
+            AppLockManager.getInstance().setCurrentAppLock(appLock);
+            appLock.enable();
+        }
```
theck13:
```suggestion
        SimplenoteAppLock appLock = new SimplenoteAppLock(this);
        AppLockManager.getInstance().setCurrentAppLock(appLock);
        appLock.enable();
```

Since the `minSdkVersion` is `23` and `DefaultAppLock.isSupportedApi()` is `Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH`, the `if` condition is always `true`, right?  If so, we can remove it.

theck13 (commented):
This solution looks better to me.  I tested it on the note and note list widgets as well as the launcher and new note shortcut with the passcode lock enabled and disabled.  It works as expected in my tests.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
hichamboushaba: Good catch, done d2a255c

theck13 (approved):

PR #1270, opened 2021-01-08
Make the sort bar always visible in the notes list

### LINKED ISSUE #907: Search: Sort order changes affect note list sorting when app is closed during search

hichamboushaba:
### Fix
This PR makes the sort bar always visible in the notes list, instead of being visible only during the search:

It's a draft as I have one question: should it be visible also when viewing the trashed notes? Can you check it @SylvesterWilmott 

### Test
1. Open the notes list.
2. Notice that the sort bar is visible.
3. Change the sort order, and confirm that it's applied.
4. Go to Settings.
5. Confirm that the last sort order in the bar is saved to settings.

### Review

### Release
`RELEASE-NOTES.txt` was updated in fb66fd149cec59e1fe16a2f14d44e44d9dbf8899 with:
> Make the sort bar always visible in the notes list


hichamboushaba:
Closing this PR as more time is needed to study this, and make a decision on how to implement it.

PR #1284, opened 2021-01-25
Update Note Sort

### LINKED ISSUE #907: Search: Sort order changes affect note list sorting when app is closed during search

theck13:
### Fix
Update the sort bar visibility and text to close #907.  The sort bar, which was only visible after submitting a search query, is updated to be always visible when searching and not.  The text is updated to match other platforms.  See the table below for details.

Before|After
-|-
Alphabetically, A-Z|Name: A-Z
Alphabetically, Z-A|Name: Z-A
Newest created date|Created: Newest
Oldest created date|Created: Oldest
Newest modified date|Modified: Newest
Oldest modified date|Modified: Oldest

The text updates also affect the **_Sort order_** item in **_Settings_**.  The order of the sort options is the sort bar match the designs.  The order of the sort options in the setting dialog are left in the order they were previously for a couple reasons.  The order is maintained for backwards compatibility with updated apps and the plan is to remove that setting completely at some point.  See the screenshots below for illustration.

update_note_sort

### Test
1. Notice sort bar is visible in note list.
2. Tap sort bar.
3. Notice popup menu with all sort options listed as described above.
4. Tap any sort option in popup menu.
5. Notice list sort order matches option selected.

### Review

### Release
`RELEASE-NOTES.txt` was updated in ac09b7e1 with:
> Added sort bar at top of note list [https://github.com/Automattic/simplenote-android/pull/1284]


AmandaRiu:
Hey @theck13 👋 Just to clarify on the sort itself...should it be case sensitive? I noticed during my testing that when changing to A-Z sorting it will sort by capital letters first and then by lowercase letters next...so if I have a note with a title "all good", it will show up after "Zoom call notes". Is that the expected functionality?

theck13:
Yeah, the sorting is based on Unicode, which treats uppercase and lowercase differently.  That wasn't changed in this pull request, but it's something we might want to update at some point for platform parity.

AmandaRiu (approved):
Tested and verified! LGTM :shipit:

pachlava:
[testing] fix/feature verified on Samsung A31 | Android 10 | Simplenote 2.17-rc-1

PR #1295, opened 2021-02-01
Update screenshots automation – 2021/02

mokagio:
### Fix

It's time to update the screenshots on the Store.
This PR contains ~~prep-work to make the automated test suite that generates the screenshots work with the latest version of the app~~ all the necessary code. _I initially planned to do it in two steps, first make the current automation work, then update it, but the second step turned out to be only a hundred lines or so and 
 didn't impact the diff size much, so I put it all in one PR to simplify the review._

### Test

- Run the tests for the "SimplenoteScreenshots" configuration and ensure the pass on both a phone and a tablet (I used Pixel 3 and Pixel C emulators)
- Run `bundle install` from the root of the project. You need to do this only once.
- Make sure you only have one Emulator running and run `bundle exec fastlane take_screeshots`. The task should succeed and open a browser window with the screenshots it took.
- Quit the Emulator and launch one for the opposite device type
- Run `bundle exec fastlane take_screeshots`. The task should succeed and open a browser window with the screenshots it took.

_If the test fails, try running them again. It worked for me the few times that happened..._ 😳 

Screen Shot 2021-02-12 at 9 32 30 pm

Screen Shot 2021-02-12 at 9 29 36 pm

### Review

### Release

### DIFF
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
@@ -258,18 +258,23 @@ public class ScreenshotTest {
     }
 
     private void loadSearchFromNotesList(String query) {
-        final int searchViewId = R.id.search_src_text;
-
         // Tap the search button in the toolbar
-        onView(withId(R.id.menu_search)).perform(click());
+        final int searchButtonId = R.id.menu_search;
+        waitForViewMatching(withId(searchButtonId), 5000);
+        onView(withId(searchButtonId)).perform(click());
+
         // Type the search query
+        final int searchViewId = R.id.search_src_text;
         onView(withId(searchViewId)).perform(typeSearchViewText(query));
         onView(withId(searchViewId)).perform(pressImeActionButton());
     }
 
     private void loadSideMenuFromNotesList() {
         // There is no R.id for the menu drawer button
-        onView(allOf(withContentDescription("Open drawer"))).perform(click());
+        final String contentDescription = "Open drawer";
+
+        waitForViewToBeDisplayed(withContentDescription(contentDescription), 2000);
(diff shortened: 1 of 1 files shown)


Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -103,17 +110,40 @@ public void screenshotTest() throws InterruptedException {
...
+        // wasn't able to find a way to achieve this – Gio
+        List<String> noteTitles = Arrays.asList(
+                NOTE_FOR_EDITOR_SHOT_TITLE,
+                "Bret Victor's Quote Collection",
+                "Back on track",
+                NOTE_FOR_INTERLINKING_SHOT_TITLE
+        );
+        for (String title:noteTitles) {
+            waitForViewMatching(allOf(withId(R.id.note_title), withText(title)), 5000);
+        }
```
mokagio, line 127: This seems to be a bit more robust that just waiting for one individual note.

jkmassel: I like this approach, but I wonder if there's a way to construct a selector that waits for all of these in parallel? That said, the test's aren't slow, so probably not worth optimizing for at this point.

jkmassel (approved):
I definitely ran into the issue with it failing occasionally, though like you said running it again seems to fix this. Perhaps we could use `fastlane` to auto-retry a couple of times in the future if there's a failure.

This is definitely a step in the right direction, so let's :shipit: and continue to iterate!

Thanks for taking care of this @mokagio!

mokagio:
Thanks for the review @jkmassel 🙇‍♂️ 

FYI @pachlava there's a few automation tests here that you may help you out. Or maybe they won't because they're a bit unstable 😞

PR #1300, opened 2021-02-01
Add Save Scroll

theck13:
### Fix
Add saving the position when scrolling a note and automatically scrolling to the saved position when opening a note.  The scroll position is saved locally per note per device.  The scroll position for a note is saved and scrolled whether the editor and markdown view is shown.  In other words, opening a note will scroll to the last scrolled position whether the note is opened showing the editor or preview.

In my testing, automatically scrolling to a position in a `WebView` didn't work unless a short delay was added after the page was finished loading.  I chose `config_mediumAnimTime`, which is 400ms, since it worked well during testing.  That choice was rather arbitrary and I'm open to other solutions suggested.

### Test
1. Tap any note in list.
2. Notice note is opened with content scrolled to the top.
3. Scroll down any amount in note.
4. Tap back arrow or navigation button.
5. Tap note from Step 1 in note list.
6. Notice note is opened with content scroll to last position.

### Review
Only one developer is required to review these changes, but anyone can perform the review.  @SylvesterWilmott, I didn't request your review or include the `[status] needs design review` label since there isn't much to see in the interface, but feel free to check these changes if you would like.

### Release
`RELEASE-NOTES.txt` was updated in d16551a2 with:
> Added scrolling to last position when opening a note [https://github.com/Automattic/simplenote-android/pull/1300]

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -580,6 +580,16 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
         return mRootView;
     }
 
+    public void removeScrollListener() {
+        mRootView.setOnScrollChangeListener(
+            new View.OnScrollChangeListener() {
+                @Override
+                public void onScrollChange(View v, int scrollX, int scrollY, int oldScrollX, int oldScrollY) {
+                }
+            }
+        );
+    }
+
     public void scrollToMatch(int location) {
         if (isAdded()) {
             // Calculate how far to scroll to bring the match into view
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -581,13 +581,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
     }
 
     public void removeScrollListener() {
-        mRootView.setOnScrollChangeListener(
-            new View.OnScrollChangeListener() {
-                @Override
-                public void onScrollChange(View v, int scrollX, int scrollY, int oldScrollX, int oldScrollY) {
-                }
-            }
-        );
+        mRootView.setOnScrollChangeListener(null);
     }
 
     public void scrollToMatch(int location) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -84,6 +84,7 @@ import java.lang.ref.WeakReference;
 import java.util.Calendar;
 import java.util.Set;
 
+import static com.automattic.simplenote.Simplenote.SCROLL_POSITION_PREFERENCES;
 import static com.automattic.simplenote.analytics.AnalyticsTracker.CATEGORY_NOTE;
 import static com.automattic.simplenote.analytics.AnalyticsTracker.Stat.EDITOR_CHECKLIST_INSERTED;
 import static com.automattic.simplenote.analytics.AnalyticsTracker.Stat.EDITOR_NOTE_CONTENT_SHARED;
@@ -159,6 +160,7 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
     private HistoryBottomSheetDialog mHistoryBottomSheet;
     private LinearLayout mError;
     private NoteMarkdownFragment mNoteMarkdownFragment;
+    private SharedPreferences mPreferences;
     private String mCss;
     private WebView mMarkdown;
     private boolean mIsPaused;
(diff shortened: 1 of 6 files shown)


nbradbury:
@theck13 I assigned myself to this since @ThomazFB is AFK.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -542,13 +563,33 @@ public void onGlobalLayout() {
                     int lineTop = layout.getLineTop(layout.getLineForOffset(matchLocation));
                     ((NestedScrollView) mRootView).smoothScrollTo(0, lineTop);
                     mShouldScrollToSearchMatch = false;
+                } else if (mNote != null && mNote.getSimperiumKey() != null) {
+                    ((NestedScrollView) mRootView).scrollTo(0, mPreferences.getInt(mNote.getSimperiumKey(), 0));
+                    mRootView.setOnScrollChangeListener(
+                        new View.OnScrollChangeListener() {
+                            @Override
+                            public void onScrollChange(View v, int scrollX, int scrollY, int oldScrollX, int oldScrollY) {
```
nbradbury, line 571: One issue here is that since the `scrollY` is being saved & restored, it won't be accurate if the note is re-opened with the device in a different orientation than when the position was saved.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -542,13 +563,33 @@ public void onGlobalLayout() {
...
+                    );
                 }
             }
         });
         setHasOptionsMenu(true);
         return mRootView;
     }
 
+    public void removeScrollListener() {
+        mRootView.setOnScrollChangeListener(
```
nbradbury: Can this be simply:

```
mRootView.setOnScrollChangeListener(null);
```

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
nbradbury: Just occurred to me that it also won't be accurate if the note is edited on another device.

nbradbury: np: the cast to `(NestedScrollView)` isn't necessary since all views support `scrollTo`.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
theck13: I thought I tried that and `setOnScrollChangeListener` didn't like `null` for some reason.  It seems to work now.  I updated the `removeScrollListener` method in 6f5da21.

nbradbury:
Do you think there should be some cleanup of the scroll positions in preferences to account for notes deleted from this device and elsewhere?

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
theck13: Yeah, this isn't a perfect solution for all situations.  The use case is when someone is editing while moving back and forth between notes in a relatively short session.  Things like remote changes and device orientation cause the scroll position to be inaccurate at times.  This is meant to try to have some feature parity between platforms even though this specific feature might be optimal on other platforms.

theck13 (commented):
> Do you think there should be some cleanup of the scroll positions in preferences to account for notes deleted from this device and elsewhere?

We could.  I don't think it would affect the app data size very much if that's the concern.  I tested 1,000 entries in the scroll position preferences file and it was 76 KB.

nbradbury:
> We could. I don't think it would affect the app data size very much if that's the concern. I tested 1,000 entries in the scroll position preferences file and it was 76 KB.

Maybe a separate PR could have a thread at startup that deletes the position pref for non-existent notes? Not a big deal since as you said it's not a lot of data, it's more of a nice to have.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
nbradbury: > Yeah, this isn't a perfect solution for all situations. The use case is when someone is editing while moving back and forth between notes in a relatively short session.

Makes total sense, I just think it could be annoying for someone who (like me) edits on multiple devices to be thrown to a random scroll position. Perhaps in a separate PR we could also store the number of characters along with the scroll position, and if different than when it was stored we default to not restoring the position. Likewise for orientation.

theck13:
> Maybe a separate PR could have a thread at startup that deletes the position pref for non-existent notes? Not a big deal since as you said it's not a lot of data, it's more of a nice to have.

Yeah, it seems like a nice-to-have enhancement and not necessarily a requirement.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
theck13: I forgot to push the changes.  The commit should be added now.

nbradbury (approved):
Looks good! :shipit:

pachlava:
[testing] fix/feature verified on Samsung A31 | Android 10 | Simplenote 2.17-rc-1

PR #1301, opened 2021-02-01
Crash Tags Edit

theck13:
### Fix
Update the method for launching the ***Edit Tags*** screen by tapping the ***Edit*** button in the ***Tags*** header of the navigation drawer to avoid an `IllegalStateException` crash on Android 6.0.x devices.

I don't think a hotfix is required after merging these changes since there is a very small number of Android 6.0.x devices using Simplenote, this crash appears to have been around for seventeen months, and it hasn't been reported before now.

### Test
Follow the steps below on a device running Android 6.0.x since the crash only occurs on devices running that version.
1. Open navigation drawer.
3. Notice app does not crash.
4. Notice ***Edit Tags*** screen is shown.

### Review

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -634,6 +634,15 @@ public class NotesActivity extends ThemedAppCompatActivity implements NoteListFr
 
         if (mTagsAdapter.getCountCustom() > 0) {
             mNavigationMenu.add(GROUP_SECONDARY, TAGS_ID, Menu.NONE, getString(R.string.tags)).setActionView(R.layout.drawer_action_edit).setEnabled(false);
+            MenuItem tagsHeader = mNavigationMenu.add(GROUP_SECONDARY, TAGS_ID, Menu.NONE, getString(R.string.tags)).setActionView(R.layout.drawer_action_edit).setEnabled(false);
+            tagsHeader.getActionView().findViewById(R.id.edit).setOnClickListener(
+                new View.OnClickListener() {
+                    @Override
+                    public void onClick(View v) {
+                        launchEditTags();
+                    }
+                }
+            );
 
             for (int i = 0; i < mTagsAdapter.getCount(); i++) {
                 String name = mTagsAdapter.getItem(i).name;
@@ -654,7 +663,7 @@ public class NotesActivity extends ThemedAppCompatActivity implements NoteListFr
         }
     }
 
-    public void launchEditTags(View view) {
+    public void launchEditTags() {
         startActivity(new Intent(NotesActivity.this, TagsActivity.class));
     }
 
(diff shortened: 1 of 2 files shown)


Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -634,6 +634,15 @@ private void updateNavigationDrawerItems() {
 
         if (mTagsAdapter.getCountCustom() > 0) {
             mNavigationMenu.add(GROUP_SECONDARY, TAGS_ID, Menu.NONE, getString(R.string.tags)).setActionView(R.layout.drawer_action_edit).setEnabled(false);
```
ParaskP7: Warning (⚠️): Consider removing this line as it is being replaced by the below one. If this line is not removed then the drawer will show two `Tags EDIT` rows, with the first one now working as expected.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -654,7 +663,7 @@ private void updateNavigationDrawerItems() {
         }
     }
 
-    public void launchEditTags(View view) {
+    public void launchEditTags() {
```
ParaskP7: Minor (🔍): Consider making this method `private` since it is no longer necessary to have this as `public`. I would even consider moving the whole `startActivity(...)` within the `onClick(...) { ... }` method above and remove this method altogether to simplify a bit.

ParaskP7 (changes requested):
👋 @theck13 !

I have reviewed and tested this PR as per the instruction.

I have found one bug should be fixed before the PR is merged to `develop`, see my warning (⚠️) comment. Please let me know when you get the change to look at it and I'll review this PR again.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
theck13: Nice catch!  That looks like a copy/paste error.  I removed the redundant line in f7144fb.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
theck13: Good idea.  I moved the `startActivity(...)` statement to the `onClick(...)` listener in 0b483fe.

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
ParaskP7: Coolio, many thanks! 🌟

Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
ParaskP7: Thanks for the refactor! 🌟

ParaskP7 (approved):

ParaskP7:
All looks good now, thanks for the update @theck13 ! 🌟

PR #1328, opened 2021-02-20
Use comma as tag delimiter

### LINKED ISSUE #1299: Tags: Use comma as a delimiter when adding tags to a note
Currently, Web/Electron and native MacOS use the comma character as a delimiter when adding tags to a note. 
Based on some internal discussion p2XJRt-2Fi-p2 Android should work the same way.

### What
When adding tags to a note inputting the comma character should submit the tag and be used as a delimiter instead of being part of the tag.

### Why
This would help be consistent across the other platforms and follow other conventions.

gayanW:
Fixes #1299

### Fix

- A comma character can now be used as a delimiter when adding tags from the tag input field within the note editor view.
- Tag name cannot contain commas.

Peek 2021-02-21 00-21

#### Use comma as a delimiter

1. Go to the note editor screen, either by tapping on an existing note or by tapping on the plus sign that creates a new note
2. Enter some text (without spaces or commas) into the tag input field
3. Then press ',' or the space key
4. Notice that the text that you entered gets added to the note as a tag
5. Repeat 2, 3, and 4

#### Disallow commas when editing tags

1. In the **'Edit Tags'** view tap on a tag in the list
2. In the **'Rename Tag'** dialog, try to enter a tag name with a comma
3. Notice the error 'Tags cannot contain commas'

#### Disallow commas when adding tags

1. In the **'Edit Tags'** view tap on the plus sign that creates a new note
2. In the **'Add Tag'** dialog, try to enter a tag name with a comma
3. Notice the error 'Tags cannot contain commas'

### Review

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
@@ -131,6 +131,12 @@ public class AddTagActivity extends AppCompatActivity implements TextWatcher {
             return;
         }
 
+        if (tag.contains(",")) {
+            mButtonPositive.setEnabled(false);
+            mTagLayout.setError(getString(R.string.tag_error_commas));
+            return;
+        }
+
         if (!TagUtils.hashTagValid(tag)) {
             mButtonPositive.setEnabled(false);
             mTagLayout.setError(getString(R.string.tag_error_length));
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
@@ -10,6 +10,7 @@ public class SearchTokenizer {
     static public final char SINGLE_QUOTE = '\'';
     static public final char GLOB = '*';
     static public final char COLON = ':';
+    static public final char COMMA = ',';
     static public final char ESCAPE = '\\';
     static public final char HYPHEN = '-';
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
@@ -10,7 +10,6 @@ public class SearchTokenizer {
     static public final char SINGLE_QUOTE = '\'';
     static public final char GLOB = '*';
     static public final char COLON = ':';
-    static public final char COMMA = ',';
     static public final char ESCAPE = '\\';
     static public final char HYPHEN = '-';
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/TagsMultiAutoCompleteTextView.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/TagsMultiAutoCompleteTextView.java
@@ -16,11 +16,8 @@ import com.automattic.simplenote.R;
 import com.automattic.simplenote.models.Tag;
 import com.simperium.client.Bucket;
 
-import static com.automattic.simplenote.utils.SearchTokenizer.COMMA;
-import static com.automattic.simplenote.utils.SearchTokenizer.EMPTY;
-import static com.automattic.simplenote.utils.SearchTokenizer.SPACE;
-
 public class TagsMultiAutoCompleteTextView extends AppCompatMultiAutoCompleteTextView implements OnItemClickListener {
+
     private Bucket<Tag> mBucketTag;
     private OnTagAddedListener mTagAddedListener;
     private TextWatcher mTextWatcher = new TextWatcher() {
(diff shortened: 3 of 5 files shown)


Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
```diff
@@ -10,6 +10,7 @@
     static public final char SINGLE_QUOTE = '\'';
     static public final char GLOB = '*';
     static public final char COLON = ':';
+    static public final char COMMA = ',';
```
dmsnell: hm. some of these make sense. colon and comma seem a little silly IMO. did you find this strange when you found them?

gayanW: COLON is used in the `SearchTokenizer#parseQuery` which returns a string that is then used to build the query condition. These constants are also used outside the SearchTokenizer. That's why I thought to define the COMMA within that same class. However, since COMMA is not used within the SearchTokenizer, I could move the COMMA variable to `TagsMultiAutoCompleteTextView`. Let me know your suggestions.

dmsnell: probably best not to change things. I wasn't so much questioning the use of constants so much as asking if these constants are pointless. kind of like `static public final int ONE = 1`

let's just leave it. I would be personally fine if we didn't create a constant for comma; it's not a name of something other than the thing itself, and any other value would break the semantic of the constant. so if you want to keep it as is that's fine but it'd also be fine if you inlined `','` instead of creating the constant.

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
```diff
@@ -131,6 +131,12 @@ public void afterTextChanged(Editable s) {
             return;
         }
 
+        if (tag.contains(",")) {
```
dmsnell, line 134: case-in-point about the constant use of `COMMA` - we're not using it here and nothing suffers from inlining this literal.

Simplenote/src/main/java/com/automattic/simplenote/utils/SearchTokenizer.java
gayanW: Maybe the intention was to facilitate refactoring by reuse. Anyway, I just inlined the character literals. as there is no obvious benefit from doing otherwise.

gayanW:
@dmsnell I updated the PR addressing your comments. Just an update in case you missed the notifications.

Simplenote/src/main/java/com/automattic/simplenote/utils/TagsMultiAutoCompleteTextView.java
```diff
@@ -83,7 +82,7 @@ public void init() {
     }
 
     public void notifyTagsChanged() {
-        String lexical = getText().toString().trim();
+        String lexical = getText().toString().replace(',', ' ').trim();
```
dmsnell, line 85: this makes me nervous, since we're making a change in a seemingly random place without any explanation as to why we're swapping spaces in for the commas.

what's the purpose here anyway? why are we doing that? is it to make sure that entered tags are split by the comma? if so, maybe we leave in the commas and handle them wherever they get split?

won't the spaces here mess up the canonical tag name anyway? `getCanonicalFromLexical` expects a valid tag name, of which one with spaces in it is not.

gayanW: This is how basically tags are being added in AppCompatMultiAutoCompleteTextView. AppCompatMultiAutoCompleteTextView is the TextView at the bottom of the Note Editor view, which is used to add tags to a note. I've inlined the method calls for brevity. The following is the original code without any of my changes.

       public void onTextChanged(CharSequence s, int start, int before, int count) {
            if (count >= 1 && s.charAt(start) == SPACE) {
                  if (TagUtils.hashTagValid(s.toString())) {
                      String lexical = getText().toString().trim();
                      String canonical = TagUtils.getCanonicalFromLexical(mBucketTag, lexical);
                      notifyTagsChanged(canonical);
            }
       }

getText() returns the text that TagsMultiAutoCompleteTextView  is displaying. Based on the condition, `s.charAt(start) == SPACE`, it would always return a text that has a leading whitespace. 

That's why trim() is being called later, so to remove that leading whitespace, like:

    String lexical = getText().toString().trim();

I only changed the following two lines of code.
```
if (count >= 1 && s.charAt(start) == SPACE) 
// TO
if (count >= 1 && (s.charAt(start) == ' ' || s.charAt(start) == ','))
```

With the refactored condition, the returned text would never contain any commas in the middle, in addition to whitespaces. So when a user enters either whitespace or a comma, it would try to save the text as a tag.

```
String lexical = getText().toString().trim();
// TO
String lexical = getText().toString().replace(',', ' ').trim();
```
trim() here trims the leading whitespace when 'space' is used as the delimiter.
And the purpose of `replace(',', ' ') is to trim the leading comma when a comma is being used as the delimiter.

The other changes made to AddTagActivity, TagDialogFragment, strings.xml are for the 'Add Tag' view.

gayanW: @dmsnell Please do let me know if it's still not clear 🙂

PR #1333, opened 2021-02-25
Add import data

theojarrus:
### Import feature
Added a function to import notes from json and plaintext

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/Importer.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Importer.java
@@ -1,0 +1,131 @@
+package com.automattic.simplenote;
+
+import android.net.Uri;
+
+import androidx.fragment.app.Fragment;
+
+import com.automattic.simplenote.models.Note;
+import com.automattic.simplenote.models.Tag;
+import com.automattic.simplenote.utils.FileUtils;
+import com.automattic.simplenote.utils.TagUtils;
+import com.simperium.client.Bucket;
+import com.simperium.client.BucketObjectNameInvalid;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.IOException;
+import java.text.ParseException;
+
+public class Importer {
+    private Bucket<Note> mNotesBucket;
+    private Bucket<Tag> mTagsBucket;
+
+    public Importer(Simplenote simplenote) {
+        mNotesBucket = simplenote.getNotesBucket();
+        mTagsBucket = simplenote.getTagsBucket();
+    }
+
+    public static void fromUri(Fragment fragment, Uri uri) throws ImportException {
+        try {
+            new Importer((Simplenote) fragment.getActivity().getApplication())
+                    .dispatchFileImport(
+                            FileUtils.getFileExtension(fragment.requireContext(), uri),
+                            FileUtils.readFile(fragment.requireContext(), uri)
+                    );
+        } catch (IOException e) {
+            throw new ImportException(FailureReason.FileError);
+        }
+    }
+
+    private void dispatchFileImport(String fileType, String content) throws ImportException {
+        switch (fileType) {
+            case "json":
+                importJsonFile(content);
+                break;
+            case "md":
+                importMarkdown(content);
+                break;
+            case "txt":
+                importPlaintext(content);
+                break;
+            default:
+                throw new ImportException(FailureReason.UnknownExportType);
+        }
+    }
+
+    private void importPlaintext(String content) {
+        addNote(Note.fromContent(content));
+    }
+
+    private void importMarkdown(String content) {
+        Note note = Note.fromContent(content);
+        note.enableMarkdown();
+
+        addNote(note);
+    }
+
+    private void addNote(Note note) {
+        // @TODO: Is there a reason to add the tags for deleted notes?
+        //        If we un-trash a note then we _do_ want the tags to exist
+        //        so for now we're _always_ creating the tags.
+        for (String tagName : note.getTags()) {
+            try {
+                TagUtils.createTagIfMissing(mTagsBucket, tagName);
+            } catch (BucketObjectNameInvalid e) {
+                // if it can't be added then remove it, we can't keep it anyway
+                note.removeTag(tagName);
+            }
+        }
+
+        mNotesBucket.add(note);
+        note.save();
+    }
+
+    private void importJsonFile(String content) throws ImportException {
+        try {
+            importJsonExport(new JSONObject(content));
+        } catch (JSONException | ParseException e) {
+            throw new ImportException(FailureReason.ParseError);
+        }
+    }
+
+    // @TODO: how should we fail on parsing?
+    //          - parse the full document and fail on _any_ failure, before importing any notes
+    //          - parse each note as we go and fail on _any_ failure, after importing some notes
+    //          - parse each note as we go, skipping any failure, importing all other notes
+    private void importJsonExport(JSONObject export) throws JSONException, ParseException {
+        JSONArray activeNotes = export.optJSONArray("activeNotes");
+        JSONArray trashedNotes = export.optJSONArray("trashedNotes");
+
+        for (int i = 0; activeNotes != null && i < activeNotes.length(); i++) {
+            addNote(Note.fromExportedJson(activeNotes.getJSONObject(i)));
+        }
+
+        for (int j = 0; trashedNotes != null && j < trashedNotes.length(); j++) {
+            Note note = Note.fromExportedJson(trashedNotes.getJSONObject(j));
+            note.setDeleted(true);
+
+            addNote(note);
+        }
+    }
+
+    public enum FailureReason {
+        FileError,
+        UnknownExportType,
+        ParseError
+    }
+
+    public static class ImportException extends Exception {
+        private FailureReason mReason;
+
+        ImportException(FailureReason reason) {
+            mReason = reason;
+        }
+
+        public FailureReason getReason() {
+            return mReason;
+        }
+    }
+}
--- a/Simplenote/src/main/java/com/automattic/simplenote/Importer.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Importer.java
@@ -17,6 +17,7 @@ import org.json.JSONObject;
 
 import java.io.IOException;
 import java.text.ParseException;
+import java.util.ArrayList;
 
 public class Importer {
     private Bucket<Note> mNotesBucket;
@@ -56,20 +57,17 @@ public class Importer {
     }
 
     private void importPlaintext(String content) {
-        addNote(Note.fromContent(content));
+        addNote(Note.fromContent(mNotesBucket, content));
     }
 
     private void importMarkdown(String content) {
-        Note note = Note.fromContent(content);
+        Note note = Note.fromContent(mNotesBucket, content);
         note.enableMarkdown();
 
         addNote(note);
     }
 
     private void addNote(Note note) {
-        // @TODO: Is there a reason to add the tags for deleted notes?
-        //        If we un-trash a note then we _do_ want the tags to exist
-        //        so for now we're _always_ creating the tags.
         for (String tagName : note.getTags()) {
             try {
                 TagUtils.createTagIfMissing(mTagsBucket, tagName);
@@ -79,7 +77,6 @@ public class Importer {
             }
         }
 
-        mNotesBucket.add(note);
         note.save();
     }
 
@@ -91,22 +88,25 @@ public class Importer {
         }
     }
 
-    // @TODO: how should we fail on parsing?
-    //          - parse the full document and fail on _any_ failure, before importing any notes
-    //          - parse each note as we go and fail on _any_ failure, after importing some notes
-    //          - parse each note as we go, skipping any failure, importing all other notes
     private void importJsonExport(JSONObject export) throws JSONException, ParseException {
         JSONArray activeNotes = export.optJSONArray("activeNotes");
         JSONArray trashedNotes = export.optJSONArray("trashedNotes");
 
+        ArrayList<Note> notesList = new ArrayList<>();
+
         for (int i = 0; activeNotes != null && i < activeNotes.length(); i++) {
-            addNote(Note.fromExportedJson(activeNotes.getJSONObject(i)));
+            Note note = Note.fromExportedJson(mNotesBucket, activeNotes.getJSONObject(i));
+            notesList.add(note);
         }
 
         for (int j = 0; trashedNotes != null && j < trashedNotes.length(); j++) {
-            Note note = Note.fromExportedJson(trashedNotes.getJSONObject(j));
+            Note note = Note.fromExportedJson(mNotesBucket, trashedNotes.getJSONObject(j));
             note.setDeleted(true);
 
+            notesList.add(note);
+        }
+
+        for (Note note : notesList) {
             addNote(note);
         }
     }
--- a/Simplenote/src/main/java/com/automattic/simplenote/Importer.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Importer.java
@@ -5,9 +5,7 @@ import android.net.Uri;
 import androidx.fragment.app.Fragment;
 
 import com.automattic.simplenote.models.Note;
-import com.automattic.simplenote.models.Tag;
 import com.automattic.simplenote.utils.FileUtils;
-import com.automattic.simplenote.utils.TagUtils;
 import com.simperium.client.Bucket;
 import com.simperium.client.BucketObjectNameInvalid;
 
@@ -16,15 +14,12 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.IOException;
-import java.text.ParseException;
 
 public class Importer {
     private Bucket<Note> mNotesBucket;
-    private Bucket<Tag> mTagsBucket;
 
     public Importer(Simplenote simplenote) {
         mNotesBucket = simplenote.getNotesBucket();
-        mTagsBucket = simplenote.getTagsBucket();
     }
 
     public static void fromUri(Fragment fragment, Uri uri) throws ImportException {
@@ -45,48 +40,35 @@ public class Importer {
                 importJsonFile(content);
                 break;
             case "md":
-                importMarkdown(content);
+                importPlaintext(content, true);
                 break;
             case "txt":
-                importPlaintext(content);
+                importPlaintext(content, false);
                 break;
             default:
                 throw new ImportException(FailureReason.UnknownExportType);
         }
     }
 
-    private void importPlaintext(String content) {
-        addNote(Note.fromContent(content));
-    }
-
-    private void importMarkdown(String content) {
-        Note note = Note.fromContent(content);
-        note.enableMarkdown();
-
-        addNote(note);
-    }
-
-    private void addNote(Note note) {
-        // @TODO: Is there a reason to add the tags for deleted notes?
-        //        If we un-trash a note then we _do_ want the tags to exist
-        //        so for now we're _always_ creating the tags.
-        for (String tagName : note.getTags()) {
-            try {
-                TagUtils.createTagIfMissing(mTagsBucket, tagName);
-            } catch (BucketObjectNameInvalid e) {
-                // if it can't be added then remove it, we can't keep it anyway
-                note.removeTag(tagName);
-            }
-        }
-
-        mNotesBucket.add(note);
+    private void importPlaintext(String content, boolean isMarkdownEnabled) {
+        Note note = mNotesBucket.newObject();
+        note.setMarkdownEnabled(isMarkdownEnabled);
+        note.setContent(content);
         note.save();
+    }
+
+    private void addNote(JSONObject properties) {
+        try {
+            mNotesBucket.insertObject(mNotesBucket.newObject().getSimperiumKey(), properties).save();
+        } catch (BucketObjectNameInvalid bucketObjectNameInvalid) {
+            // this won't happen because mNotesBucket.newObject() generates unused identifier
+        }
     }
 
     private void importJsonFile(String content) throws ImportException {
         try {
             importJsonExport(new JSONObject(content));
-        } catch (JSONException | ParseException e) {
+        } catch (JSONException e) {
             throw new ImportException(FailureReason.ParseError);
         }
     }
@@ -95,19 +77,16 @@ public class Importer {
     //          - parse the full document and fail on _any_ failure, before importing any notes
     //          - parse each note as we go and fail on _any_ failure, after importing some notes
     //          - parse each note as we go, skipping any failure, importing all other notes
-    private void importJsonExport(JSONObject export) throws JSONException, ParseException {
+    private void importJsonExport(JSONObject export) throws JSONException {
         JSONArray activeNotes = export.optJSONArray("activeNotes");
         JSONArray trashedNotes = export.optJSONArray("trashedNotes");
 
         for (int i = 0; activeNotes != null && i < activeNotes.length(); i++) {
-            addNote(Note.fromExportedJson(activeNotes.getJSONObject(i)));
+            addNote(activeNotes.getJSONObject(i));
         }
 
         for (int j = 0; trashedNotes != null && j < trashedNotes.length(); j++) {
-            Note note = Note.fromExportedJson(trashedNotes.getJSONObject(j));
-            note.setDeleted(true);
-
-            addNote(note);
+            addNote(trashedNotes.getJSONObject(j));
         }
     }
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/Importer.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Importer.java
@@ -5,7 +5,9 @@ import android.net.Uri;
 import androidx.fragment.app.Fragment;
 
 import com.automattic.simplenote.models.Note;
+import com.automattic.simplenote.models.Tag;
 import com.automattic.simplenote.utils.FileUtils;
+import com.automattic.simplenote.utils.TagUtils;
 import com.simperium.client.Bucket;
 import com.simperium.client.BucketObjectNameInvalid;
 
@@ -14,12 +16,15 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.IOException;
+import java.text.ParseException;
 
 public class Importer {
     private Bucket<Note> mNotesBucket;
+    private Bucket<Tag> mTagsBucket;
 
     public Importer(Simplenote simplenote) {
         mNotesBucket = simplenote.getNotesBucket();
+        mTagsBucket = simplenote.getTagsBucket();
     }
 
     public static void fromUri(Fragment fragment, Uri uri) throws ImportException {
@@ -40,35 +45,48 @@ public class Importer {
                 importJsonFile(content);
                 break;
             case "md":
-                importPlaintext(content, true);
+                importMarkdown(content);
                 break;
             case "txt":
-                importPlaintext(content, false);
+                importPlaintext(content);
                 break;
             default:
                 throw new ImportException(FailureReason.UnknownExportType);
         }
     }
 
-    private void importPlaintext(String content, boolean isMarkdownEnabled) {
-        Note note = mNotesBucket.newObject();
-        note.setMarkdownEnabled(isMarkdownEnabled);
-        note.setContent(content);
-        note.save();
+    private void importPlaintext(String content) {
+        addNote(Note.fromContent(content));
     }
 
-    private void addNote(JSONObject properties) {
-        try {
-            mNotesBucket.insertObject(mNotesBucket.newObject().getSimperiumKey(), properties).save();
-        } catch (BucketObjectNameInvalid bucketObjectNameInvalid) {
-            // this won't happen because mNotesBucket.newObject() generates unused identifier
+    private void importMarkdown(String content) {
+        Note note = Note.fromContent(content);
+        note.enableMarkdown();
+
+        addNote(note);
+    }
+
+    private void addNote(Note note) {
+        // @TODO: Is there a reason to add the tags for deleted notes?
+        //        If we un-trash a note then we _do_ want the tags to exist
+        //        so for now we're _always_ creating the tags.
+        for (String tagName : note.getTags()) {
+            try {
+                TagUtils.createTagIfMissing(mTagsBucket, tagName);
+            } catch (BucketObjectNameInvalid e) {
+                // if it can't be added then remove it, we can't keep it anyway
+                note.removeTag(tagName);
+            }
         }
+
+        mNotesBucket.add(note);
+        note.save();
     }
 
     private void importJsonFile(String content) throws ImportException {
         try {
             importJsonExport(new JSONObject(content));
-        } catch (JSONException e) {
+        } catch (JSONException | ParseException e) {
             throw new ImportException(FailureReason.ParseError);
         }
     }
@@ -77,16 +95,19 @@ public class Importer {
     //          - parse the full document and fail on _any_ failure, before importing any notes
     //          - parse each note as we go and fail on _any_ failure, after importing some notes
     //          - parse each note as we go, skipping any failure, importing all other notes
-    private void importJsonExport(JSONObject export) throws JSONException {
+    private void importJsonExport(JSONObject export) throws JSONException, ParseException {
         JSONArray activeNotes = export.optJSONArray("activeNotes");
         JSONArray trashedNotes = export.optJSONArray("trashedNotes");
 
         for (int i = 0; activeNotes != null && i < activeNotes.length(); i++) {
-            addNote(activeNotes.getJSONObject(i));
+            addNote(Note.fromExportedJson(activeNotes.getJSONObject(i)));
         }
 
         for (int j = 0; trashedNotes != null && j < trashedNotes.length(); j++) {
-            addNote(trashedNotes.getJSONObject(j));
+            Note note = Note.fromExportedJson(trashedNotes.getJSONObject(j));
+            note.setDeleted(true);
+
+            addNote(note);
         }
     }
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
@@ -477,6 +477,11 @@ public class PreferencesFragment extends PreferenceFragmentCompat implements Use
         }
     }
 
+    private enum TextMode {
+        WITHOUT_MARKDOWN,
+        WITH_MARKDOWN
+    }
+
     private void importData(Uri uri) {
         try {
             String exportContents = FileUtils.readFile(requireContext(), uri);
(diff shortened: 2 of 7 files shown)


Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +484,109 @@ public int compare(String text1, String text2) {
...
+                                note.setPinned(noteJson.getBoolean(key));
+                                break;
+                            case "markdown":
+                                note.setMarkdownEnabled(noteJson.getBoolean(key));
+                                break;
+                            case "publicURL":
+                                // note.setPublished(?);
+                                break;
+                        }
+                    }
```
dmsnell: the note data structure has a relatively fixed form. do we need to iterate over all the keys? I think we'll want all these fields and directly assigning them would be clearer and terser. that is, we don't really want to use a note if it doesn't contain `content` but this iteration+switch is happy to do just that.

```java
note.setContent(noteJson.has("content") ? noteJson.getString("content") : "");
note.setCreationDate(noteJson.has("creationDate") ? DateTimeUtils.getDateCalendar(noteJson.getString("creationDate")) : now());
```

could be nice to move this into a new function such as `noteFromJson(JSONValue json, List<String> trashedIds)` and then it'd be easier to test in isolation as well.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +484,109 @@ public int compare(String text1, String text2) {
...
+                    note.save();
+                }
+            } else {
```
dmsnell: can we move these different decoders into their own methods? that will leave the top-level dispatch looking clear and the separate decoders more specific

```java
String exportContents = readFile(uri);

switch (getFileExtension(uri)) {
	case "json":
		return importJsonExport(exportContents);

	case "txt":
		return importPlaintextFile(exportContents);

	case "md":
		return importPlaintextFile(exportContents);

	default:
		Toast.makeText(…"unknown file format"…);
}
```

after this change it'll be easier to isolate those exception domains vs. having this broad catch-all

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +484,109 @@ public int compare(String text1, String text2) {
...
+                Note note = noteBucket.newObject();
+                note.setContent(noteContent);
+                note.setCreationDate(Calendar.getInstance());
+                note.setModificationDate(note.getCreationDate());
```
dmsnell: I'll have to check, but I believe that we'll want to fill in default values, such as an empty list of tags and system tags

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +484,109 @@ public int compare(String text1, String text2) {
...
+        return stringBuilder.toString();
+    }
+
+    public String getFileExtension(Uri uri) {
+        String extension;
+        if (uri.getScheme().equals(ContentResolver.SCHEME_CONTENT)) {
+            final MimeTypeMap mime = MimeTypeMap.getSingleton();
+            extension = mime.getExtensionFromMimeType(requireContext().getContentResolver().getType(uri));
+        } else {
+            extension = MimeTypeMap.getFileExtensionFromUrl(Uri.fromFile(new File(uri.getPath())).toString());
```
dmsnell: does Android do anything to more accurately estimate the MIME type from a document? if that's the case, if it does more than look at the extension, would it be more helpful if we used that vs. relying on the extension? or a combination of both?

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: First wrote it like this, but then found a solution with an iterator better. I'll change it! And where should I place the `noteFromJson(JSONValue json, List<String> trashedIds)` method? Ok if I leave it in a PreferencesFragment?

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Looks good, I will change this too

Simplenote/src/main/java/com/automattic/simplenote/utils/DateTimeUtils.java
```diff
@@ -37,4 +38,12 @@ public static String getDateTextString(Context context, Calendar calendar) {
         );
         return new SimpleDateFormat(pattern, Locale.getDefault()).format(calendar.getTime());
     }
+
+    public static Calendar getDateCalendar(String json) throws ParseException {
+        String pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'";
```
dmsnell, line 43: this looks like an ISO8601 string. I see `parseOffsetISO8601` in SDK 24 which I guess means we can't use that here. I looked briefly through `androidx` and didn't see one. do you know if there's an existing ISO8601 parser available in one of the compatibility or other libraries we already use?

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
dmsnell: Good question. You might play around with it some and tell us what you think. I could see it having a good home in `Note` as well. That would bring it closer to the rest of the note semantics/models.

dmsnell:
@theo-jedi this is a great addition and I know people will love being able to import notes or accounts on their Android device. we're a bit backed up especially with Android so I apologize for our delays, which will be longer than usual. thanks for your contribution! I want to make sure we get you the proper attention on this so we can get it in, so please feel free to ping if we don't respond quickly enough.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Ok, I thought that these fields are already filled by default when creating the note object, so I figured it might look like extra code. Should I add filling tags?

dmsnell: if they already get defaults it should be fine. looks like you beat me to confirming that 😉

theojarrus:
Of course, I understand, thanks a lot, I will try to make corrections as soon as possible.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Good, thank you, I will think about it.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: 😁 I'll check again just in case

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Yes, I think I found a better solution, I'll fix it!

Simplenote/src/main/java/com/automattic/simplenote/utils/DateTimeUtils.java
theojarrus: Ok, I'll look for solutions!

theojarrus:
Now it looks much better, thank you! Unfortunately, I couldn't find better solution about ISO8601 string, but everything else seems to be fixed.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +481,101 @@ public int compare(String text1, String text2) {
...
+        for (int i = 0; i < activeNotes.length(); i++) {
+            notesArray.put(activeNotes.getJSONObject(i));
+        }
+        for (int j = 0; j < trashedNotes.length(); j++) {
+            notesArray.put(trashedNotes.getJSONObject(j));
+            trashedIds.add(trashedNotes.getJSONObject(j).getString("id"));
+        }
+        for (int k = 0; k < notesArray.length(); k++) {
+            JSONObject noteJson = notesArray.getJSONObject(k);
+            noteFromJson(noteJson, trashedIds);
```
dmsnell: it would be preferable if this function returned a note (for testing and independence etc…). if we made that change this line would become

```java
noteFromJson(noteJson, trashedIds).save();
```

theojarrus: Okay, I'll do it now

theojarrus:
Done, thank you! 😀 Anything else?

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +481,101 @@ public int compare(String text1, String text2) {
...
+        Simplenote currentApp = (Simplenote) requireActivity().getApplication();
+        Bucket<Note> noteBucket = currentApp.getNotesBucket();
+        Note note = noteBucket.newObject();
+        note.setContent(noteJson.has("content") ? noteJson.getString("content") : "");
+        note.setCreationDate(noteJson.has("creationDate") ? DateTimeUtils.getDateCalendar(noteJson.getString("creationDate")) : Calendar.getInstance());
+        note.setModificationDate(noteJson.has("lastModified") ? DateTimeUtils.getDateCalendar(noteJson.getString("lastModified")) : Calendar.getInstance());
+        note.setTags(noteJson.has("tags") ? jsonToList(noteJson.getJSONArray("tags")) : new ArrayList<String>());
+        note.setPinned(noteJson.has("pinned") && noteJson.getBoolean("pinned"));
+        note.setMarkdownEnabled(noteJson.has("markdown") && noteJson.getBoolean("markdown"));
+        note.setDeleted(noteJson.has("id") && trashedIds.contains(noteJson.getString("id")));
```
dmsnell: not going to import any chords, tracks, or photos? 😉

theojarrus: 😁 Thank you for your attention to my project! I really like Simplenote, that's why I took it as a basis. I use it every day, thank you very much for your work, I am very grateful to you!

theojarrus: So, what about merge? 😀

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +481,101 @@ public int compare(String text1, String text2) {
...
+                    importPlaintextFile(exportContents);
+                    break;
+                default:
+                    Toast.makeText(requireContext(), getString(R.string.import_unknown), Toast.LENGTH_SHORT).show();
+                    return;
+            }
+            Toast.makeText(requireContext(), getString(R.string.import_message_success), Toast.LENGTH_SHORT).show();
+        } catch (Exception e) {
+            Toast.makeText(requireContext(), getString(R.string.import_message_failure), Toast.LENGTH_SHORT).show();
+        }
```
dmsnell: this catch is uncomfortably broad. can we separate the exceptions from `readFile` from the exceptions from the importers? additionally, we have an `AppLog` that would be useful here for debugging issues that appear during the import. if you aren't sure how to add that we can do that. let's add a log when starting the import, indicating which type, and when it succeeds and fails. it would either fit as type `ACTION` or we might create a new type, `IMPORT`

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +481,101 @@ public int compare(String text1, String text2) {
...
+        Note note = noteBucket.newObject();
+        note.setContent(noteJson.has("content") ? noteJson.getString("content") : "");
+        note.setCreationDate(noteJson.has("creationDate") ? DateTimeUtils.getDateCalendar(noteJson.getString("creationDate")) : Calendar.getInstance());
+        note.setModificationDate(noteJson.has("lastModified") ? DateTimeUtils.getDateCalendar(noteJson.getString("lastModified")) : Calendar.getInstance());
+        note.setTags(noteJson.has("tags") ? jsonToList(noteJson.getJSONArray("tags")) : new ArrayList<String>());
+        note.setPinned(noteJson.has("pinned") && noteJson.getBoolean("pinned"));
+        note.setMarkdownEnabled(noteJson.has("markdown") && noteJson.getBoolean("markdown"));
+        note.setDeleted(noteJson.has("id") && trashedIds.contains(noteJson.getString("id")));
+        return note;
+    }
```
dmsnell: looking at this now it does feel like it belongs in `Note.java` more than it does here. would you be willing to move it into that module and then import it? if now we can do that before merging.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
dmsnell: this is going to have some issues as I don't know that we can or should rely on imported notes containing an id. I believe that it's not really necessary anyway, more a byproduct of the way we've chosen to build an intermediate list of notes on import.

instead we could traverse each category of notes once and avoid needing to pass in `List<String> trashedIds`. there are a couple ways we can convey the trashed status but one easy way to do it would having to think about it much is to manually alter the JSON or the output note.

```java
for (int i = 0; i < activeNotes.length(); i++) {
	noteFromJson(activeNotes.getJSONObject(i)).save();
}

for (int i = 0; i < trashedNotes.length(); i++) {
	Note note = noteFromJson(trashedNotes.getJSONObject(i));
	note.setDeleted(true);
	note.save();
}
```

(edit: corrected typo from `notesArray` to `activeNotes`/`trashedNotes`)

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +481,101 @@ public int compare(String text1, String text2) {
...
+
+    private List<String> jsonToList(JSONArray json) {
+        List<String> list = new ArrayList<>();
+        for (int i = 0; i < json.length(); i++) {
+            try {
+                list.add(json.getString(i));
+            } catch (JSONException e) {
+                e.printStackTrace();
```
dmsnell: we'll want to match our error logging with the rest of the app vs. dumping a stack trace.

in addition, it looks like the `Note` class stores its list of tags as a `JSONArray` so we could expose a new method for `setTags(JSONArray tags)` and skip the encode/decode loop.

on this point, are we able to directly instantiate a note with the existing Note class, without conversion? I doubt it works 100% but the constructor can take a JSON object. it may be the case that we can perform minimal transformation and/or slightly modify `Note.java` so we can go straight from the imported JSON object to a note.

```java
new Note(activeNotes.getObject(i)).save();
```

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +481,101 @@ public int compare(String text1, String text2) {
...
+            stringBuilder.append(line);
+            stringBuilder.append("\n");
+        }
+        inputStream.close();
+        return stringBuilder.toString();
+    }
+
+    public String getFileExtension(Uri uri) {
+        return MimeTypeMap.getSingleton().getExtensionFromMimeType(requireContext().getContentResolver().getType(uri));
+    }
```
dmsnell, line 492: these two might be great in a `FileUtils` class in the `utils` folder.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -456,6 +481,101 @@ public int compare(String text1, String text2) {
         }
     }
 
+    private void importData(Uri uri) {
+        try {
+            String exportContents = readFile(uri);
+            switch (getFileExtension(uri)) {
```
dmsnell: I tried confirming the return values for `getType(uri)` and didn't see anything helpful. did you see a list? I saw that it does chop off the first part of the mime type.

we probably want to add unit tests to cover this function for the expected file types we will support importing. can it really suggest `md` as the extension? I guess if it detects plaintext _and_ has `.md`?

this is a nice update from before though. I like that there's some leeway if someone has a file with a missing or wrong file extension.

dmsnell (commented):
Thanks for your continued work and patience. I ended up leaving more comments than I anticipated, but there were a few things I wanted to make sure we covered. We will also be performing testing on this as well as the merge. I don't know that it's totally ready for that now, but it looks fairly solid and after a few details are ironed out I think we'll be good.

theojarrus:
Sure! Thank you, understood, I'll do it!

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: I add exception handling. I think it's better to add the logs to you, because I don't know how to do it most correctly.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Yes, this is much better, but I want to note that I move it there as a static method.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Yes, I had doubts about this piece of code, so much better, thanks!

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: I changed tags adding, but I doubt about the note. In theory, we can use the following code:

`Simplenote currentApp = (Simplenote) activity.getApplication();`
`Bucket<Note> noteBucket = currentApp.getNotesBucket();`
`Note note = new Note("note - " +  UUID.randomUUID(), activeNotes.getJSONObject(i));`
`noteBucket.add(note);`
`note.save();`

It looks much better, but I doubt this is a good idea, since I am not sure how correctly the Ыimperium alogyrythms will work in this case. It seems to me that it is preferable to use `noteBucket.newObject();`

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Yes! Wanted to do it but was afraid to create new files.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: I'll leave this as it was for now, and then we'll decide!

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
dmsnell: all good. I know that fear, especially entering an open-source repo for new contributions.

we'll let you know if we want something to change, or we'll just end up changing it on merge.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
dmsnell: thanks. I might need to consult with a few people on this. please pardon the delay, but expect a few days gap.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
dmsnell: we'll take care of them, thanks

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Maybe I misunderstood, but I guess it works a little differently. It seems to me that it makes no difference to us what extension the file is, except for json, of course. All the rest will definitely be text (this is exactly what mimes are for), because I set mime types `application/json` and `text/*`, which includes txt, md, html and other, but there's no difference for us between them.

theojarrus: Here's a list (section "common examples"): en.wikipedia.org/wiki/Media_type

theojarrus:
Pushed the current changes, I'm waiting for decisions on other issues 😃

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Cool, thank you! 😀

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Sure, no problem!

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
theojarrus: Good, thank you too!

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
dmsnell: my point was that I don't see how it's possible here to ever match on `"md"`, unless Android is making a guess on that which I don't think it would?

there is a distinction that's possible, though we could handle all `text/plain` files the same way, but that distinction is whether or not we toggle markdown support in the note. if we happen to know it's a `.md` file _and_ it's plaintext then why not go ahead and import it with markdown enabled. otherwise, if it's plaintext import it just as raw text.

theojarrus: Ah, I understand, I apologize for what you had to explain. However, unfortunately, I still did not understand what the problem was with the written code, I added the method `getFileExtension(Context context, Uri uri)` that gets the file extension from uri. `getType(uri)` returns not an extension, but a mimetype, which I'm using for `getExtensionFromMimeType(mimeType)`

theojarrus:
Thank you very much for your work again!

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
dmsnell: here's the central question: can `getExtensionFromMimeType(mimeType)` ever return `"md"`?

the code is written as if it can, but I want to make sure that's the case.

theojarrus: Yes, of course I checked!

theojarrus: Well, should I add a parameter `boolean isMarkdownEnabled` to `importPlaintextFile(String exportContents)` method?

dmsnell: yeah I think so, or possibly better a string or enum `WithMarkdown | WithoutMarkdown` to avoid dangling and curious `true` values. I don't foresee any other options we'd want to pass in there such as tags or share urls so it should be good to introduce this one flag

theojarrus: Good, did! It seemed to me, that enum will be better than a string

theojarrus:
Good afternoon, is there any news? 😀

dmsnell:
@theo-jedi no news, apologies. will try and check in again next week. we haven't forgotten about you though.

theojarrus:
All good, don't worry, I understand, I'll wait, thank you!

theojarrus:
Good afternoon, is there any other way I can help? Sorry if i'm too persistent!

dmsnell:
last night I was fiddling with this; we aren't going to be able to be very fast here for a bit. doesn't bother us for you to ask; in fact that's fine just to keep it fresh, but at the moment things are way backed up.

when I was fiddling I realized it could very well be best if we create an `Importer` class or a `NoteImporter` class to contain a lot of the logic. my plan is to create a branch off of here so you can see some of the stuff I'm doing and give your own feedback, then add that aforementioned logging and finishing bits.

@SylvesterWilmott you might want to take a look at this at some point if you have some spare cycles. we're introducing new messaging around imports and failed imports, currently as toast messages.

importing the Simplenote JSON file introduces a complexity around when the file has the wrong format of data. it's possible that we might be able to import some of the notes while others fail. in that case would it be better to prevent importing any notes and say "Your file is corrupt" or better "Your file is corrupt at note blah-blah-id" or would it be better to go ahead and import all of the notes we can and say "some notes failed to import" or add a list "these notes failed to import." adding the list would complicate the UI quite a bit I think.

theojarrus:
Very well! Great code, no words, I still need to learn from you. And a good idea with a branch. Thank you for your efforts, this is very nice!

As for the `Importer.java` class, I really like it, I myself, again, was afraid to create new structural units in your project, but this is much better.

And about "todo" questions, here are a couple of my thoughts:

1. Is there a reason to add the tags for deleted notes?
In my opinion - yes. I checked that when adding a note to the trash, the tags remain (until they are removed from the tags menu), so I think we shouldn't lose this information.

2. How should we fail on parsing?
Quite a difficult question for a perfectionist. 😁 If we parse the whole file to check it for correctness before importing, then we will have to store a lot of `Note` objects in `ArrayList`, and then we need to add them into `Bucket`, I'm not sure if this is good. In addition, it seems to me that it is still worth importing notes with the correct format (I think the user would prefer to lose one or two notes rather than all). But then the question arises about the UI of the import result. Showing id of corrupted notes is a cool idea, I would like this feature, but I don't think it will be informative for non-programmer users (also, corrupted notes may not have id, and as you said, it will cause difficulties with the interface). I think a solution would be to add a counter and write something like "Imported: 10 notes, failed: 3 notes"

dmsnell:
> In my opinion - yes. I checked that when adding a note to the trash, the tags remain (until they are removed from the tags menu), so I think we shouldn't lose this information.

That makes sense to me. I would like to hear what @SylvesterWilmott has to say. Also, maybe someone (possibly @sandymcfadden) could double-check what the Electron app does, as it already has an importer working.

> then we will have to store a lot of Note objects in ArrayList, and then we need to add them into Bucket, I'm not sure if this is good

For small accounts I have no concern. If we have 5000 notes though, let's estimate the memory required to do this. I've seen export files get up to 100MB or so. Even in that case if we run 250MB in memory then we're pretty heavy, but also something that I think a phone with 2GB RAM should be fine with, possibly even a 1GB phone. I think that if someone were to run into an issue importing thousands of notes on their old phone we could direct them to the web app and do it there. That seems reasonable to me.

So I think we're probably fine with the option of moving them all into a temporary array to parse the input document, then create the notes (sorry, I know I suggested we eliminate the temporary array earlier, before we discussed the parsing failure scenario).

> And a good idea with a branch. 

I'm still not totally comfortable interacting on forked branches and while I think I could have committed to your branch directly (if you have enabled that option in your fork, or if it's the default option when forking), but I wanted to toss out the changes to you without tromping on them. If you are happy with them you can probably merge it into your branch. Something like this, though maybe what I write isn't totally correct 🤷‍♂️ 

```bash
git remote add upstream https://github.com/automattic/simplenote-android.git
git fetch remote
git merge remote add/import-on-pr-1333
```

Either this or something where you can pull in our change.

sandymcfadden:
>double-check what the Electron app does, as it already has an importer working.

I agree keeping the tags for trashed notes is the way to go. This is how we do it in Electron as well.

theojarrus:
I apologize for taking so long. Pushed your commits to this branch!

> For small accounts I have no concern... 

Fine, I understand, then I can do so!

theojarrus:
Hmm... I'm trying to test new code, It's crashing, I'll try to fix it.

I apologize, the address of line `at com.automattic.simplenote.Importer.addNote(Importer.java:81)` may be wrong, because I have already made changes with the temporary array for json import, but the error shows the method for importing the plaintext, which is not affected by the changes.

Of course, found it! Because we must use `noteBucket.newObject()` to put notes into Bucket and that's why `mNotesBucket.add(note)` doesn't work, this method sets new bucket.

I also noticed the `insertObject(String key, JSONObject properties)` and `addObject(T object)` methods in `Bucket.java`, but, unfortunately, we can't use them.

theojarrus:
Well, I finally found a better solution for how to import notes (using `insertObject(String key, JSONObject properties)` for json). I'm really sorry I didn't notice it right away, this dragged on the time. Check it out! The only problem is it crashes because of the date format, any ideas how to fix this?

Simplenote/src/main/java/com/automattic/simplenote/Importer.java
```diff
@@ -0,0 +1,110 @@
...
+    private void importJsonExport(JSONObject export) throws JSONException {
+        JSONArray activeNotes = export.optJSONArray("activeNotes");
+        JSONArray trashedNotes = export.optJSONArray("trashedNotes");
+
+        for (int i = 0; activeNotes != null && i < activeNotes.length(); i++) {
+            addNote(activeNotes.getJSONObject(i));
+        }
+
+        for (int j = 0; trashedNotes != null && j < trashedNotes.length(); j++) {
+            addNote(trashedNotes.getJSONObject(j));
```
dmsnell: if we do this don't we lose the trashed status of the note?

Simplenote/src/main/java/com/automattic/simplenote/Importer.java
```diff
@@ -0,0 +1,110 @@
...
+    // @TODO: how should we fail on parsing?
+    //          - parse the full document and fail on _any_ failure, before importing any notes
+    //          - parse each note as we go and fail on _any_ failure, after importing some notes
+    //          - parse each note as we go, skipping any failure, importing all other notes
+    private void importJsonExport(JSONObject export) throws JSONException {
+        JSONArray activeNotes = export.optJSONArray("activeNotes");
+        JSONArray trashedNotes = export.optJSONArray("trashedNotes");
+
+        for (int i = 0; activeNotes != null && i < activeNotes.length(); i++) {
+            addNote(activeNotes.getJSONObject(i));
```
dmsnell: sorry for the error in my branch. by the way, won't the error still occur here on account of the fact that the export format is different than the Simperium data model? we can't shove in invalid data…

for example, the lines loading the time formats were there to convert to the internal representations (unless it was those representations that were wrong, as I know that Simplenote stores the data as seconds-since-epoch).

I like simplifying the code but I don't think it's a good direction to skip parsing - we'll just be asking for data corruption and crashes.

Simplenote/src/main/java/com/automattic/simplenote/Importer.java
theojarrus: No, because the jsonobject has parameter "deleted", but anyway it doesn't work

Simplenote/src/main/java/com/automattic/simplenote/Importer.java
theojarrus: Yes, of course I understand... The problem is that this method simply creates an object with the given parameters. We expect the markdown field to be a boolean, but the user can replace it with a string and this will crash the application. Now I'll try to roll back the changes

theojarrus:
Something like that. It works pretty well for me, what do you think? Do we need to do something else? Maybe it's time to add an `AppLog`?

dmsnell:
Yeah @theo-jedi the code is looking good. I wish we didn't have to send the notes bucket into `Note.fromExportedJson` and the other, but I think it's something tolerable for now.

At this point most of the finishing work will likely be on our side. I'll try and figure out best how to do that since I haven't done it before, whether that's us merging onto this branch or merging this into one of ours.

theojarrus:
Great, thanks for the experience, it was a pleasure working with you! 😊

I suggest to send note object into `Note.fromContent` and `Note.fromExportedJson` instead of bucket, and then change call in `Importer` to `Note.fromContent(mNotesBucket.newObject(), content)`. I can do this, I just thought that it is better not to make new edits in case you started the final preparation.

Hope everything goes well, thank you so much again, keep in touch.

dmsnell:
> I can do this, I just thought that it is better not to make new edits in case you started the final preparation.

that's something I considered as well, but then it seemed to be even more confusing of an interface than passing in the bucket itself. I think this butts up against the way this app interacts with the Simperium library and so I don't think we need to get hung up here in this PR on it. thanks for being open to change and feedback though!

> Great, thanks for the experience, it was a pleasure working with you! 😊

Same back to you!

theojarrus:
Okay, got it! Thank you again!

SylvesterWilmott:
This all looks great, thanks @theo-jedi. UI is perfectly done.

>We're introducing new messaging around imports and failed imports, currently as toast messages.

If I can make a couple of suggestions, I think we could make the messages more user-friendly:

- `Data loaded` > `Import successful`

Our general error copy convention is "Cannot [action]":

- `Parse error` > `Cannot parse file`
- `File error` > `Cannot read file`

Hopefully the above accurately represent what the error actually is, if not feel free to suggest something. Could you also clarify what would trigger `Date error`?

>In my opinion - yes. I checked that when adding a note to the trash, the tags remain (until they are removed from the tags menu), so I think we shouldn't lose this information.

Echoing @sandymcfadden. Tags should be kept for trashed tags.

theojarrus:
Thank you so much! 😊

> If I can make a couple of suggestions, I think we could make the messages more user-friendly. Hopefully the above accurately represent what the error actually is, if not feel free to suggest something

Yes, of course, I'll do it now!

> Could you also clarify what would trigger `Date error`.

Oh sorry, I forgot to remove this line, it was previously used instead of `Parse error`, now it is no longer needed.

> Echoing @sandymcfadden. Tags should be kept for trashed tags.

Yes, @dmsnell  and I came to the conclusion that the tags should be keeped for trashed notes, that's how it is now

theojarrus:
Good evening, let me know when you're ready to merge. If I need to make any other final edits (or any actions with git), I'm here 😉

theojarrus:
Good afternoon, have any news?

dmsnell:
> good afternoon, have any news?

unfortunately not. I'm doing too many things simultaneously. this is a great contribution; it's all on us to get it moving faster. thanks for your patience.

SylvesterWilmott:
Green light from me design-wise. Thanks for the updates @theo-jedi

theojarrus:
> unfortunately not. I'm doing too many things simultaneously. this is a great contribution; it's all on us to get it moving faster. thanks for your patience.

Sure, all good! I just got worried because there was no response to the latest changes. Sorry for the unnecessary worry, I will wait, good luck with your work, and thank you!

theojarrus:
> Green light from me design-wise. Thanks for the updates @theo-jedi

Great. Thank you for your attention to this code! 😊

dmsnell:
Status update for you @theo-jedi:
 - this is really good work and we want to merge it
 - our development is paused temporarily as we shuffle some things on the Android app
 - we won't forget this PR but we shouldn't expect any movement soon
 - when we do merge it we need to add the `AppLog` entries (that's on us at Automattic)
 - thank you for your patience and diligence; we're going to stress that a bit more but will update you when we resume work on this

theojarrus:
Hey, good day @dmsnell! What about working on the merge of this pull request? Is it still not available?

danilo04 (approved):
Hi @theo-jedi, I am going to merge this PR. Thank you for the implementation of this feature. I will be adding and checking a few details to make sure the feature is complete before release. I have created issue #1417 to keep track of the changes that need to be added.

theojarrus:
Great, thank you so much @danilo04! I was really glad to help, very nice! 😊

PR #1337, opened 2021-03-07
Refactor Tag#renameTo to support renaming unassigned tags

### LINKED ISSUE #1319: Unassigned tags can't be renamed
### Expected
Any tag can be renamed

### Observed
If tag is not used in notes, it can't be renamed

### Reproduced
1. Add a new unique tag to any note
2. Make sure tag was created
3. Remove tag from the note
4. Open tags editor and try renaming the recently created tag ~> this doesn't work

Device Make: Samsung
Device Model: A31
Android Version: 10
Simplenote Version: 2.17-rc-1

gayanW:
### Fix

Fixes https://github.com/Automattic/simplenote-android/issues/1319

This refactors `Tag#renameTo` to support renaming unassigned tags, as the current implementation does not work for unassigned tags.

### Test

1. Open the navigation drawer.
2. Tap the **Edit** button in the header named '**Tags**' to open the tag editor.
2. Add a new tag.
3. Rename it.
4. Note that it is being renamed.
5. Rename it to a lexical variation.
6. Note that it is being renamed to the lexical variation.

### Review

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
@@ -11,8 +11,6 @@ import com.simperium.client.Query.ComparisonType;
 
 import org.json.JSONObject;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Locale;
 
 import static com.automattic.simplenote.models.Note.TAGS_PROPERTY;
@@ -78,55 +76,38 @@ public class Tag extends BucketObject {
     }
 
     public void renameTo(String tagOld, String tagNew, int index, Bucket<Note> notesBucket) throws BucketObjectNameInvalid {
-        // When old tag ID is equal to new tag hash, tag is being renamed to lexical variation.
-        boolean isOldIdEqualToNewHash = getSimperiumKey().equals(TagUtils.hashTag(tagNew));
+        if (tagOld.equals(tagNew)) return;
+
+        renameTagInNotes(tagOld, tagNew, notesBucket);
+
         //noinspection unchecked
         Bucket<Tag> tagsBucket = (Bucket<Tag>) getBucket();
-        // Get all notes with old tag to update.
-        ObjectCursor<Note> notes = findNotes(notesBucket, tagOld);
 
-        while (notes.moveToNext()) {
-            Note note = notes.getObject();
-            List<String> tagsNew = new ArrayList<>();
-            List<String> tagsHash = new ArrayList<>();
+        boolean isOldIdEqualToNewHash = getSimperiumKey().equals(TagUtils.hashTag(tagNew));
+        if (isOldIdEqualToNewHash) {
+            setName(tagNew);
+            save();
 
-            // Create lists of note's tags excluding old tag.
-            for (String tag : note.getTags()) {
-                if (!tag.equals(tagOld)) {
-                    tagsNew.add(tag);
-                    tagsHash.add(TagUtils.hashTag(tag));
-                }
+        } else {
+            // Create new tag if canonical tag doesn't already exist.
+            if (!TagUtils.hasCanonicalOfLexical(tagsBucket, tagNew)) {
+                TagUtils.createTag(tagsBucket, tagNew, index);
             }
 
-            // Add lexical tag to note.  Update this tag's name and save it.
-            if (isOldIdEqualToNewHash) {
-                tagsNew.add(tagNew);
+            delete();
+        }
+    }
 
-                if (!getName().equals(tagNew)) {
-                    setName(tagNew);
-                    save();
-                }
-            // Add new canonical tag to note and create new tag.  Delete this tag.
-            } else {
-                // Add new tag if note doesn't already have same hashed tag.
-                if (!tagsHash.contains(TagUtils.hashTag(tagNew))) {
-                    tagsNew.add(TagUtils.getCanonicalFromLexical(tagsBucket, tagNew));
-                }
+    private void renameTagInNotes(String tagOld, String tagNew, Bucket<Note> notesBucket) {
+        ObjectCursor<Note> notesWithOldTag = findNotes(notesBucket, tagOld);
 
-                // Create new tag if canonical tag doesn't already exist.
-                if (!TagUtils.hasCanonicalOfLexical(tagsBucket, tagNew)) {
-                    TagUtils.createTag(tagsBucket, tagNew, index);
-                }
-
-                delete();
-            }
-
-            // Add new tags to note and save it.
-            note.setTags(tagsNew);
+        while (notesWithOldTag.moveToNext()) {
+            Note note = notesWithOldTag.getObject();
+            TagUtils.renameTagInNote(note, tagOld, tagNew);
             note.save();
         }
 
-        notes.close();
+        notesWithOldTag.close();
     }
 
     public ObjectCursor<Note> findNotes(Bucket<Note> notesBucket, String name) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
@@ -2,6 +2,7 @@ package com.automattic.simplenote.utils;
 
 import android.util.Log;
 
+import com.automattic.simplenote.models.Note;
 import com.automattic.simplenote.models.Tag;
 import com.simperium.client.Bucket;
 import com.simperium.client.BucketObjectMissingException;
@@ -11,6 +12,9 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.text.Normalizer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.ListIterator;
 import java.util.Locale;
 
 public class TagUtils {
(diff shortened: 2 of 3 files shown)


Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
```diff
@@ -79,54 +79,59 @@ public void setIndex(Integer tagIndex) {
 
     public void renameTo(String tagOld, String tagNew, int index, Bucket<Note> notesBucket) throws BucketObjectNameInvalid {
         // When old tag ID is equal to new tag hash, tag is being renamed to lexical variation.
-        boolean isOldIdEqualToNewHash = getSimperiumKey().equals(TagUtils.hashTag(tagNew));
+        boolean isRenamingToLexicalTag = getSimperiumKey().equals(TagUtils.hashTag(tagNew));
```
dmsnell: this new name is confusing me. since it may not be obvious, here are how we use the terms:

 - a tag entity exists in the tag bucket, operated on within this `Tag` class
 - tag entities are keyed/referenced by their `entity id`, or `simperiumKey`, which is the `tagHash` - an encoded version of all the possible `lexical variations` of the same tag name
 - the tag `name` is one of multiple possible `lexical variations` of the tag name which result from different ways we can represent a given name in Unicode.

this boolean flag is determining whether or not the new name collapses to the same tag hash as already exists or to a new one. if it's the same tag hash then we can limit our update to the `name` property of the tag. if it collapses to a different hash then we have to take a more complicated approach, creating a new tag with the new hash/name, updating all of the notes which referenced the old tag, and then deleting the old tag.

 - `bookmarks` -> `Bookmarks` is a different lexical variation of the same tag hash `bookmarks` so we can update the `name` property and be done.
 - `bookmarks` -> `favorites` implies updating every note that references `bookmarks` and ensuring that a tag with the `tagHash = "favorites"` exists

this is long-winded but I hope it clarifies

Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
```diff
@@ -79,54 +79,59 @@ public void setIndex(Integer tagIndex) {
...
-                    tagsNew.add(TagUtils.getCanonicalFromLexical(tagsBucket, tagNew));
-                }
-
-                // Create new tag if canonical tag doesn't already exist.
-                if (!TagUtils.hasCanonicalOfLexical(tagsBucket, tagNew)) {
-                    TagUtils.createTag(tagsBucket, tagNew, index);
-                }
-
-                delete();
+            delete();
```
dmsnell, line 97: let's keep the delete operation at the end of the sequence. we don't want to inadvertently trigger an unwanted sequence of updates on another device when it sees a tag disappear, before we've updated the notes referencing the tag

Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
```diff
@@ -79,54 +79,59 @@ public void setIndex(Integer tagIndex) {
...
-            note.setTags(tagsNew);
-            note.save();
+        if (isRenamingToLexicalTag) {
+            tagsNew.add(tagNew);
+
+        } else {
+            // Add new tag if note doesn't already have a tag with the same hash.
+            if (!tagsHash.contains(TagUtils.hashTag(tagNew))) {
+                tagsNew.add(TagUtils.getCanonicalFromLexical(tagsBucket, tagNew));
+            }
```
dmsnell: I don't think we have to introduce as much complexity in the note side; the logic should be straightforward for all updates regardless of whether we're changing the `tagHash` in the rename.

dmsnell (commented):
thanks for the contribution @gayanW! I left a few notes and I think we can clarify some of the logic in this PR. have a look over it and let me know your thoughts

Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
gayanW: Thanks for the detailed reply. It clarifies a lot of things for me.

I'll rename the variable back to `isOldIdEqualToNewHash`. Let me clarify this once again. Please correct me if I'm wrong.

- For a particular tag, `simperiumKey`, `tagHash`, and `entity id` are the same.
- When the hash of the new tag equals to the hash of the old tag, the new tag is either a lexical variation or equal to the old tag.

The reason I thought of renaming that variable is that I saw the following code line in the `TagDialogFragment` class:

https://github.com/Automattic/simplenote-android/blob/edb554cb02993ebf87a7258de92d00c9ab06a07b/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java#L101

I'll look into the code once again and will get back to you addressing the other comments.

dmsnell: > For a particular tag, simperiumKey, tagHash, and entity id are the same.

`simperiumKey` and `entityId` are _always_ the same because they are two names referring to the same value. we started using `entityId` to help disambiguate this very thing across the different platforms but that was after the Android library was written.

`tagHash` is a relatively new term and it _should_ equal the entity id but in many cases it _won't_ because of old bugs in this and the other apps. it's possible to find all sorts of different entity ids "in the wild" in existing accounts.

I should have defined "lexical variations" a bit better: there are rules collapsing a tag name into the tag hash so that we don't end up with duplicate tags when only one was meant:
 - letter casing in the tag name should be irrelevant: `Book` and `book` refer to the same tag
 - Unicode variants should be irrelevant: `ü` (U+FC) and `ü` (U+75 U+0308) refer to the same tag

The string value stored in the `tag.name` property is what we call the "canonical name" or the "canonical variation" because that's what we intend to display around the app as the tag name. The tag hash is generally not a displayable value because it gets URL-encoded.

---

What's really going on here is a question about re-using an existing tag entity with `entityId == tagHash` or if the rename caused it to map to a different tag hash entirely. Tags are a bit wonky in the database because notes refer to them by their name (or some lexical variation of their name) while they are referenced by their hash, which is a form of their name.

gayanW:
Hi @dmsnell. I updated the PR. I moved the `renameTagInNote` method out of the `Tag` class, as it's not bound to a particular Tag, nor it references any state. 

The iterator here is used to replace the old tag with the new tag while maintaining the order of the tags, instead of appending the new tag to the end of the tags list.

https://github.com/Automattic/simplenote-android/blob/ae66fa3092b8656e09ab2576c0abb02cf9870fa7/Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java#L191-L205

Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
```diff
@@ -165,4 +169,42 @@ private static String replaceEncoded(String encoded) {
...
+                if (isOldHashEqualNew || !tagHashList.contains(tagNewHash)) {
+                    // Add new tag
+                    iterator.add(tagNew);
+                }
+            }
+
+        }
+
+        note.setTags(tags);
+    }
```
dmsnell, line 208: I'm still a bit confused with this function, specifically the multiple iterations, which I don't think are necessary. here's another formulation that I think is supposed to do the same thing, though it covers one more case, which is where we already had a tag in a note that matches the new name of the renamed tag. this situation represents certain data corruption, but it definitely exists in practice.

```erlang
public static void renameTagInNote(Note note, String tagOld, String tagNew) {
    List<String> newTags = new ArrayList<>();
    String oldHash = hashTag(tagOld);
    String newHash = hashTag(tagNew);

    for (String tag : note.getTags()) {
        String hash = hashTag(tag);

        if (hash.equals(newHash)) {
            continue;
        }

        if (hash.equals(oldHash)) {
            newTags.add(tagNew);
            continue;
        }

        newTags.add(tag);
    }

    note.setTags(newTags);
}
```

Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java
```diff
@@ -78,55 +76,38 @@ public void setIndex(Integer tagIndex) {
     }
 
     public void renameTo(String tagOld, String tagNew, int index, Bucket<Note> notesBucket) throws BucketObjectNameInvalid {
-        // When old tag ID is equal to new tag hash, tag is being renamed to lexical variation.
-        boolean isOldIdEqualToNewHash = getSimperiumKey().equals(TagUtils.hashTag(tagNew));
+        if (tagOld.equals(tagNew)) return;
+
+        renameTagInNotes(tagOld, tagNew, notesBucket);
```
dmsnell, line 81: how about we let this whole process exist inside of `TagUtils` - I like how you moved part of it there already, and I'd rather not coupled the notes bucket into the tag bucket if possible, though I acknowledge the inherent challenge presented with the data model.

```java
TagUtils.renameTag(String oldTagName, String newTagName);
```

if we do that then we can let `TagUtils.renameTag()` call `Tag.rename()` and also iterate over all the notes. this `Tag.renameTo` could be a bit simpler. hmm. thinking…

dmsnell:
@gayanW what would you say is the core reason that we couldn't rename unassigned tags? before we get too far here I wonder if we should step back and ask that prime question again. What is the cause for the bug and what is the solution? It feels like we're doing more than resolve that one thing here, which might actually be fine, though I want to make sure the fix is as clear as the problem.

Thanks for all your work on this! I love it!

Simplenote/src/main/java/com/automattic/simplenote/utils/TagUtils.java
gayanW: I think the suggested implementation doesn't support renaming a tag to a lexical variation. But the old implementation does.

We'd also have to support the following two scenarios since the old implementation also supports those.

| Scenario                                | Tags              | Rename       | Expected Result    |
| -----------------------                  | ----------------- | ---------------- | ----------    |
| Rename to lexical                 | one               | one  -->  ONE    | ONE        |
| Rename to existing in a lexical form   | one two         | one  -->  TWO | two          |


    @Test
    public void renameToLexical() {
        Note note = new Note("key");
        note.setTags(tagList("one"));

        TagUtils.renameTagInNote(note, "one", "ONE");
        assertThat(note.getTags(), is(tagList("ONE")));
    }

    @Test
    public void renameToExistingLexical() {
        Note note = new Note("key");
        note.setTags(tagList("one", "two"));

        TagUtils.renameTagInNote(note, "one", "TWO");
        assertThat(note.getTags(), is(tagList("two")));
    }

I actually just refactor and restructure the logic that is used to rename a tag in a note. `TagUtilsTest#testRenameTagInNote` test is being added to make sure that the final outcome is as same as the old code. The reason I used an iterator is to maintain the order of the tags.

For instance previously if we rename a tag, the renamed tag is being appended to the end of the tag string, like: 
 
| Tags              | Rename       |  Result    |
 | ----------------- | ---------------- | ----------    |
| One two three | One  -->  one | two three one      |

Please have a look at the unit test `TagUtilsTest#testRenameTagInNote` and let me know if `TagUtils#renameTagInNote` method can be further simplified.

gayanW:
https://github.com/Automattic/simplenote-android/blob/25c4eccdd4f153ad6a5dda7b4bb7d8de427f337b/Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java#L80-L130

@dmsnell The above is the original code. Let me try brief what it does:

- It first retrieves all the note objects containing the old tag.
- Then it Iterates over the retrieved notes.
      - Within the loop, it modifies the TAGS_PROPERTY of each note (This modifies the Note#TAGS_PROPERTY, and not the Tag itself)

https://github.com/Automattic/simplenote-android/blob/25c4eccdd4f153ad6a5dda7b4bb7d8de427f337b/Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java#L86-L88

All renaming operations are performed within the loop. So the problem is if there aren't any notes containing the `tagOld`, no operation is performed at all. So to fix it, the Tag itself should be renamed despite it being assigned to any note or not.

Therefore I had to refactor the method as follows:

https://github.com/Automattic/simplenote-android/blob/ae66fa3092b8656e09ab2576c0abb02cf9870fa7/Simplenote/src/main/java/com/automattic/simplenote/models/Tag.java#L78-L99

belcherj:
Thanks for bearing with us and that this has taken so long. We are currently quite busy working on some projects that have caused the delay. Please give us a little more time.

PR #1340, opened 2021-03-10
Sign Up Flow Update

ParaskP7:
### Details

This PR updates the `Sing Up` flow.

#### Sign Up

https://user-images.githubusercontent.com/9729923/110635503-d5bd1580-81b3-11eb-99f8-2fd71a8f74ba.mp4

#### Magic Link

https://user-images.githubusercontent.com/9729923/110635721-1583fd00-81b4-11eb-8a49-c5bb925ede19.mp4

#### Sign Up
1. Start `Simplenote` from scratch.
2. Tap on `SIGN UP`.
3. Enter your `Email` and click the `SIGN UP` button.
3. Verify that the `Confirmation` screen is shown.
4. Verify that the email your entered is shown in bold.
5. Verify that the `support@simplenote.com` link redirects to opening your email client, ready to send an email to Simplenote support.

#### Magic Link
1. Wait for the email, with the magic link, to arrive.
2. Click on the magic link.
3. Verify that you are redirected to the browser to complete the `Sign Up` process.
4. Verify that your `Email` is pre-filled.
5. Enter your `Password` and click the `Sign up` button.
6. Verify that you are redirected to the `Simplenote` app and the `Sign Up` process is complete.
7. Verify that the `Welcome to Simplenote` note is present.
8. Go to `Setting` and verify that your account is the expected one.

### Review

Only one developer and one designer are required to review these changes, anyone can perform the review, but it is recommended that as many developers are possible are involved so that the overall solution ends up being the intended one, without unforeseeable side effects.

### Release

These changes might need release notes.


ParaskP7:
👋 @SylvesterWilmott !

Me and @hichamboushaba implemented an extra alert dialog for the case where the user tries to use a magic link that points to a different email, while they are already logged in with another email. This is what we end up doing:

image

Please let us know what would be the most appropriate alert dialog title, message and actions.

🙏 

PS: @eshurakov , maybe this is something you want to do on your side on iOS as well?

sandymcfadden:
Thanks @ParaskP7 I'm going to go through and test this now. For the already logged-in modal, we did add one on the web/electron app. You can see a screenshot of how it looks in [this issue](https://github.com/Automattic/simplenote-electron/issues/2734).

ParaskP7:
👋 @dmsnell !
👋 @eshurakov !

While testing the solution we were randomly getting the `Review Your Account` screen just after completing the `Sign Up` flow. Also, we got that a few time during a normal `Sign Up` flow, see below:



In this screenshot above you will notice the `null` email, which randomly happens during the `Sign In` flow. This is not affected by any of our changes. A similar screenshot to above, but this time with the `signed up` email (instead of `null`), also randomly happens during the `Sign Up` flow.

@hichamboushaba and me identifies that the root cause of that is because of this little snippet of come within the `Simplenote` application class (line `224`).

```
        try {
            account = mAccountBucket.get(KEY_EMAIL_VERIFICATION);
        } catch (BucketObjectMissingException bucketObjectMissingException) {
            // Show review account when entity does not exist.
            showReviewAccountOrVerifyEmail(activity, false);
            return;
        }
```

The `checkReviewAccountOrVerifyEmail(...)` is called in two places, in the above snippet it's called if the email verification bucket wasn't found, and below it, it's called if the email is not verified actually. We think that the first place is problematic, because it may occur before syncing is done, and it's what caused our issue. Shouldn't we ignore the exception, and don't show the review screen unless we are sure it's not verified?

In seems that this is a `Simperium` bucket sync problem, but we are not sure. Can you maybe provide us with more info on that one and maybe help us identify an appropriate solution to this problem?

🙏 

PS: It seems that this problem happens in `develop` too.

ParaskP7:
👋 @sandymcfadden !

> Thanks @ParaskP7 I'm going to go through and test this now.

🌟 

> For the already logged-in modal, we did add one on the web/electron app. You can see a screenshot of how it looks in this issue.

Many thanks for the pointers on that, much appreciated! 🙏

eshurakov:
> PS: @eshurakov , maybe this is something you want to do on your side on iOS as well?

We thought about it and decided to silently ignore it because it's an unlikely edge case. Same happens on iOS when handling sign in with Wordpress. If you're logged in, it is ignored.

ParaskP7:
👋 @eshurakov !

> We thought about it and decided to silently ignore it because it's an unlikely edge case. Same happens on iOS when handling sign in with Wordpress. If you're logged in, it is ignored.

Many thanks for letting us know. 🙏

SylvesterWilmott:
Looking good!

A few notes. On the screen where we show the message to say "check your email":

1. It looks as though the content is centered in the viewport (ie between the app bar and the system bar) as opposed to being centered in the screen which makes it look visually off-center. Can we center it so it doesn't take the height of the app bars into consideration?
2. We should avoid orphans in the text for example the word "help" at the end.
3. The mail icon can inherit the colour from the icons we have on the "review your account" screen for both light and dark modes.

sandymcfadden:
I've gone through and tested this a number of times and everything seems to work as expected except for what you mentioned in the [comment above](https://github.com/Automattic/simplenote-android/pull/1340#issuecomment-795447284) with the email confirmation. The accounts should indeed be confirmed automatically when going through this flow.

ParaskP7:
👋 @SylvesterWilmott !

> A few notes. On the screen where we show the message to say "check your email":

🙏 

> 1. It looks as though the content is centered in the viewport (ie between the app bar and the system bar) as opposed to being centered in the screen which makes it look visually off-center. Can we center it so it doesn't take the height of the app bars into consideration?

You are right, on it! 🌟 

> 2. We should avoid orphans in the text for example the word "help" at the end.

Will investigate and get back to you on that. 🤔 

> 3. The mail icon can inherit the colour from the icons we have on the "review your account" screen for both light and dark modes.

You are right about the `mail` icon color, on it! 🌟 

About the `dark` modes, I believe that the authentication and credential screens on Android don't have a `dark` mode in general. Should we do anything about that?

ParaskP7:
👋 @sandymcfadden !

> I've gone through and tested this a number of times and everything seems to work as expected except for what you mentioned in the comment above with the email confirmation. The accounts should indeed be confirmed automatically when going through this flow.

Many, many thanks! 🙏 

The `...except for what you mentioned in the comment above with the email confirmation.` you mentioned there, does it mean that you indeed hit this problem as well?

ParaskP7:
👋 @SylvesterWilmott !

Quick update below:

> 1. It looks as though the content is centered in the viewport (ie between the app bar and the system bar) as opposed to being centered in the screen which makes it look visually off-center. Can we center it so it doesn't take the height of the app bars into consideration?

This is now done: 330b63c4bda84de7fe2c068edf32b3f016f9c3c6

> 2. We should avoid orphans in the text for example the word "help" at the end.

Warning (⚠️): This is not trivial and will take a lot of effort to implement. Also, depending on devices. Me and @hichamboushaba suggest that unless this is a blocker, we should just keep that as is for now. Maybe create an separate issue to improve on that later on. Wdyt @SylvesterWilmott ?

> 3. The mail icon can inherit the colour from the icons we have on the "review your account" screen for both light and dark modes.

This is now done: 47ebe6308b0fbb47350abc1ab39a3e67ea0e1e50

About the `dark` modes, I believe that the authentication and credential screens on Android don't have a `dark` mode in general. Should we do anything about that?

> Me and @hichamboushaba implemented an extra alert dialog for the case where the user tries to use a magic link that points to a different email, while they are already logged in with another email.

This is our latest update on that: 4d05a3cf793db18dce0315259e0eebda286f3d2d

image

dcalhoun:
I tested this flow on a Samsung Galaxy S20 FE 5G. The flow succeeded as outlined and expected. 👍🏻 

The only issue I encountered was that tapping the link in the email would not open my browser. I also tried long-pressing the link and selecting "Open in Browser" from the context menu. That did not work either. I was forced to copy the URL and paste it into my browser. If no one else experiences this, it likely is just a configuration specific to my testing device (e.g. Personal vs Work profiles).

Simplenote/src/main/res/layout/fragment_confirmation.xml
```diff
@@ -0,0 +1,43 @@
...
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:gravity="center"
+    android:orientation="vertical">
+
+    <androidx.appcompat.widget.AppCompatImageView
+        android:layout_width="72dp"
+        android:layout_height="72dp"
+        android:layout_gravity="center"
+        android:layout_marginBottom="16dp"
```
AmandaRiu: np: Hard-coded dimension values. There is a `dimens.xml` file with the standard dimensions used across the app. Using those would ensure the padding and margins transfer nicely to larger devices like tablets.

Simplenote/src/main/res/layout/fragment_confirmation.xml
```diff
@@ -0,0 +1,43 @@
...
+        android:layout_height="wrap_content"
+        android:layout_marginStart="24dp"
+        android:layout_marginEnd="24dp"
+        android:layout_marginBottom="20dp"
+        android:gravity="center"
+        android:textColor="@color/gray_80"
+        android:textSize="16sp"
+        tools:text="We've sent an email to example@email.com. Please check your inbox and follow the instructions." />
+
+    <TextView
```
AmandaRiu: It looks like there are a some custom components used throughout the simplenote app that are already styled for a consistent look. Maybe something to consider. Example: `com.automattic.simplenote.widgets.RobotoMediumTextView`

AmandaRiu (commented):
Hey @ParaskP7 👋 Review complete. Code looks good - just left a couple minor comments for review and consideration. Also noticed the following during testing:

1. I _did_ run into the **null** email page but I see it's already been explored. 
2. I was able to signup with an email that was already connected to Simplenote and didn't receive an email. I see that's a support message - just thought it was strange that we can't verify that up front. 

None of the above are blockers and it looks like they are both known, just wanted to note it here.

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignupFragment.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    private HttpUrl buildUrl() {
+        return new HttpUrl.Builder()
+            .scheme(HTTP_SCHEME)
+            .host(HTTP_HOST)
+            .addPathSegment(SIMPLENOTE_ACCOUNT_PATH)
+            .addPathSegment(SIMPLENOTE_REQUEST_SIGNUP_PATH)
+            .build();
+    }
```
dmsnell, line 165: this looks like it would be hard to search for in the source code (since the URL we want doesn't exist anywhere as a whole piece) and it's also confusing what it's really hitting, having to jump around in the file to understand. what are your thoughts on making this less abstract and using `HttpUrl.parse()` since it's all static anyway?

```java
HttpUrl.parse("https://app.simplenote.com/account/request-signup");
```

Simplenote/src/main/java/com/automattic/simplenote/utils/AuthUtils.java
```diff
@@ -55,4 +63,30 @@ public static void logOut(Simplenote application) {
 
         WidgetUtils.updateNoteWidgets(application);
     }
+
+    public static boolean hasDifferentEmail(Simplenote application, Uri uri) {
+        User user = application.getSimperium().getUser();
+        return user.getStatus().equals(User.Status.AUTHORIZED) &&
+            !user.getEmail().equals(extractEmailFromMagicLink(uri));
+    }
```
dmsnell: this function seems to highly couple the entire application as well as the auth flow. is it necessary? if we say it _is_, maybe it has a happier home in the application itself, or somewhere closer to the user (I realize we're getting it directly from Simperium, and while it would be nice if we could update the Simperium lib a quick fix here is probably necessary in the meantime)

```java
// Simperium.java

public boolean loggedInAs(String email) {
	User user = mSimperium.getUser();

	if (user.getStatus().equals(User.Status.AUTHORIZED)) {
		return false;
	}

	return user.getEmail().toLowerCase(US).equals(email.toLowerCase(US));
}
```

Note that I've supplied the `US` locale (which I hope is `en-US` since that's the real goal) so that we don't end up changing the email comparison based on the device language, as `equalsIgnoreCase()` would.

```java
if ((Simplenote) getApplication.loggedInAs(AuthUtils.extractEmailFromMagicLink(uri))) {
	…
}
```

In this construction we've removed the coupling and also made it easier to test or document the behavior of "is this the email of the logged-in user?"

Simplenote/src/main/java/com/automattic/simplenote/utils/AuthUtils.java
```diff
@@ -55,4 +63,30 @@ public static void logOut(Simplenote application) {
...
+
+    public static void magicLinkLogin(Simplenote application, Uri uri) {
+        String userEmail = extractEmailFromMagicLink(uri);
+        String spToken = uri.getQueryParameter("token");
+
+        User user = application.getSimperium().getUser();
+        user.setAccessToken(spToken);
+        user.setEmail(userEmail);
+        user.setStatus(User.Status.AUTHORIZED);
```
dmsnell: on a similar note would it make sense to pass control back to the main application? I know all this was previously handled by the Simperium library itself so it all surely is a bit awkward.

```java
(Simplenote) getApplication.login(userEmail, spToken);
```

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignupFragment.java
```diff
@@ -0,0 +1,264 @@
...
+            .url(buildUrl())
+            .post(buildJsonBody(email))
+            .header(ACCEPT_LANGUAGE, getLanguage())
+            .build();
+    }
+
+    private RequestBody buildJsonBody(String email) {
+        JSONObject json = new JSONObject();
+        try {
+            json.put(JSON_USERNAME, email);
```
dmsnell: is `JSON_USERNAME` necessary or does it get in the way of comprehension? I personally find it a small nuisance to read a name like that, jump to the top of the file, and then realize that the value matches the identifier. In these situations I like to ask myself, what if the JSON field were renamed? would we end up coming in here and renaming the constant as well? If the constant is 1:1-bound to the value, why not use the value and de-abstract, making it clearer when reading what JSON is being produced?

```java
json.put("username", email);
```

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.java
```diff
@@ -0,0 +1,39 @@
...
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+
+import com.automattic.simplenote.utils.AuthUtils;
+
+import net.openid.appauth.RedirectUriReceiverActivity;
+
+public class DeepLinkActivity extends AppCompatActivity {
+    private static final String AUTHENTICATION_SCHEME = "auth";
+    private static final String LOGIN_SCHEME = "login";
```
dmsnell, line 18: we also have internal links of the form `simplenote://note/NOTEID` - is this important here?

dmsnell (commented):
this is awesome work - thanks for the help, thanks for jumping in and getting all the setup work done. I know it was a heavy task considering that the app didn't previously handle much of this itself.

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignupFragment.java
hichamboushaba: I agree it confusing a bit, especially with multiple path segments, what I would do is passing the path as a single call using `addPathSegments`, but keep the host separated, and this to make it easier using different hosts, if we want to hit a different server for the development builds, we can then have the host in the `gradle.properties` like the simperium app id.
We were actually planning to discuss this point with you, because right now, the external contributors have no way to test the signup process, because we only offer the simperium server `history-analyst-dad` for debugging, but as it's not connected to `app.simplenote.com`, the signup won't work, I think it would be awesome if we can have a different development server connected to this simperium entity.

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.java
hichamboushaba: Are those only internal for internote, or are supposed to be handled from external apps too? The only deeplink we found in the app was `simplenote://auth?...` used for wpcom login, that's why it's the only one we have here.
The goal of creating `DeepLinkActivity` was making handling different deeplinks easier, by acting as a proxy and dispatching the deeplink to the responsible part.

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignupFragment.java
dmsnell: we probably won't have a signup server available for community contributors, unfortunately. if someone wants to fake the response they would be best to use any available network proxy/faker tool

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.java
dmsnell: It's my impression they should only be used internally. This question I raise is more from my lack of understanding than from my understanding though. Mainly I wanted to make sure y'all were aware of it and this wasn't limited by oversight.

ParaskP7:
👋 @dcalhoun !

> I tested this flow on a Samsung Galaxy S20 FE 5G. The flow succeeded as outlined and expected. 👍🏻

Many thanks for testing the solution David! 🙏 

> The only issue I encountered was that tapping the link in the email would not open my browser. I also tried long-pressing the link and selecting "Open in Browser" from the context menu. That did not work either. I was forced to copy the URL and paste it into my browser. If no one else experiences this, it likely is just a configuration specific to my testing device (e.g. Personal vs Work profiles).

Which `link in the email` do you refer to? Do you mean the `simplenote@support.com` link on the mail confirmation screen right after sending the sing up request?

ParaskP7:
👋  @dmsnell !

> this is awesome work - thanks for the help, thanks for jumping in and getting all the setup work done. I know it was a heavy task considering that the app didn't previously handle much of this itself.

Thank you so much for helping us on that, and for all the comments you left, me and @hichamboushaba will try to deal with them one by one today! 🙏

ParaskP7:
👋 @AmandaRiu !

> Hey @ParaskP7 👋 Review complete. Code looks good - just left a couple minor comments for review and consideration. Also noticed the following during testing:

Many thanks for the code review Amanda! 🙏 

> I did run into the null email page but I see it's already been explored.

Yes, we are aware of that and we will need make a decision on it. Thanks for confirming it! 🌟 

> I was able to signup with an email that was already connected to Simplenote and didn't receive an email. I see that's a support message - just thought it was strange that we can't verify that up front.

I am not sure if that is the expected behavior. Maybe @dmsnell and @eshurakov can help us figure out what to do in this case or what is the expected flow. Many thanks for trying that out! 🌟 

> None of the above are blockers and it looks like they are both known, just wanted to note it here.

Once again, many many thanks for your code review! 🙏

Simplenote/src/main/res/layout/fragment_confirmation.xml
ParaskP7: This is now done here: 00ec8ef3a5b0670f631f6b9e83bbede7649ea29b

Simplenote/src/main/res/layout/fragment_confirmation.xml
ParaskP7: This is now done here: cd8332561494a94bb3e13fe25059386075b91161

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignupFragment.java
ParaskP7: 👋 @dmsnell and @hichamboushaba !

I agree with both of you and I was thinking the same on the multiple `addPathSegment(...)` build calls. So, I think it is better if we group those and leave the other parts as is. I also agree with @hichamboushaba on his points. This kind of API call on Android, using `OkHttp` directly are dated, libraries like [Retrofit](https://square.github.io/retrofit/), which are the go to libraries for that matter, for many years now, are doing the same, but by using annotations instead to expose the segment path, the rest is abstracted elsewhere. Thus, I choose the same path in order to align with that.

PS: If you really feel strong about that we can definitely change that to a full path: `https://app.simplenote.com/account/request-signup`

ParaskP7: This is now done here: 96eeb89cddb0af23a644d80ec83f6b2855f4dda2

Simplenote/src/main/java/com/automattic/simplenote/utils/AuthUtils.java
ParaskP7: This is now done here: 9655f7dad165db92d7ad60b2f1f588b97cdc83a3

PS: We decided to go another way, let us know what you think.

Simplenote/src/main/java/com/automattic/simplenote/utils/AuthUtils.java
ParaskP7: This is now done here: d4c12c8dc9748f950ef68d51ddeaf59c247ad8e1

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignupFragment.java
ParaskP7: This is done here: 75965c66f31de062b473afb61866c9e2b152d099

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.java
hichamboushaba: Thanks for bringing it up, we confirmed that it's not related, internal note linking doesn't use deeplinks: https://github.com/Automattic/simplenote-android/blob/6d32ae2a2e4f48359598ba4c2fd9cd0d45687d3a/Simplenote/src/main/java/com/automattic/simplenote/utils/SimplenoteLinkify.java#L62-L90

ParaskP7:
👋 @AmandaRiu and @dmsnell !

@hichamboushaba and me addressed, to the best of our abilities, your comments. I have re-requested your review. Let us know what you think.

PS: @dmsnell  and @eshurakov , can you please take a look at this questions of ours since it hasn't been addressed yet: https://github.com/Automattic/simplenote-android/pull/1340#issuecomment-795447284

SylvesterWilmott:
>This is not trivial and will take a lot of effort to implement... unless this is a blocker, we should just keep that as is for now

That's fine for me.

>About the dark modes, I believe that the authentication and credential screens on Android don't have a dark mode in general. Should we do anything about that?

That is true! No action needed for that one.

I think we can keep the copy of the dialog more concise:

Already Logged In
You are already logged in as [email]. Log out and try again.
[Cancel] [Log Out]

Thanks!

ParaskP7:
👋 @SylvesterWilmott !

> That's fine for me.

🙏 

> That is true! No action needed for that one.

👌 

> I think we can keep the copy of the dialog more concise:
> Already Logged In
> You are already logged in as [email]. Log out and try again.
> [Cancel] [Log Out]

This is now done here: dc071e74a33fb7c4b2c1c212c7dbb0a255594543

PS: Thank YOU! 💫

dcalhoun:
> Which `link in the email` do you refer to? Do you mean the `simplenote@support.com` link on the mail confirmation screen right after sending the sing up request?

@ParaskP7 I meant the link to confirm your email address that is included in the email you receive after signing up. So, not in the app, but located in my email client. Because it is located outside of the app, I do not believe it to be an actual issue. I just wanted to share my experience in case anyone else experienced the same.

ParaskP7:
👋 @dcalhoun !

> @ParaskP7 I meant the link to confirm your email address that is included in the email you receive after signing up. So, not in the app, but located in my email client. Because it is located outside of the app, I do not believe it to be an actual issue. I just wanted to share my experience in case anyone else experienced the same.

Oh yes, thanks for making that clear! 🙏 

Yes, this is outside of the app's control and probably device, email and browser app specific. Maybe if you share your device, email app and browser app? I am so used in just using `Gmail` and `Chrome`, but this might not be the case for you.

dcalhoun:
> Yes, this is outside of the app's control and probably device, email and browser app specific. Maybe if you share your device, email app and browser app? I am so used in just using `Gmail` and `Chrome`, but this might not be the case for you.

@ParaskP7 tapping the confirm email link resulting in nothing happening (i.e. my browser did not open) occurred on a Samsung Galaxy S20 FE 5G running Android 11. The email client was Gmail `2021.02.05.357775197`. The browser is Google Chrome `88.0.4324.152` (set as OS' default browser).

Please keep in mind that I am novice Android user, so it's quite likely user error. I am primarily an iOS user.

ParaskP7:
👋 @dcalhoun !

> @ParaskP7 tapping the confirm email link resulting in nothing happening (i.e. my browser did not open) occurred on a Samsung Galaxy S20 FE 5G running Android 11. The email client was Gmail 2021.02.05.357775197. The browser is Google Chrome 88.0.4324.152 (set as OS' default browser).

I see nothing wrong with the setup as you are using a standard device, along with Google email and browser app clients. Probably there is some kind of setup done on that device that is causing this behavior. I am not sure if you have tried with other links as well? Maybe find another email within your `Gmail` app, and try to open any link you can find, just to start troubleshooting it.

PS: Feel free to DM me if you would like to troubleshoot this at some point with an additional pair of eyes.

AmandaRiu (approved):
Looks good to me 👍 I'll hold off on merging so @dmsnell can review the changes from his comments as well.

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.java
```diff
@@ -0,0 +1,43 @@
...
+            startActivity(intent);
+        } else if (uri.getHost().equals(LOGIN_SCHEME)) {
+            Intent intent = new Intent(this, NotesActivity.class);
+            Simplenote app = (Simplenote) getApplication();
+            String email = AuthUtils.extractEmailFromMagicLink(uri);
+            if (app.isLoggedIn() &&
+                !email.toLowerCase(Locale.US).equals(app.getUserEmail().toLowerCase(Locale.US))) {
+                intent.putExtra(NotesActivity.KEY_ALREADY_LOGGED_IN, true);
+            } else {
+                AuthUtils.magicLinkLogin((Simplenote) getApplication(), uri);
```
dmsnell, line 36: thanks for updating this. I can't leave a comment in the existing conversation, as all GitHub is showing me is a button to "resolve conversation," so I'm leaving this here.

I think this will be a valuable change to keep the login semantics separated, both in exposing `getUserEmail()` and also in adding `login` to `Simperium.java`. I don't understand the motivation behind taking a single semantic intention "is the user with this email logged in" and splitting it into separate internal details ("what is the email of the user" and "is the user logged in") (because we still end up duplicating logic from the main app in this activity right here) but this does seem better than it was and I don't want to hold up the PR on this point.

for curiosity, what was the purpose in extracting these internal details?

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -212,6 +218,30 @@ public boolean isInBackground() {
         return mIsInBackground;
     }
 
+    public void login(String email, String spToken) {
```
dmsnell: thanks for adding this function! I could see it being a little confusing having a login function taking two strings that aren't username-and-password, but I think we can evaluate that in a follow-up issue if it would hold up this pr.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -212,6 +218,30 @@ public boolean isInBackground() {
...
+        // Manually authorize the user with Simperium
+        User user = mSimperium.getUser();
```
dmsnell, line 223: interesting. I didn't realize we could do this right here. is there _always_ a user object in Simperium that we modify?

dmsnell (approved):
thanks for interacting with my feedback, and thanks for making this all work!

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.java
hichamboushaba: We actually don't need to check "if the user with this email is logged in", but it's "is the user logged in with a different email", which seems pretty coupled to the new signup flow, and not something reusable, and which should not be handled by the Application class, so we decided to just expose the email of the currently logged in user, and keep the logic of the new signup flow contained in its place.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
hichamboushaba: We don't know the internals of Simperium very well, this logic was actually copied from the logic with WPCom account, as it's similar in taking an email and a token, and authenticating the user with them.
But I can confirm that Simperium has always a user object that we can modify: https://github.com/Simperium/simperium-android/blob/4de51b2684da7bc76dfd625d1295b4bc0488a803/Simperium/src/main/java/com/simperium/Simperium.java#L74

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
hichamboushaba: I renamed the function to `loginWithToken` to make it clearer.

AmandaRiu:
Merging PR. Any design changes can be handled in a followup PR.

hichamboushaba:
I created an issue to keep track of the issue mentioned in [this comment](https://github.com/Automattic/simplenote-android/pull/1340#issuecomment-795447284): https://github.com/Automattic/simplenote-android/issues/1343

cc @dmsnell @eshurakov

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.java
dmsnell: > We actually don't need to check "if the user with this email is logged in", but it's "is the user logged in with a different email"

this took me a long time to understand and realize my mistake. thanks for clarifying the logic. it makes more sense now.

pachlava:
@mokagio should this be assigned to corresponding milestone? Thanks!

PR #1354, opened 2021-03-31
Tweak screenshots automation to shot note with links instead of inter-note linking view

mokagio:
### Fix
@SylvesterWilmott pointed out that, because we don't have descriptions in the screenshots in the Play Console, the shot with the inter-note linking view might seem confusing to potential users.

Screen Shot 2021-03-31 at 4 14 43 pm

We decided to use a screenshot of a note with internal links instead (p2XJRt-2t6-p2#comment-4784):

image

### Test

- Checkout this branch
- Run `bundle install`
- Launch a Simulator
- Run `bundle exec fastlane take_screenshots`

The process should eventually open your browser and show something like this:

image

Note that:

- It's not 100% stable, so if it fails try giving it another go. _This is absolutely awful, but let's be pragmatic and keep in mind that_ at the moment _this automation is used only for the occasional screenshot generation, so investing too much time in making it bulletproof is,_ unfortunately, _not on top of our priority list_.
- At least once, I got the new screenshot _without_ the syntax highlight for the links. Not sure how that happened. Again, I re-run the tests and it came out fine.

### Review

### Release


Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -126,20 +127,15 @@ public void screenshotTest() throws InterruptedException {
...
-        selectNoteFromNotesList(NOTE_FOR_EDITOR_SHOT_TITLE);
+        selectNoteAndTakeScreenshotFromNotesList(NOTE_FOR_EDITOR_SHOT_TITLE, "01-note");
 
-        // It can happen that the email verification screen appears on the note editor instead of
-        // the note list screen, so look for one and dismiss it if found.
-        dismissVerifyEmailScreenIfNeeded();
-
-        Screengrab.screenshot("01-note");
-
-        dismissNoteEditor();
```
mokagio, line 137: I dried all the code (red) in the new method (green) because the shot for the note with the internal links uses the same sequence.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -319,7 +327,7 @@ private void loadThemeSwitcherFromNotesList() {
         loadThemeSwitcherFromSettings();
     }
 
-    private void takeInterNoteLinkingScreenshotFromNotesList() throws InterruptedException {
+    private void takeInterNoteLinkingScreenshotFromNotesList(String noteName, String screenshotName) throws InterruptedException {
```
mokagio, line 375: I added these parameters to make the method seem less like a black box. In particular, I found it annoying to have to scroll or do some text search to find the place in which the screenshot file name was set.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
mokagio: I dried all the code (red) in the new method (green) because the shot for the note with the internal links uses the same sequence.

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -368,6 +376,11 @@ private void takeInterNoteLinkingScreenshotFromNotesList() throws InterruptedExc
         onView(withText("Trash")).perform(click());
         onView(withId(R.id.menu_empty_trash)).perform(click());
         onView(withText("Empty")).perform(click());
+
+        // Go back to notes list
+        loadSideMenuFromNotesList();    // The trash activity is obviously not the notes list one,
+                                        // but this method works anyways
+        onView(withText("All Notes")).perform(click());
```
mokagio, line 428: This logic was previously in the main method, but it belongs here because it restores the app to the state in which it was before calling this method.

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.java
```diff
@@ -129,7 +132,7 @@ public void run() {
                                     }
                                 }
                             },
-                            requireContext().getResources().getInteger(android.R.integer.config_mediumAnimTime)
+                            delay
```
mokagio, line 135: From 4ada7bb8735238ad284515f53b6161cdbcd611e2: 

While working on the screenshot automation, the tests started consistently failing after dismissing a note editor activity because the `NoteMarkdownFragment` was not attached to a context from the call edited in this commit.

I'm not sure why that would happen or if its of any concern in production.

With my limited Android knowledge, this seems to be a legit workaround. We access the context at a time in which we are sure it exists. The value we extract is constant over the application lifecycle anyways, so there's no downside in reading it from the context at a different time
than when it's going to be used.

@hichamboushaba @ParaskP7 you folks have worked on this repo recently and might know why that's happening and if there's a better way to do it. It think that `requireContext()` call appeared as part of 16e90824.

ParaskP7: 👋 @mokagio !

Thanks for taking care of this! 🌟 

I think this is a valid change as we need to treat this `delay` as constant and the way you did it has that effect within the `onPageFinished(...)` scope. There are many ways that this can further improve, but I believe this change is enough for now.

🙏

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
```diff
@@ -126,20 +127,15 @@ public void screenshotTest() throws InterruptedException {
...
-
-        dismissNoteEditor();
-
-        takeInterNoteLinkingScreenshotFromNotesList();
-
-        loadSideMenuFromNotesList();
-        onView(withText("All Notes")).perform(click());
+        // What we'd like to do is take a screenshot of the interlinking interface with:
+        //
+//        takeInterNoteLinkingScreenshotFromNotesList(NOTE_FOR_INTERLINKING_SHOT_TITLE, "03-inter-note-linking");
```
ParaskP7: Minor (🔍): Since all the comments start after the 8 character, I would consider align this comment to match the rest of them (in terms of indentation).

ParaskP7 (approved):
👋 @mokagio !

I was able to see the correct screenshots appear on my browser after the command completed. The only diff to your screenshots was that the keyboard didn't appear on my screenshots (5th one), but, I guess this is somehow expected, right? See below:

image

I have left only minor (🔍) comments for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. Also, it might be best if @jkmassel takes a look as well since I am not that much familiar with the overall logic within the `ScreenshotTest` class. 

PS: Feel free to ignore my minor comment and merge the PR yourself.
PSS: I also 👍 all your comments since I 💯 agree with them.

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.java
mokagio: Thank you for looking over this 🙇‍♂️

Simplenote/src/androidTest/java/com/automattic/simplenote/screenshots/ScreenshotTest.java
mokagio: Good point! I fixed up the original commit with this change in 80eb7aed.

ParaskP7: Thank you Gio! 🙏

mokagio:
@jkmassel I got to the bottom of the failure you experienced when reviewing. In between Petros' review and yours, the account password changed. I updated it, if you retry now (after updating the local secrets) it should work, with the same caveat from the original description: the tests are not super stable 😕

Thanks!

jkmassel (approved):
Worked great this time.

One potential area for improvement – it'd be super if it could ensure that an emulator exists and is started before attempting to run (so I don't have to do it by hand).

This would also help in CI I think.

For a future PR :)

PR #1376, opened 2021-04-18
Setup test dependencies and infrastructure for UI tests

### LINKED ISSUE #1362: Setup test dependencies and infrastructure for UI tests
Parent: #1361

TODO

- [x] Add dependencies for Kotlin since the tests will be written in Kotlin
- [x] Verify project has all dependencies needed for UI tests
- [x] Create double for `Bucket<Tag>`
- [x] Create double for `Bucket<Note>`
- [x] Create `SimplenoteTest` class which extends `Simplenote` Application class for tests to double `Bucket` dependencies
- [x] Create `SimplenodeAppRunner` class and point `testInstrumentationRunner` in `build.gradle` to it

dandominguez04:
Fixes #1362

### Fix
This PR adds test dependencies and infrastructure for UI tests to the project.

### Test
This PR just modifies the main runner class to instrumented tests. Test current instrumented tests.

1. Run `NoteListFragmentTest` (requires a valid account defined in `gradle.properties`)
2. Run `ScreenshotTest  (requires a valid account defined in `gradle.properties`)

### Review
This PR is part of a trial project guided by @ParaskP7. 

### Release

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.application'
 apply plugin: 'io.sentry.android.gradle'
+apply plugin: 'kotlin-android'
 
 android {
     buildToolsVersion '29.0.2'
@@ -37,6 +38,12 @@ android {
         targetSdkVersion 29
 
         testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
+    }
+
+    // Avoid duplicate annotations dependencies
+    configurations {
+        cleanedAnnotations
+        compile.exclude group: 'org.jetbrains' , module:'annotations'
     }
 
     lintOptions {
@@ -84,6 +91,10 @@ dependencies {
     implementation 'com.simperium.android:simperium:0.10.0'
     implementation 'com.automattic:tracks:1.2'
     implementation 'org.wordpress:passcodelock:2.0.2'
+
+    // Kotlin and KTX
+    implementation 'androidx.core:core-ktx:1.3.2'
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
 
     // Google Support
     implementation 'androidx.appcompat:appcompat:1.1.0'
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -72,10 +72,10 @@ buildscript {
 dependencies {
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.mockito:mockito-core:2.8.47'
-    androidTestImplementation 'androidx.test:core:1.0.0'
+    androidTestImplementation 'androidx.test:core:1.2.0'
     androidTestImplementation 'androidx.annotation:annotation:1.0.0'
-    androidTestImplementation 'androidx.test:runner:1.1.0'
-    androidTestImplementation 'androidx.test:rules:1.1.0'
+    androidTestImplementation 'androidx.test:runner:1.2.0'
+    androidTestImplementation 'androidx.test:rules:1.2.0'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.1.0'
     androidTestImplementation 'androidx.test.espresso:espresso-contrib:3.1.0'
     androidTestImplementation('androidx.test.espresso:espresso-contrib:3.1.0') {
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -72,10 +72,10 @@ buildscript {
 dependencies {
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.mockito:mockito-core:2.8.47'
-    androidTestImplementation 'androidx.test:core:1.2.0'
+    androidTestImplementation 'androidx.test:core:1.3.0'
     androidTestImplementation 'androidx.annotation:annotation:1.0.0'
-    androidTestImplementation 'androidx.test:runner:1.2.0'
-    androidTestImplementation 'androidx.test:rules:1.2.0'
+    androidTestImplementation 'androidx.test:runner:1.3.0'
+    androidTestImplementation 'androidx.test:rules:1.3.0'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.1.0'
     androidTestImplementation 'androidx.test.espresso:espresso-contrib:3.1.0'
     androidTestImplementation('androidx.test.espresso:espresso-contrib:3.1.0') {
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -92,8 +92,7 @@ dependencies {
     implementation 'com.automattic:tracks:1.2'
     implementation 'org.wordpress:passcodelock:2.0.2'
 
-    // Kotlin and KTX
-    implementation 'androidx.core:core-ktx:1.3.2'
+    // Kotlin
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
 
     // Google Support
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -40,12 +40,6 @@ android {
         testInstrumentationRunner 'com.automattic.simplenote.SimplenoteAppRunner'
     }
 
-    // Avoid duplicate annotations dependencies
-    configurations {
-        cleanedAnnotations
-        compile.exclude group: 'org.jetbrains' , module:'annotations'
-    }
-
     lintOptions {
         checkReleaseBuilds false
     }
@@ -89,9 +83,11 @@ dependencies {
     androidTestImplementation 'tools.fastlane:screengrab:2.0.0'
     // Automattic and WordPress dependencies
     implementation 'com.simperium.android:simperium:0.10.0'
-    implementation 'com.automattic:tracks:1.2'
+    implementation('com.automattic:tracks:1.2') {
+        exclude group: "org.jetbrains", module: "annotations-java5"
+    }
     implementation 'org.wordpress:passcodelock:2.0.2'
-    
+
     // Google Support
     implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation 'com.google.android.material:material:1.1.0-beta01'
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -37,7 +37,7 @@ android {
         minSdkVersion 23
         targetSdkVersion 29
 
-        testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
+        testInstrumentationRunner 'com.automattic.simplenote.SimplenoteAppRunner'
     }
 
     // Avoid duplicate annotations dependencies
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -91,10 +91,7 @@ dependencies {
     implementation 'com.simperium.android:simperium:0.10.0'
     implementation 'com.automattic:tracks:1.2'
     implementation 'org.wordpress:passcodelock:2.0.2'
-
-    // Kotlin
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-
+    
     // Google Support
     implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation 'com.google.android.material:material:1.1.0-beta01'
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteAppRunner.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteAppRunner.kt
@@ -1,0 +1,12 @@
+package com.automattic.simplenote
+
+import android.app.Application
+import android.content.Context
+import androidx.test.runner.AndroidJUnitRunner
+
+class SimplenoteAppRunner : AndroidJUnitRunner() {
+    @Throws(InstantiationException::class, IllegalAccessException::class, ClassNotFoundException::class)
+    override fun newApplication(cl: ClassLoader, className: String, context: Context): Application {
+        return super.newApplication(cl, SimplenoteTest::class.java.name, context)
+    }
+}
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteAppRunner.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteAppRunner.kt
@@ -5,7 +5,6 @@ import android.content.Context
 import androidx.test.runner.AndroidJUnitRunner
 
 class SimplenoteAppRunner : AndroidJUnitRunner() {
-    @Throws(InstantiationException::class, IllegalAccessException::class, ClassNotFoundException::class)
     override fun newApplication(cl: ClassLoader, className: String, context: Context): Application {
         return super.newApplication(cl, SimplenoteTest::class.java.name, context)
     }
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteTest.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteTest.kt
@@ -6,8 +6,7 @@ import com.automattic.simplenote.utils.TestBucket
 import com.simperium.client.Bucket
 
 class SimplenoteTest : Simplenote() {
-    // Decides whether use the default bucket implementation or our own TestBucket
-    var useDefaultBucketImpl = true
+    var useTestBucket = false // Decides whether to use the test bucket implementation or the default.
 
     private val tagsBucket = object : TestBucket<Tag>("tags") {
         override fun build(key: String?): Tag {
@@ -21,19 +20,8 @@ class SimplenoteTest : Simplenote() {
         }
     }
 
-    override fun getTagsBucket(): Bucket<Tag> {
-        if (useDefaultBucketImpl) {
-            return super.getTagsBucket()
-        }
+    override fun getTagsBucket(): Bucket<Tag> = if (useTestBucket) tagsBucket else super.getTagsBucket()
 
-        return tagsBucket
-    }
-
-    override fun getNotesBucket(): Bucket<Note> {
-        if (useDefaultBucketImpl) {
-            return super.getNotesBucket()
-        }
-
-        return notesBucket
-    }
+    override fun getNotesBucket(): Bucket<Note> = if (useTestBucket) notesBucket else super.getNotesBucket()
 }
+
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteTest.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteTest.kt
@@ -6,6 +6,9 @@ import com.automattic.simplenote.utils.TestBucket
 import com.simperium.client.Bucket
 
 class SimplenoteTest : Simplenote() {
+    // Decides whether use the default bucket implementation or our own TestBucket
+    var useDefaultBucketImpl = true
+
     private val tagsBucket = object : TestBucket<Tag>("tags") {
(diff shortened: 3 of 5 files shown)


ParaskP7:
👋 @dandominguez04 !

Just as a quick FYI, and obviously very minor, we usually have this `Fixes #1362` line at the top of any PR. Some automation tools might also depend on that.

dandominguez04:
@ParaskP7 I updated the description of the PR, thanks!

ParaskP7:
👋 @dandominguez04 !

> @ParaskP7 I updated the description of the PR, thanks!

Thank YOU! 💪 Actually, I would suggest to keep this as `Fixes #1362` since the tooling might look for that exact word, the `Fixes`. 😅 PS: You can drop the `.` at the end.

Simplenote/build.gradle
```diff
@@ -36,7 +37,13 @@ android {
         minSdkVersion 23
         targetSdkVersion 29
 
-        testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
+        testInstrumentationRunner 'com.automattic.simplenote.SimplenoteAppRunner'
```
ParaskP7, line 40: Warning (⚠️): By changing the `testInstrumentationRunner`, all current tests, that require the default `AndroidJUnitRunner` runner, will be starting to fail with the below stack trace, which is unrelated to having a valid account defined in `gradle.properties`. Have you considered an approach where multiple test instrumentation runners can be used.

We need to make sure that the already existing instrumentation tests are unaffected by the new `integration` ones? Otherwise running those locally using the CLI or remotely on CI will stop working (`./gradlew connectedDebugAndroidTest`). Isn't that right?

```
com.automattic.simplenote.NoteListFragmentTest > testNonSearchCursorReturnsAfterSearchApplied[Pixel_4_API_30(AVD) - 11] FAILED 
        java.lang.NullPointerException: Attempt to invoke interface method 'void com.simperium.storage.StorageProvider$BucketStore.reset()' on a null object reference
        at com.simperium.client.Bucket.reset(Bucket.java:902)
[InstrumentationResultParser]: Ignoring repeated failed event for com.automattic.simplenote.NoteListFragmentTest#testNonSearchCursorReturnsAfterSearchApplied. Stack: java.lang.RuntimeException: Unable to resume activity {com.automattic.simplenote.debug/com.automattic.simplenote.NotesActivity}: java.lang.NullPointerException: Attempt to invoke interface method 'void com.simperium.storage.StorageProvider$BucketStore.reset()' on a null object reference
at android.app.ActivityThread.performResumeActivity(ActivityThread.java:4444)
at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:4476)
at android.app.servertransaction.ResumeActivityItem.execute(ResumeActivityItem.java:52)
at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
at android.os.Handler.dispatchMessage(Handler.java:106)
at android.os.Looper.loop(Looper.java:223)
at android.app.ActivityThread.main(ActivityThread.java:7656)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
Caused by: java.lang.NullPointerException: Attempt to invoke interface method 'void com.simperium.storage.StorageProvider$BucketStore.reset()' on a null object reference
at com.simperium.client.Bucket.reset(Bucket.java:902)
at com.automattic.simplenote.utils.AuthUtils.logOut(AuthUtils.java:26)
at com.automattic.simplenote.NotesActivity.startLoginActivity(NotesActivity.java:1332)
at com.automattic.simplenote.NotesActivity.onResume(NotesActivity.java:276)
at android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1456)
at androidx.test.runner.MonitoringInstrumentation.callActivityOnResume(MonitoringInstrumentation.java:743)
at android.app.Activity.performResume(Activity.java:8135)
at android.app.ActivityThread.performResumeActivity(ActivityThread.java:4434)
... 11 more
```

Simplenote/build.gradle
```diff
@@ -36,7 +37,13 @@ android {
...
+    }
+
+    // Avoid duplicate annotations dependencies
+    configurations {
+        cleanedAnnotations
+        compile.exclude group: 'org.jetbrains' , module:'annotations'
```
ParaskP7: Suggestion (💡): This is indeed unfortunate that we need that here, probably a library is conflicting. I think this needs to be resolved on a library level. I am not sure if we really need to take a look at it now or later, but I'll definitely add a task for later, just to make sure we don't forget to look into that again at the end. Wdyt?

PS: I usually suggest atomic commits and in this case I would have added this configuration as a separate commit, which I would explain within the why it is actually needed and what it solves.

Simplenote/build.gradle
```diff
@@ -65,10 +72,10 @@ buildscript {
 dependencies {
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.mockito:mockito-core:2.8.47'
-    androidTestImplementation 'androidx.test:core:1.0.0'
+    androidTestImplementation 'androidx.test:core:1.2.0'
     androidTestImplementation 'androidx.annotation:annotation:1.0.0'
-    androidTestImplementation 'androidx.test:runner:1.1.0'
-    androidTestImplementation 'androidx.test:rules:1.1.0'
+    androidTestImplementation 'androidx.test:runner:1.2.0'
+    androidTestImplementation 'androidx.test:rules:1.2.0'
```
ParaskP7: Question (❓): Why did you choose to update to these libraries to `1.2.0` and not the actual latest version (which is the `1.3.0`)?

Simplenote/build.gradle
```diff
@@ -85,6 +92,10 @@ dependencies {
     implementation 'com.automattic:tracks:1.2'
     implementation 'org.wordpress:passcodelock:2.0.2'
 
+    // Kotlin and KTX
+    implementation 'androidx.core:core-ktx:1.3.2'
```
ParaskP7: Question (❓): Why did you add this library now, where will it help? I am asking that because I don't even see it being utilised on your next commit.

PS: IMHO I believe that it would be actually better if we were to be adding any libraries right when we would need them. Obviously, the `Kotlin` configuration is kind of different and thus this is okay.

Simplenote/build.gradle
```diff
@@ -85,6 +92,10 @@ dependencies {
...
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
ParaskP7: Question (❓): With the newest `Kotlin` plugin version, the need to explicitly define a dependency is removed, as such wouldn't it be okay if this dependency is removed altogether?

Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteAppRunner.kt
```diff
@@ -0,0 +1,12 @@
+package com.automattic.simplenote
+
+import android.app.Application
+import android.content.Context
+import androidx.test.runner.AndroidJUnitRunner
+
+class SimplenoteAppRunner : AndroidJUnitRunner() {
+    @Throws(InstantiationException::class, IllegalAccessException::class, ClassNotFoundException::class)
```
ParaskP7: Question (❓): Why did you add the `@Throws` annotation here, is that required somehow?

Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteTest.kt
```diff
@@ -0,0 +1,28 @@
+package com.automattic.simplenote
+
+import com.automattic.simplenote.models.Note
+import com.automattic.simplenote.models.Tag
+import com.automattic.simplenote.utils.TestBucket
+import com.simperium.client.Bucket
+
+class SimplenoteTest : Simplenote() {
```
ParaskP7, line 8: Praise (❤️): Well done!

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestBucket.kt
```diff
@@ -0,0 +1,52 @@
+package com.automattic.simplenote.utils
+
+import com.simperium.client.Bucket
+import com.simperium.client.BucketObjectMissingException
+import com.simperium.client.Syncable
+
+abstract class TestBucket<T : Syncable>(name: String) : Bucket<T>(
```
ParaskP7: 1) Praise (❤️): This is nicely thought, kudos!
2) Minor (🔍): Consider reformatting this class. I believe the formatting for this class can be further improved (see constructor).

ParaskP7 (changes requested):
👋 @dandominguez04 !

I have reviewed and tested this PR as per the instructions, well done, I really like what you did with the `SimplenoteTest` and `TestBucker` classes! 🌟 ❤️ 

I have left one warning (⚠️) comment, few questions (❓), some suggestions (💡) and a minor (🔍) comment for you to consider. I am not going to approve this PR just yet, since I wan't us to discuss the `testInstrumentationRunner` problem first.

dandominguez04:
Great feedback @ParaskP7. Let me work on improving your points.

ParaskP7:
👋 @dandominguez04 !

> Great feedback @ParaskP7. Let me work on improving your points.

Awesome, will take a look again tomorrow (or when you are ready)!

Simplenote/build.gradle
dandominguez04: I was careful on bumping libraries and I did not want to bump it to far from the current version. But `1.3.0` (I did not check `1.3.0` was the last one) is not a mayor version, thus I think it is OK to bump it to `1.3.0`.

Simplenote/build.gradle
dandominguez04: It is hard to specify different instrumentation class for different tests. We can do that with `Robolectric` (http://robolectric.org/configuring/) but we would need to add the dependency. The problem we are having here is that `NoteListFragmentTest` tries to use our tags bucket and notes buckets and out `TestBucket` does not redefine the method `reset()`. 

To solve this problem we are going to define a parameter that we can change to decide whether return the default bucket implementation (Simperium implementation) or our `TestBucket`. We can specify this parameter at the test level. The default behavior is to return the default implementation. For the tests we will define, we will set this variable to `false` to use `TestBucket` objects.

Simplenote/build.gradle
dandominguez04: Agree. I will remove it until we use it (if we use it).

Simplenote/build.gradle
dandominguez04: That is true. I double check and I removed it.

Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteTest.kt
dandominguez04: Thanks!

Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteAppRunner.kt
dandominguez04: I do not think it is needed. I will remove it.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestBucket.kt
dandominguez04: Is there a specific standard for Kotlin conventions that you follow at Automattic? I have changed the constructor but I am not sure if it complies with your standard.

Simplenote/build.gradle
dandominguez04: I found the problem in tracks and Kotlin. I could do this and the error disappears:
```gradle
implementation('com.automattic:tracks:1.2') {
    exclude group: "org.jetbrains", module: "annotations-java5"
}
```

But I am not sure the behavior of tracks will remain the same.

ParaskP7: 👋 @dandominguez04 !

Yes, I think this is a much better alternative since it targets the problems explicitly. Thanks for finding out the conflicting dependency. 🙏 

> But I am not sure the behavior of tracks will remain the same.

I believe it will, I am 99.9% sure of it, but you can never be absolutely sure with anything, can you... 😅

Simplenote/build.gradle
ParaskP7: 👋 @dandominguez04 !

Yeap, I agree, I think since we are bumping the versions, bumping up to the latest might be beneficial for us. And as you said, since this is not a major version update, I too think we are okay.

Simplenote/build.gradle
ParaskP7: Brilliant! 🙏

Simplenote/build.gradle
ParaskP7: Great, thanks! 🙏

Simplenote/src/androidTest/java/com/automattic/simplenote/SimplenoteAppRunner.kt
ParaskP7: Many thanks! 🙏

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestBucket.kt
ParaskP7: 👋 @dandominguez04 !

> Is there a specific standard for Kotlin conventions that you follow at Automattic? I have changed the constructor but I am not sure if it complies with your standard.

We are trying to use the default Kotlin style guide, the official one. And since this repo is just starting on Kotlin, I suggest we do the exact same thing, just use the default Kotlin style guide, the exact same thing that the community is using.

So, in this occasion, the only change would be on the constructor indeed.

ParaskP7:
👋 @dandominguez04 !

I noticed a couple of things and I wanted to provide you a couple of tips on them:
1) As an author we usually don't resolve a reviewer's comment. This is the responsibility of the reviewer, that is, to check whether a comment has been resolved and to then resolve it.
2) The author usually provides the commit `hash` that resolves the comment (if any). This way it is easier for the reviewer to review the update that resolves the comment and in general to document the associated change.
3) When you finish applying any change, based on a reviewer's commenting, you can click the `Re-request review` button next to the reviewer name on the top right to make sure that this PR will again appear on the reviewer list of pending PR to review.

Wdyt?

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestBucket.kt
ParaskP7: PS: I was think of reformatting that and keeping the multiline approach that you started with (see below), but the one line approach you ended up with is ok too.

```
abstract class TestBucket<T : Syncable>(name: String) : Bucket<T>(
    null,
    name,
    null,
    null,
    null,
    null
) {
```

Simplenote/build.gradle
ParaskP7: 👋 @dandominguez04 !

> It is hard to specify different instrumentation class for different tests. We can do that with Robolectric (http://robolectric.org/configuring/) but we would need to add the dependency.

I am not sure how `Robolectric` will solve that but I don't mind the extra dependency if it does.

PS: `Robolectric` and `Espresso` as progressively merging together and are now both part of Google, although I must admit that `Robolectric` development is not as active as I would have liked to. However, lately it started to pick up again, so... 🤷 

> The problem we are having here is that NoteListFragmentTest tries to use our tags bucket and notes buckets and out TestBucket does not redefine the method reset().

Yes, exactly that. But, this is not only it, not having the `reset()`, as `TestBucket` cannot be used at all with the existing instrumentation test since they are using the actual live `Bucket` and not mocking anything.

> To solve this problem we are going to define a parameter that we can change to decide whether return the default bucket implementation (Simperium implementation) or our TestBucket. We can specify this parameter at the test level. The default behavior is to return the default implementation. For the tests we will define, we will set this variable to false to use TestBucket objects.

I really like that solution, it is straightforward and doesn't require much build related configuration to what you already did. Seeing what you already do with the `BaseUITest.kt` on your subsequent #1378 PR, this help understanding its usage. Great thinking! 🌟 

```
    @Before
    fun setup() {
        application = ApplicationProvider.getApplicationContext() as SimplenoteTest
        // Make sure to use TestBucket buckets in UI tests
        application.useDefaultBucketImpl = false

        tagsBucket = application.tagsBucket as TestBucket<Tag>
        tagsBucket.clear()
    }
```

The only thing I might have changed, but this is very minor and merely a suggestion, is the `useDefaultBucketImpl` field to `useTestBucket` and revert the logic from `true` to `false`. This is so that to make everything related to the `SimplenoteTest` a test bucket first approach.

Then, a couple of additional refactoring would make this `SimplenoteTest` look like the below:

```
class SimplenoteTest : Simplenote() {
    var useTestBucket = false // Decides whether to use the test bucket implementation or the default.

    private val tagsBucket = object : TestBucket<Tag>("tags") {
        override fun build(key: String?): Tag {
            return Tag(key)
        }
    }

    private val notesBucket = object : TestBucket<Note>("notes") {
        override fun build(key: String?): Note {
            return Note(key)
        }
    }

    override fun getTagsBucket(): Bucket<Tag> = if (useTestBucket) tagsBucket else super.getTagsBucket()

    override fun getNotesBucket(): Bucket<Note> = if (useTestBucket) notesBucket else super.getNotesBucket()
}
```

Then, for the `BaseUITest`, this will make the setup more explicit towards the test buckets instead of the default, since whoever would like to extend from the `BaseUITest` would now know that it actually enables the test buckets instead of disabling the default, like:

```
        application.useTestBucket = true
```

Wdyt? I know naming is hard, so please take all that with a grain of salt and go with whatever you decide is best. 🤝

dandominguez04:
HI @ParaskP7 ,

> As an author we usually don't resolve a reviewer's comment. This is the responsibility of the reviewer, that is, to check whether a comment has been resolved and to then resolve it.

Noted. 

> The author usually provides the commit hash that resolves the comment (if any). This way it is easier for the reviewer to review the update that resolves the comment and in general to document the associated change.

I think that helps a lot. Let me follow that practice.

> When you finish applying any change, based on a reviewer's commenting, you can click the Re-request review...

Noted. I always communicate directly so I follow a bad practice 🙄️.

Simplenote/build.gradle
dandominguez04: Yes, I think it is simple. I applied the change in commit 62295bd6.

ParaskP7:
👋 @dandominguez04 !

> Noted.

🙏 

> I think that helps a lot. Let me follow that practice.

🙏 

> Noted. I always communicate directly so I follow a bad practice 🙄️.

Don't think about it like that. 🤗  Every team and organization does it differently. I was merely pointing out what we do here in case you would want to follow it as well. 👍

Simplenote/build.gradle
dandominguez04: Yes, I also will test this manually. I have applied the changed in 72285a54f04c8fa0e02e3bbc7803cd97587ecd8b.

Simplenote/build.gradle
ParaskP7: 👋 @dandominguez04 !

Many thanks, looking good! 🥇 

PS: Thanks for providing me with the commit that applies the change. 🙏

Simplenote/build.gradle
ParaskP7: Great, LGTM! 👍

ParaskP7 (approved):
👋 @dandominguez04 !

I have reviewed and tested this PR again, everything LGTM, great job working on the suggested changes! 🌟 🌟

PR #1387, opened 2021-04-23
Opt-out from translation for all strings using resource references

mokagio:
### Fix
See https://github.com/Automattic/simplenote-android/commit/d15153fadfdc6945982ee60df4bd723c719fb8c0#r49837835. Thanks @AliSoftware!

### Test
No manual test required.

### Review

@ParaskP7, I asked for your review as an Android subject matter expert because of your recent work on the project. Thanks! 🙇‍♂️ 

### Release

### DIFF
--- a/Simplenote/src/main/res/values/strings.xml
+++ b/Simplenote/src/main/res/values/strings.xml
@@ -113,10 +113,10 @@
 
     <!-- Tags -->
     <string name="add_tag">Add Tag</string>
-    <string name="dialog_tag_add_button_negative">@android:string/cancel</string>
-    <string name="dialog_tag_add_button_positive">@string/save</string>
-    <string name="dialog_tag_add_hint">@string/tag_name</string>
-    <string name="dialog_tag_add_title">@string/add_tag</string>
+    <string name="dialog_tag_add_button_negative" translatable="false">@android:string/cancel</string>
+    <string name="dialog_tag_add_button_positive" translatable="false">@string/save</string>
+    <string name="dialog_tag_add_hint" translatable="false">@string/tag_name</string>
+    <string name="dialog_tag_add_title" translatable="false">@string/add_tag</string>
     <string name="dialog_tag_error_message">An error occurred adding the tag. Please, try again. If the problem continues, contact us at %1$s%2$s%3$s%4$s%5$s for help.</string>
     <string name="dialog_tag_error_message_email" translatable="false">\u003c\u0061\u0020\u0068\u0072\u0065\u0066\u003d\u0022\u006d\u0061\u0069\u006c\u0074\u006f\u003a\u0073\u0075\u0070\u0070\u006f\u0072\u0074\u0040\u0073\u0069\u006d\u0070\u006c\u0065\u006e\u006f\u0074\u0065\u002e\u0063\u006f\u006d\u0022\u003e\u0073\u0075\u0070\u0070\u006f\u0072\u0074\u0040\u0073\u0069\u006d\u0070\u006c\u0065\u006e\u006f\u0074\u0065\u002e\u0063\u006f\u006d\u003c\u002f\u0061\u003e</string>
     <string name="tag_error_empty">Tags cannot be empty</string>
@@ -385,6 +385,6 @@
     <string name="simperium_hint_email">Email</string>
     <string name="simperium_hint_password">Password</string>
     <string name="simperium_subtitle">The simplest way to keep notes.</string>
-    <string name="simperium_title">@string/app_name</string>
+    <string name="simperium_title" translatable="false">@string/app_name</string>
     <string name="simperium_url">http://simplenote.com</string>
 </resources>
--- a/Simplenote/src/main/res/values/strings.xml
+++ b/Simplenote/src/main/res/values/strings.xml
@@ -112,7 +112,6 @@
     <string name="share_analytics_summary">Help us improve Simplenote by sharing usage data with our analytics tool. %1$sLearn more%2$s</string>
 
     <!-- Tags -->
-    <string name="test_dialog_string">@string/save</string>
     <string name="add_tag">Add Tag</string>
     <string name="dialog_tag_add_button_negative" translatable="false">@android:string/cancel</string>
     <string name="dialog_tag_add_button_positive" translatable="false">@string/save</string>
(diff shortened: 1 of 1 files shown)


AliSoftware:
One thing worth checking is why this happened in the first place, because we're supposed to have Peril warnings about those via [this](https://github.com/Automattic/peril-settings/blob/master/peril-settings.json#L108) + [this](https://github.com/Automattic/peril-settings/blob/master/org/pr/android.ts#L21-L29) when one forgets to add the attribute to a reference string.

- Did the Peril warning not work when those entries were introduced?
- Did it work but was ignored during review? (and should we make it an error instead of a warning then to ensure it can't be ignored in the future?)
- Or is it just because the rule was not in place yet at the time those strings were added to the repo, and the rule was only added later? (But if that was this latter case my understanding of the rule is that **any** future change to the `values/strings.xml` file even for other entries would re-run the check on the whole file content and warn anyway at that point…)

In any case I think it's worth checking the Peril rule works by now, maybe by temporarily introducing a bogus entry on purpose in a commit in this PR and see what Peril says. If it doesn't warn, we need to fix the Peril rule to avoid the regression and similar cases to happen again in the future.

Simplenote/src/main/res/values/strings.xml
```diff
@@ -113,10 +113,10 @@
 
     <!-- Tags -->
     <string name="add_tag">Add Tag</string>
-    <string name="dialog_tag_add_button_negative">@android:string/cancel</string>
-    <string name="dialog_tag_add_button_positive">@string/save</string>
-    <string name="dialog_tag_add_hint">@string/tag_name</string>
-    <string name="dialog_tag_add_title">@string/add_tag</string>
+    <string name="dialog_tag_add_button_negative" translatable="false">@android:string/cancel</string>
```
AliSoftware, line 116: Nice catch, I think I'd personally have missed the case for `@android:string` (and would only have thought about searching for `@string/`) 👍

I noticed that's a case that we also forgot about in our Peril rule (which only checks for `@string` not `@android:string`), so probably worth fixing by adding those to the Peril rule too

Simplenote/src/main/res/values/strings.xml
```diff
@@ -113,10 +113,10 @@
...
+    <string name="dialog_tag_add_button_positive" translatable="false">@string/save</string>
+    <string name="dialog_tag_add_hint" translatable="false">@string/tag_name</string>
+    <string name="dialog_tag_add_title" translatable="false">@string/add_tag</string>
     <string name="dialog_tag_error_message">An error occurred adding the tag. Please, try again. If the problem continues, contact us at %1$s%2$s%3$s%4$s%5$s for help.</string>
     <string name="dialog_tag_error_message_email" translatable="false">\u003c\u0061\u0020\u0068\u0072\u0065\u0066\u003d\u0022\u006d\u0061\u0069\u006c\u0074\u006f\u003a\u0073\u0075\u0070\u0070\u006f\u0072\u0074\u0040\u0073\u0069\u006d\u0070\u006c\u0065\u006e\u006f\u0074\u0065\u002e\u0063\u006f\u006d\u0022\u003e\u0073\u0075\u0070\u0070\u006f\u0072\u0074\u0040\u0073\u0069\u006d\u0070\u006c\u0065\u006e\u006f\u0074\u0065\u002e\u0063\u006f\u006d\u003c\u002f\u0061\u003e</string>
```
AliSoftware, line 121: I'm not sure what that is (didn't took the time to decode the hex entities to decipher the message, I'm on my phone so not easy ^^) and why this is escaped like that. If that's supposed to be an actual text once hex-decoded, not sure why it would be untranslated anyway… if there's a valid reason to all that, maybe worth an XML comment in there? 🤔

ParaskP7 (dismissed):
👋 @mokagio !

I have reviewed these changes and they seems okay from where I am standing. 👍 

If the tooling on your side, `Peril` ,`GlotPress` and all, works as expected with this change then we are 🟢  to merge.

Simplenote/src/main/res/values/strings.xml
AliSoftware: Now that I'm back on my Mac, I checked and the hex, once decoded, translates to this (well I've just replaced a couple of chars with `_` in my translation to avoid crawling – see below – but I'm sure you can guess what they actually are):
```
<a href="m_ilto:su_port_a_simple_ote_com">su_port_a_simple_ote_com</a>
```

* I can now understand why this would be `translatable="false` (unrelated to the other cases with `@string/…`, just because an email address doesn't need translation)
* But I'm still not clear why this was all in hex rather than be `&lt;a href="…"&gt;…&lt;/a&gt;`.

My guess is this was done to obfuscate the email address from robots that would try to crawl the OSS repo for email addresses (which is why I've obfuscated it in this very comment too, just in case), especially after the attack we had a while ago (for account creation, but still)? But even if that's the reason, which would indeed justify obfuscating with hex, I think we could at least benefit from an XML comment 🙂

mokagio:
Just a comment here to say I haven't forgotten about this but haven't had the time to followup to research "why did this happen?" or address the other comments Olivier left.

Thanks your patience.

ParaskP7:
👋 @mokagio !

Any updates on this PR?

PS: For now, I will unassign myself from it for now, just to not have it sit in my assigned list on GitHub. However, and when you or anyone else progresses this PR any further, please feel free to add me as an assigner again. 😎

ParaskP7 (commented):
I am not sure what is the latest state on this PR, as such I am just commenting here to ask for an update.

ParaskP7:
I am also adding @danilo04 to take this over now.

mokagio:
Yey @danilo04! Excited to have you on board. Sorry to drop an half backed issue in your plate. Ping me if you need any info.

danilo04:
No problem @mokagio! Let me test to see if I find what was the problem.

danilo04:
Hi 👋  @mokagio and @AliSoftware, 

I tested peril in commit ad91121096d151afaf9bb3f9828939d7b7682d21 but peril didn't complain. I will try to add more tests to see if I can find the root cause why peril is not complaining in these cases.

danilo04 (approved):
I am going to approve this PR so we don't block it. I have opened a new issue (#1435) to test with more detail why peril warnings are not firing. 

cc @mokagio @AliSoftware

mokagio:
Thanks @danilo04 for taking over this

PR #1390, opened 2021-04-27
Updates the release toolkit version

JavonDavis:
This is a big version jump. The changes specifically addressed in this PR include

* Updates to the `get_prs_list` callsite 
* Changes to use the new action to download translations from GlotPress.

To test: Create a temporary branch and run `bundle exec fastlane download_translations` this should download the translations from GlotPress, verify that this works as expected.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -175,7 +175,6 @@ SUPPORTED_LOCALES = [
   # bundle exec fastlane download_translations
   #####################################################################################
   lane :download_translations do |options|
-    # Adapt the GP project URL and lint_task as needed if testing a project other than WordPress
     android_download_translations(
       res_dir: File.join(ENV['PROJECT_NAME'], 'src', 'main', 'res'),
       glotpress_url: 'https://translate.wordpress.com/projects/simplenote/android/',
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -25,22 +25,22 @@ ENV["PROJECT_NAME"]="Simplenote"
 ENV["validate_translations"]="buildRelease"
 
 SUPPORTED_LOCALES = [
-  { glotpress: "ar", google_play: "ar",  promo_config: {}},
-  { glotpress: "de", google_play: "de-DE",  promo_config: {} },
-  { glotpress: "es", google_play: "es-ES",  promo_config: {} },
-  { glotpress: "fr", google_play: "fr-FR",  promo_config: {} },
-  { glotpress: "he", google_play: "iw-IL",  promo_config: {} },
-  { glotpress: "id", google_play: "id",  promo_config: {} },
-  { glotpress: "it", google_play: "it-IT",  promo_config: {} },
-  { glotpress: "ja", google_play: "ja-JP",  promo_config: {} },
-  { glotpress: "ko", google_play: "ko-KR",  promo_config: {} },
-  { glotpress: "nl", google_play: "nl-NL",  promo_config: {} },
-  { glotpress: "pt-br", google_play: "pt-BR",  promo_config: {} },
-  { glotpress: "ru", google_play: "ru-RU",  promo_config: {} },
-  { glotpress: "sv", google_play: "sv-SE",  promo_config: {} },
-  { glotpress: "tr", google_play: "tr-TR",  promo_config: {} },
-  { glotpress: "zh-cn", google_play: "zh-CN",  promo_config: {} },
-  { glotpress: "zh-tw", google_play: "zh-TW",  promo_config: {} },
+  { glotpress: "ar", android: "ar", google_play: "ar",  promo_config: {}},
+  { glotpress: "de", android: "de", google_play: "de-DE",  promo_config: {} },
+  { glotpress: "es", android: "es", google_play: "es-ES",  promo_config: {} },
+  { glotpress: "fr", android: "fr", google_play: "fr-FR",  promo_config: {} },
+  { glotpress: "he", android: "he", google_play: "iw-IL",  promo_config: {} },
+  { glotpress: "id", android: "id", google_play: "id",  promo_config: {} },
+  { glotpress: "it", android: "it", google_play: "it-IT",  promo_config: {} },
+  { glotpress: "ja", android: "ja", google_play: "ja-JP",  promo_config: {} },
+  { glotpress: "ko", android: "ko", google_play: "ko-KR",  promo_config: {} },
+  { glotpress: "nl", android: "nl", google_play: "nl-NL",  promo_config: {} },
+  { glotpress: "pt-br", android: "pt-rBR", google_play: "pt-BR",  promo_config: {} },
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -163,6 +166,24 @@ SUPPORTED_LOCALES = [
...
+  #####################################################################################
+  # download_translations
+  # -----------------------------------------------------------------------------------
+  # This lane download the string translations from GlotPress
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane download_translations
+  #####################################################################################
+  lane :download_translations do |options|
+    # Adapt the GP project URL and lint_task as needed if testing a project other than WordPress
```
AliSoftware: We can get rid of that comment from the PR test instructions now 🙂 
```suggestion
```

fastlane/Fastfile
```diff
@@ -180,7 +201,12 @@ SUPPORTED_LOCALES = [
     android_finalize_prechecks(options)
     configure_apply(force: is_ci)
     hotfix = android_current_branch_is_hotfix
-    android_update_metadata(options) unless hotfix
+    android_download_translations(
+      res_dir: File.join(ENV['PROJECT_NAME'], 'src', 'main', 'res'),
+      glotpress_url: 'https://translate.wordpress.com/projects/simplenote/android/',
+      locales: SUPPORTED_LOCALES,
+      lint_task: 'lintRelease'
+    ) unless hotfix
```
AliSoftware: Any reason why you're not invoking the other lane (to DRY code) instead of repeating its code here?

```suggestion
    download_translations(options) unless hotfix
```

fastlane/Pluginfile
```diff
@@ -2,4 +2,4 @@
 #
 # Ensure this file is checked in to source control!
 
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.13.0'
+gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', branch: 'release/0.17.2'
```
AliSoftware: 🚨  We need to switch this back to a tag (once new release toolkit is indeed released) before merging this PR

fastlane/Fastfile
```diff
@@ -25,22 +25,22 @@ ENV["PROJECT_NAME"]="Simplenote"
...
+  { glotpress: "it", android: "it", google_play: "it-IT",  promo_config: {} },
+  { glotpress: "ja", android: "ja", google_play: "ja-JP",  promo_config: {} },
+  { glotpress: "ko", android: "ko", google_play: "ko-KR",  promo_config: {} },
+  { glotpress: "nl", android: "nl", google_play: "nl-NL",  promo_config: {} },
+  { glotpress: "pt-br", android: "pt-rBR", google_play: "pt-BR",  promo_config: {} },
+  { glotpress: "ru", android: "ru", google_play: "ru-RU",  promo_config: {} },
+  { glotpress: "sv", android: "sv", google_play: "sv-SE",  promo_config: {} },
+  { glotpress: "tr", android: "tr", google_play: "tr-TR",  promo_config: {} },
+  { glotpress: "zh-cn", android: "zh-rCN", google_play: "zh-CN",  promo_config: {} },
+  { glotpress: "zh-tw", android: "zh-rTW", google_play: "zh-TW",  promo_config: {} },
```
AliSoftware, line 43: I see that `tools/exported-language-codes.csv` also had an entry for `cy` (Welsh), which means that the legacy script was also pulling translations for `cy` from GlotPress, while this locale is absent from the `SUPPORTED_LOCALES` list above.

This led me to wonder:
 - When [looking at GlotPress](https://translate.wordpress.com/projects/simplenote/android/), I see that Welsh (`cy`) is only translated to 29%.
 - But I also see that there are many more locales in this GlotPress project that are not listed here… and were not listed in the `tools/exported-languages-codes.csv` in the first place either.

Which raises the question:
 - Should we stop pulling the translations from locales other than the Mag16 that listed above, and just ignore all other locales (which are all only <30% translated according to GlotPress), in other words keep the list you have above?
 - Should we add just `cy` (Welsh) to that list above, in order to match what was done before the change made by this PR (since `cy` was listed in `tools/exported-language-codes.csv`), but not add any other? (not sure why we were in that state of making an exception for cy but not others, but maybe there was a reason)
 - Or should we all all the locales that are supported by GlotPress, even the one with low translation %? Which means that for those locales the app will only be partially translated (and fallback to English for any string copy that isn't). Which one could argue that it's better than nothing, but other could argue that it might be more confusing and frustrating to users of those locales than just having the consistency of everything in English

@belcherj @mokagio any thoughts on this? My take is if we have those locales in the GlotPress project to begin with it would be a waste not to use them at all and ignore them completely, but I can also see a rationale for preferring avoiding partially-translated app.

AliSoftware: And btw this raises the same question for the release notes ([GlotPress project](https://translate.wordpress.com/projects/simplenote/android/release-notes/)).

@JavonDavis if we end up deciding that we add more locales to that list – to download _every_ locale supported by the GlotPress projects, not just the Mag16 – we will need to check if the list of locales we want to pull for the `strings.xml` files (main GlotPress project / `android_download_translations`) and for the `release_notes.xml` (Release Notes GP project / `android_create_xml_release_notes`) end up being the same. If it ends up being a different list of locales, we will need 2 different tables (see my suggestions to handle that in the comments on wordpress-mobile/release-toolkit#229)

AliSoftware (changes requested):
Thanks for taking care of this!

There are some things to DRY, but also mostly a question about which locales we want to download, both for the `strings.xml` files (app translations) and for the release notes metadata.

* You should also take the occasion to remove the now obsolete `tools/update-translations.sh` and `tools/exported-language-codes.csv` files now that there are replaced by the new action
* I also wonder if the `tools/get-translated-release-notes.sh` – and associated `tools/extract_release_notes.rb` and `tools/release-notes-language-codes.csv` files – are still in use? I think they have long been replaced by the call to `android_create_xml_release_notes` a while ago but those old scripts never got removed? In that case they should also be safe to remove (a grep on the repo didn't show any call to `get-translated-release-notes.sh` at least). @mokagio can you confirm or deny this?

fastlane/Fastfile
mokagio: > @belcherj @mokagio any thoughts on this? My take is if we have those locales in the GlotPress project to begin with it would be a waste not to use them at all and ignore them completely, but I can also see a rationale for preferring avoiding partially-translated app.

FWIW, I'm on the "let's used them all" camp side.

fastlane/Fastfile
JavonDavis: Thanks for catching this! Yep we can definitely reuse the lane here.

mokagio:
_Miss you Javon..._ 😢 👋

Issue #1369, opened 2021-04-14
Add AddTagViewModel to restructure UI and orchestration logic in AddTagActivity


dandominguez04:
Parent: #1361

TODO

- Add infraestructure to integrate ViewModels in Activities and Fragments
- Refactor data management code to AddTagViewModel
- Refactor views to use ViewBinding
- Add tests for AddTagViewModel

PR #1393, opened 2021-05-03
Add dependencies  for ViewModels and restructure AddTagActivity to use ViewModels and Viewbinding

### LINKED ISSUE #1368: Add dependencies for Viewmodels, Livedata
Parent: #1361

TODO

- Add dependency for ViewModels `androidx.lifecycle:lifecycle-viewmodel-ktx`
- Add dependency for LiveData `androidx.lifecycle:lifecycle-livedata-ktx`
- Add ViewBinding configuration

dandominguez04:
Fixes #1368, #1369

### Fix
This PR adds `ViewModel` and `LiveData` dependencies to the project (issue #1368). It also restructure the code of `AddTagActivity` to use ViewModels and Viewbinding (issue #1369) in order to decouple data managment logic from the view. 

#### Automatic Tests
1. Run `AddTagActivityTest` inside `androidTest`. It requires to run an on emulator or physical device. 
2. Run `AddTagViewModelTest inside `test`

#### Manual Tests
1. Open Simplenote app
2. Tab on left menu
3. Tab on EDIT tags
4. Tab on Add Tag button (+ button on right-bottom corner)
5. Try adding valid tags and tags with errors (e.g. tag with spaces, tags that already exists)

### Review
This PR will be reviewed by @planarvoid. 

### Release


Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
```diff
@@ -95,56 +77,87 @@ public void onClick(View v) {
...
+                    DisplayUtils.hideKeyboard(binding.tagInput);
+                }
+            }
+        });
+
+        // Setup observer for result of saving a tag
+        viewModel.isResultOK().observe(this, new Observer<Boolean>() {
+            @Override
+            public void onChanged(Boolean tagSaved) {
+                if (tagSaved) {
```
planarvoid: I think this moves the logic unnecessarily to the activity. I'd rather either use 2 live data instances for finish and dialog actions, or even better have something like `val event: LiveData<Event>` where event is an enum with values like `FINISH` and `SHOW_ERROR`. 
Moreover this solution can crash because the live data event could be consumed more than once (it will most likely not happen but it's possible). I'd create a Github issue for the future to introduce something like a generic event that could be consumed only once (check WPAndroid for the approach).

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
```diff
@@ -0,0 +1,65 @@
...
+class AddTagViewModel(
+        private val tagsBucket: Bucket<Tag>,
+        defaultDispatcher: CoroutineDispatcher = Dispatchers.Default
+) : ScopedViewModel(defaultDispatcher) {
+
+    private val _showKeyboard = MutableLiveData<Boolean>()
+    val showKeyboard: LiveData<Boolean> = _showKeyboard
+
+    private val _tagError = MutableLiveData<Int?>()
+    val tagError: LiveData<Int?> = _tagError
```
planarvoid: I'd intuitively replace this by an `UiState` object which would contain both the error and the text value. Both would be populated in the `validateTag` call and the `text` value would be used on save instead of retrieving the value from the field. Let me know what you think! Maybe it's out of scope of this PR. This UI State could contain the `showKeyboard` as well. 

Overall I think the ViewModel needs to expose only 2 live data instances - the UI state and the events that are supposed to be consumed only once  - navigation events, dialogs, finish events and so on.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/ScopedViewModel.kt
```diff
@@ -0,0 +1,16 @@
+package com.automattic.simplenote.viewmodels
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.*
+import kotlin.coroutines.CoroutineContext
+
+abstract class ScopedViewModel(private val defaultDispatcher: CoroutineDispatcher) : ViewModel() {
```
planarvoid: I know we use this approach in WPAndroid but I think you don't need this class (there are some historical reasons why it's used there). I think you can just use the plan `ViewModel` and the `viewModelScope.launch` here. In the unit tests you could replace the main dispatcher or inject it and call viewModelScope.launch(injectedDispatcher). I think we will eventually replace it in WPAndroid as well.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        _tagError.value = null
+    }
+
+    fun saveTag(tagName: String) {
+        try {
+            _showKeyboard.value = false
+
+            TagUtils.createTagIfMissing(tagsBucket, tagName)
```
planarvoid: If you implement the unit tests, this would've to be wrapped.

planarvoid (commented):
Thanks for the changes! It's definitely better than before and it's working well! I'd personally really implement a proper unit test for the ViewModel. I think it won't be that hard and has to be done at one point anyway.

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
```diff
@@ -15,78 +15,60 @@
...
                 MorphCircleToRectangle.DURATION
-            );
-        }
+        );
 
-        mButtonNegative = findViewById(R.id.button_negative);
-        mButtonNegative.setOnClickListener(
+        binding.buttonNegative.setOnClickListener(
             new View.OnClickListener() {
                 @Override
```
planarvoid: I'd run this click through the view model as well (add an `onNegativeClick` method which would emit a `finish` event)

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
```diff
@@ -15,78 +15,60 @@
...
+    private void setupViews() {
+        binding.title.setText(getString(R.string.add_tag));
+        binding.tagInput.addTextChangedListener(this);
+        new Handler().postDelayed(
                 new Runnable() {
                     @Override
                     public void run() {
-                        mTagInput.requestFocus();
-                        DisplayUtils.showKeyboard(mTagInput);
+                        viewModel.showKeyboard();
```
planarvoid: maybe use a different method name? the activity doesn't need to know that this triggers `showKeyboard`. Maybe something like `onAnimationFinished` or `onViewLoaded` or anything that sounds right

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
```diff
@@ -15,78 +15,60 @@
...
-        mTagInput = findViewById(R.id.tag_input);
-
-        if (mTagInput != null) {
-            mTagInput.addTextChangedListener(this);
+        setContentView(binding.getRoot());
+    }
 
-            new Handler().postDelayed(
+    private void setupViews() {
+        binding.title.setText(getString(R.string.add_tag));
```
planarvoid, line 59: NIT: I quite like the something like 
```
private void setupView() = with(binding) {
}
```
If you do that, you don't have to use the `binding.view` calls and just call directly the views. It's little more readable.

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
```diff
@@ -15,78 +15,60 @@
...
 import java.util.Objects;
 
 public class AddTagActivity extends AppCompatActivity implements TextWatcher {
-    private Bucket<Tag> mBucketTag;
-    private Button mButtonNegative;
-    private Button mButtonPositive;
-    private TextInputEditText mTagInput;
-    private TextInputLayout mTagLayout;
+    private AddTagViewModel viewModel;
+    private ActivityTagAddBinding binding;
```
planarvoid: this will cause memory leaks. Generally you need to set the binding to null in `onDestroy`. However, in this case you don't need to have this as a field at all. You can just initialize it and pass it as a parameter to all the methods because it's only used in the `onCreate` method, right?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/ScopedViewModel.kt
dandominguez04: Ok sounds good. Fixed in c11e2f2e680613700fdff1b2559b0da36fe07e9b

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
dandominguez04: Fixed in ff0380dcc28559b1f4652e58177aa23988a8e8e0.

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
dandominguez04: Fixed in ff0380dcc28559b1f4652e58177aa23988a8e8e0.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
dandominguez04: I have added a `UiState` field and also one to handle the events. The only concern with the current solution is that it creates a new UiState for each text change in the field. 

Fixed in ff0380dcc28559b1f4652e58177aa23988a8e8e0.

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
dandominguez04: I have added an `event` field that handle the following events:

- CLOSE: closes the activity without saving a tag
- FINISH: closes the activity after the tag was saved correctly and send the result of the activity as OK
- SHOW_ERROR: shows a dialog with a error message in case the tag could not be saved

Fixed in ff0380dcc28559b1f4652e58177aa23988a8e8e0.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/ScopedViewModel.kt
dandominguez04: I removed it for now. I will take a look when I need to use coroutines (mostly in `TagsActivity`).

Fixed in c11e2f2e680613700fdff1b2559b0da36fe07e9b.

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.java
dandominguez04: This was changed to `viewModel.start()` to represent a loaded or started state in the UI. Fixed in ff0380dcc28559b1f4652e58177aa23988a8e8e0.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
dandominguez04: Fixed in d3a5980c7eaad1273fe417ffac4786a8c2d56ea0.

dandominguez04:
Hi @planarvoid. I have convert the `AddTagViewModel` to run as unit tests using a `TagsRepository` double (Fixed in d3a5980c7eaad1273fe417ffac4786a8c2d56ea0). I have also made changes corresponding specific changes you asked (see each comment for the commit).

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    fun close() {
+        _event.postValue(Event.CLOSE)
+    }
+
+    data class UiState(val tagName: String, val status: Status, val errorMsg: Int = -1)
+
+    enum class Status {
+        STARTED,
```
planarvoid: This is looking really good! I was just wondering if the `STARTED` and the `SAVING` shouldn't be Events. What I'm thinking is that they are not really exclusive with tag/error states, right? If we were able to check whether the keyboard is already showing, I'd say that there should be a field on the `UiState` called something like `isKeyboardShowing`. However, if we want to keep the current behaviour, the `show keyboard` and `hide keyboard` are more events sent in `onCreate`, not really a state. 

If we do this change, we can remove the status field from the `UiState` and simplify it.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    fun start() {
+        _uiState.value = UiState("", Status.STARTED)
+    }
+
+    fun close() {
+        _event.postValue(Event.CLOSE)
+    }
+
+    data class UiState(val tagName: String, val status: Status, val errorMsg: Int = -1)
```
planarvoid: I think I'd make the `errorMsg` a nullable Int and treat the `errorMsg == null` as the valid state. What do you think?

Simplenote/src/test/java/com/automattic/simplenote/repositories/FakeTagsRepository.kt
```diff
@@ -0,0 +1,35 @@
+package com.automattic.simplenote.repositories
+
+import com.automattic.simplenote.models.Tag
+import com.automattic.simplenote.utils.TagUtils
+
+class FakeTagsRepository : TagsRepository {
```
planarvoid: I think we should rather use a `mock` than a `fake`. This class contains a lot of logic and makes the test a little bit error prone. Let me know what you think 👍

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/AddTagViewModelTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        fakeTagsRepository.failAtSave = false
+
+        viewModel = AddTagViewModel(fakeTagsRepository)
+    }
+
+    @Test
+    fun validateEmptyTag() {
+        viewModel.updateUiState("")
+
+        //  assertEquals(viewModel.tagError.value, R.string.tag_error_empty)
```
planarvoid: I think this is not on purpose, right?

planarvoid (commented):
Thanks for the changes! It's looking much better! I have a few more comments to polish the things a little bit. Let me know what you think 👍

Simplenote/build.gradle
```diff
@@ -45,7 +45,10 @@ android {
     }
 
     testOptions {
-        unitTests.returnDefaultValues = true
+        unitTests {
+            includeAndroidResources = true
```
planarvoid: I guess this is not necessary any more, right?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
dandominguez04: Yeah, I think that change together with nullable of errorMessage can help remove the status field. Although, I need to figure out how not to call show keyboard everytime the UI state changes.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/AddTagViewModelTest.kt
dandominguez04: No, hehe. Fixed in edacb59763f915297662d0e7afc528c58a63511c.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
dandominguez04: Fixed in ee386d583f644a87bfff7074d9d38910947b955c.

Simplenote/build.gradle
dandominguez04: I guess not. We just have one test and it does not needed.

Simplenote/src/test/java/com/automattic/simplenote/repositories/FakeTagsRepository.kt
dandominguez04: Yeah, the fake is harder to maintain. Fixed in 7ca591ddf390341f3153a9cd791eca16c7a4925c.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
dandominguez04: Yeah, I changed to nullable and use that property to determine whether there is an error or not.

dandominguez04:
Hi @planarvoid, I have made some small changes regarding your comments.

Simplenote/src/androidTest/java/com/automattic/simplenote/BaseUITest.kt
```diff
@@ -49,13 +49,7 @@ open class BaseUITest {
...
 
     protected fun getRandomString(len: Int): String {
-        val random = SecureRandom()
-        val bytes = ByteArray(len)
-        random.nextBytes(bytes)
-
-        return (bytes.indices)
-                .map { charPool[random.nextInt(charPool.size)] }
-                .joinToString("")
+        return getRandomStringOfLen(len)
```
planarvoid, line 52: I know this was here before so let's just keep it. However, I don't think it's a good idea to have random values in tests (assuming it is random and not constant. I'm not sure if SecureRandom has a default seed or is indeed random).

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
planarvoid: I don't understand why we need the `isKeyboardShowing` field on the UiState. Why can't we just hide keyboard on `FINISH` or `SHOW_ERROR` events (that's the only place where it's set to `false` and that's where the dialog is closed)? Is there a state I'm missing where we need to keep the information about the keyboard being hidden?

planarvoid (commented):
Thanks for the changes! I have one last question/suggestion and I think after that we should be able to merge this PR 👍

Simplenote/src/androidTest/java/com/automattic/simplenote/BaseUITest.kt
dandominguez04: I agree that randomness can be a source of flakyness, but for the tests that this function is used (check length of input), the property of randomness on the content does not affect. There is not randomness on the size property. Randomness can be useful to create inputs for testing given they follow some properties. For example, property-based testing uses randomness to generate inputs but that follow some properties.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/AddTagViewModel.kt
dandominguez04: Yes, makes sense and make the UiState simpler. Fixed in c1ba4baf52392ce9b648329474a66c6f6fb9859e.

dandominguez04:
Thanks for all the discussion @planarvoid. I think the code is much simpler now thanks to the suggestions.

PR #1396, opened 2021-05-11
Add TagDialogViewModel to restructure UI and orchestration logic in TagDialogFragment

### LINKED ISSUE #1370: Add TagDialogViewModel to restructure UI and orchestration logic in TagDialogFragment
Parent: #1361

TODO

- Create `TagDialogViewModel`
- Restructure `TagDialogFragment` to use `TagDialogViewModel` to handle data management and state tasks

dandominguez04:
Fixes #1370 

### Fix
This PR restructure the code of `TagDialogFragment` to use ViewModels (`TagDialogViewModel`)  in order to decouple data management logic from the view. 

#### Automatic Tests
1. Run `TagDialogFragmentTest` inside `androidTest`. It requires to run an on emulator or physical device. 
2. Run `TagDialogViewModelTest inside `test`

#### Manual Tests
1. Open Simplenote app
2. Tab on left menu
3. Tab on EDIT tags
4. Tab on a tag that have notes associated
5. Try adding valid tags and tags with errors (e.g. tag with spaces, tags that already exists)
6. Try merging tags (rename a tag to a tag that already exists)

### Review
This PR will be reviewed by @ParaskP7 .

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
@@ -126,10 +126,7 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
                 dismiss();
             } else if (event instanceof TagDialogEvent.ShowErrorEvent) {
                 Context context = requireContext();
-                DialogUtils.showDialogWithEmail(
-                        context,
-                        context.getString(R.string.rename_tag_message)
-                );
+                DialogUtils.showDialogWithEmail(context, context.getString(R.string.rename_tag_message));
             } else if (event instanceof TagDialogEvent.ConflictEvent) {
                 TagDialogEvent.ConflictEvent conflictEvent = (TagDialogEvent.ConflictEvent) event;
                 showDialogErrorConflict(conflictEvent.getCanonicalTagName(), conflictEvent.getOldTagName());
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
@@ -23,10 +23,7 @@ import androidx.lifecycle.ViewModelProvider;
 import com.automattic.simplenote.models.Note;
 import com.automattic.simplenote.models.Tag;
 import com.automattic.simplenote.utils.DialogUtils;
-import com.automattic.simplenote.viewmodels.CloseEvent;
-import com.automattic.simplenote.viewmodels.ConflictEvent;
-import com.automattic.simplenote.viewmodels.FinishEvent;
-import com.automattic.simplenote.viewmodels.ShowErrorEvent;
+import com.automattic.simplenote.viewmodels.TagDialogEvent;
 import com.automattic.simplenote.viewmodels.TagDialogViewModel;
 import com.automattic.simplenote.viewmodels.ViewModelFactory;
 import com.google.android.material.textfield.TextInputEditText;
@@ -125,16 +122,16 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
         });
 
         viewModel.getEvent().observe(this, event -> {
-            if (event instanceof CloseEvent || event instanceof FinishEvent) {
+            if (event instanceof TagDialogEvent.CloseEvent || event instanceof TagDialogEvent.FinishEvent) {
                 dismiss();
-            } else if (event instanceof ShowErrorEvent) {
+            } else if (event instanceof TagDialogEvent.ShowErrorEvent) {
                 Context context = requireContext();
                 DialogUtils.showDialogWithEmail(
                         context,
                         context.getString(R.string.rename_tag_message)
                 );
-            } else if (event instanceof ConflictEvent) {
-                ConflictEvent conflictEvent = (ConflictEvent) event;
+            } else if (event instanceof TagDialogEvent.ConflictEvent) {
+                TagDialogEvent.ConflictEvent conflictEvent = (TagDialogEvent.ConflictEvent) event;
                 showDialogErrorConflict(conflictEvent.getCanonicalTagName(), conflictEvent.getOldTagName());
             }
         });
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
@@ -8,7 +8,6 @@ import android.content.res.Configuration;
 import android.os.Bundle;
 import android.text.Editable;
 import android.text.TextWatcher;
-import android.util.Log;
 import android.view.ContextThemeWrapper;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -16,18 +15,23 @@ import android.widget.Button;
 import android.widget.TextView;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatDialogFragment;
+import androidx.lifecycle.ViewModelProvider;
 
-import com.automattic.simplenote.analytics.AnalyticsTracker;
 import com.automattic.simplenote.models.Note;
 import com.automattic.simplenote.models.Tag;
 import com.automattic.simplenote.utils.DialogUtils;
-import com.automattic.simplenote.utils.TagUtils;
+import com.automattic.simplenote.viewmodels.CloseEvent;
+import com.automattic.simplenote.viewmodels.ConflictEvent;
+import com.automattic.simplenote.viewmodels.FinishEvent;
+import com.automattic.simplenote.viewmodels.ShowErrorEvent;
+import com.automattic.simplenote.viewmodels.TagDialogViewModel;
+import com.automattic.simplenote.viewmodels.ViewModelFactory;
 import com.google.android.material.textfield.TextInputEditText;
 import com.google.android.material.textfield.TextInputLayout;
 import com.simperium.client.Bucket;
-import com.simperium.client.BucketObjectNameInvalid;
 
 public class TagDialogFragment extends AppCompatDialogFragment implements TextWatcher, OnShowListener {
     public static String DIALOG_TAG = "dialog_tag";
@@ -38,7 +42,6 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
     private Button mButtonNegative;
     private Button mButtonNeutral;
     private Button mButtonPositive;
-    private String mTagOld;
     private Tag mTag;
     private TextInputEditText mEditTextTag;
     private TextInputLayout mEditTextLayout;
@@ -48,6 +51,8 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
     private View.OnClickListener mClickListenerPositiveConflict;
     private View.OnClickListener mClickListenerPositiveRename;
 
+    private TagDialogViewModel viewModel;
+
     public TagDialogFragment(Tag tag, Bucket<Note> bucketNote, Bucket<Tag> bucketTag) {
         mTag = tag;
         mBucketNote = bucketNote;
@@ -55,14 +60,31 @@ public class TagDialogFragment extends AppCompatDialogFragment implements TextWa
     }
 
     @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        ViewModelFactory viewModelFactory = new ViewModelFactory(mBucketTag, mBucketNote, this, null);
+        ViewModelProvider viewModelProvider = new ViewModelProvider(this, viewModelFactory);
+        viewModel = viewModelProvider.get(TagDialogViewModel.class);
+
+        viewModel.start(mTag);
+    }
+
+    @Override
     public void onConfigurationChanged(@NonNull Configuration newConfig) {
         super.onConfigurationChanged(newConfig);
-        dismiss();
+        viewModel.close();
     }
 
     @NonNull
     @Override
     public Dialog onCreateDialog(Bundle savedInstanceState) {
+        setupDialog();
+
+        return mDialogEditTag;
+    }
+
+    private void setupDialog() {
         Context context = new ContextThemeWrapper(requireContext(), R.style.Dialog);
         AlertDialog.Builder builder = new AlertDialog.Builder(context);
         builder.setTitle(R.string.rename_tag);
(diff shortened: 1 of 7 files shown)


Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+        _uiState.value = UiState(tag.name, tag)
+    }
+
+    fun close() {
+        _event.postValue(CloseEvent)
+    }
+
+    fun updateUiState(tagName: String) {
+        if (tagName.isEmpty()) {
+            _uiState.value = _uiState.value!!.copy(tagName = tagName, errorMsg = R.string.tag_error_empty)
```
ParaskP7: Minor (🔍): Consider replacing `!!` with `?`. I usually avoid the usage of `!!` unless it is absolutely necessary. Maybe it is in this case and I am not seeing it, is it?

PS: Same applies for tests.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        // If the tag did not changed, do not anything
+        if (currentState.tagName == currentState.oldTag.name) {
+            _event.postValue(FinishEvent)
+            return
+        }
+
+        // If there is another tag with the same name
+        if (!tagsRepository.isTagMissing(currentState.tagName)) {
```
ParaskP7: Question (❓): How is this the same with `if (hasCanonicalTag && !isRenamingToLexicalTag)`. Can you help me understand?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+            )
+        } else {
+            _event.postValue(ShowErrorEvent)
+        }
+    }
+
+    data class UiState(val tagName: String, val oldTag: Tag, val errorMsg: Int? = null)
+}
+
+sealed class TagDialogEvent
```
ParaskP7: Suggestion (💡): I suggest enclosing the events within the parent sealed class with braces (`{ ... }`), to make it stricter.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
```diff
@@ -81,80 +103,60 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
-            public void onClick(View v) {
-                String tagNew = getTagNew();
-                int index = mTag.hasIndex() ? mTag.getIndex() : mBucketTag.count();
-                tryToRenameTag(tagNew, index);
+        viewModel.getEvent().observe(this, event -> {
+            if (event instanceof CloseEvent || event instanceof FinishEvent) {
+                dismiss();
+            } else if (event instanceof ShowErrorEvent) {
+                Context context = requireContext();
+                DialogUtils.showDialogWithEmail(
```
ParaskP7: Minor (🔍): Consider making this a one-liner.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
```diff
@@ -81,80 +103,60 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
     }
 
     @Override
     public void onShow(DialogInterface dialog) {
         mButtonNegative = mDialogEditTag.getButton(DialogInterface.BUTTON_NEGATIVE);
         mButtonNeutral = mDialogEditTag.getButton(DialogInterface.BUTTON_NEUTRAL);
         mButtonPositive = mDialogEditTag.getButton(DialogInterface.BUTTON_POSITIVE);
+
+        // Set observers when views are available
+        setObservers();
```
ParaskP7, line 142: Question (❓): Why is `setObservers()` being trigger during `onShow(...)` instead of `onCreate(...)`, and more specifically before the call to `viewModel.start(mTag)`?

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/TagDialogViewModelTest.kt
```diff
@@ -0,0 +1,126 @@
...
+import com.automattic.simplenote.utils.getLocalRandomStringOfLen
+import org.junit.Assert.*
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.mockito.Mockito.`when`
+import org.mockito.Mockito.mock
+
+
+class TagDialogViewModelTest {
```
ParaskP7, line 16: Suggestion (💡): After I have completed the testing solution to the best of my abilities, I usually run the tests with coverage to see whether I missed testing something obvious. In this case the below were missed:
- The `close()` function is untested.
- The `if (currentState == null) { ... }` paths within the `renameTagIfValid()` and `renameTag()` functions are untested.

PS: I would also test the `renameTag()` function individually. I know this function is tested through the `renameTagIfValid()`, but consider what would happen if this parent calling function is, for example, removed at some point. Wdyt?

ParaskP7 (changes requested):
👋 @dandominguez04 !

The only thing that made it a bit hard for me to review was this single 8b3fda6ff641de9ddfe2e44e2459f677d742653e migration commit, which migrated the old logic to the new one in a single go. This commit utilised the previous 33dc0d835c3ca8740281d5e64cb613502fa846f2 view model commit, which already had the migration setup. If it was me, I would have done it more gradually, commit-by-commit, which would help me make sure I am not missing anything and the reviewer to be able to follow my train of thoughts. For example, I would split the commit as follows:
- Setup a skeleton `ViewModel` and its test class.
- Add this skeleton `ViewModel` to the `Fragment`.
- Add the new `ViewModel.start()` functionality and migrate accordingly, including tests.
- Add the new `ViewModel.close()` functionality and migrate accordingly, including tests.
- Add the new `ViewModel.updateUiState(...)` functionality and migrate old code accordingly, including tests.
- Add the new `ViewModel.renameTag()` functionality and migrated old code accordingly, including tests. Actually, in this case I would first move this whole old functionality to view model as is and then I'll start rewriting it to the new way. This is due to the complexity of it and to make sure I and the reviewer will not missing anything on the subsequent refactoring steps.
- Add the new `ViewModel.renameTagIfValid()` functionality and migrated old code accordingly, including tests. Actually, in this case I would first move this whole old functionality to view model as is and then I'll start rewriting it to the new way.  This is due to the complexity of it and to make sure I and the reviewer will not missing anything on the subsequent refactoring steps.

Wdyt, about the above, and splitting this kind of migration in multiple commits instead?

PS: I have also left a couple of minor (🔍) comments, some questions (❓) and suggestions (💡) for you to consider.

dandominguez04:
Hi @ParaskP7, thanks for the review. To answer some of your questions:

>  The only thing that made it a bit hard for me to review was this single 8b3fda6 migration commit, which migrated the old logic to the new one in a single go. This commit utilised the previous 33dc0d8 view model commit, which already had the migration setup. 

To be honest, I was a little bit blocked in the last PR since there was a little confusion about who was going to finish it. I started working on this PR some time before the last one to the point I had implement most of the changes. Thus, the commit history does not look organic since I did not want to mess it up like in the first two PRs I created. What do you recommend to handle such cases when there is some dependency on a PR that is blocking and you want to keep working on a new PR?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
dandominguez04: Yes,every time I use `!!`, I do not feel good. There should be an invariant that when `updateUiState` is called, `_uiState.value` is never `null`. However, the compiler does not see this case. I have added a defensive check for the invariant that shows an error in case the invariant is not satisfies and hints the type system that the value is not null (Fixed in 1c0ef7c43e46b831b05cc3f8f73f9c58c52d2f39).

I also fixed the tests in 47f23d7ce08e1746b19c32b4e052e9e5d6eb1533.

ParaskP7:
👋 @dandominguez04 !

> To be honest, I was a little bit blocked in the last PR since there was a little confusion about who was going to finish it. I started working on this PR some time before the last one to the point I had implement most of the changes. Thus, the commit history does not look organic since I did not want to mess it up like in the first two PRs I created. What do you recommend to handle such cases when there is some dependency on a PR that is blocking and you want to keep working on a new PR?

Yeap, this happens more often than I would have liked as well. But, I would either work with stash changes, a tmp branch or start the new solution on a commit-by-commit basis and rebase that later on before opening the new PR. It is always hard to distinguish where to do what and sometimes it doesn't end up working as expected.

In this cases, I would just start fresh, having whatever is in my hands and do it again from scratch. I actually, find out that doing it twice sometime helps me figure out problems with my initial try, like code reviewing my own self. Also, doing it for the second time will take less time, more ofter it is just about laying down the solution in a more structured way, filling out previously missed gaps.

In the end, what I try to avoid is to have the reviewer think harder than need be, trying to figure out if something was missed along the lines, matching the diffs mentally, and thus compromise their focus from the important bits of reviewing my code, where the value lies. Obviously, all that take more time from the author, but at the same time give time to the reviewer, which ends up being balanced with a positive effect, and the quality of the overall PR improved.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
ParaskP7: > Yes,every time I use !!, I do not feel good. 

Yes, I know, right! 😅 

>  There should be an invariant that when updateUiState is called, _uiState.value is never null. However, the compiler does not see this case. I have added a defensive check for the invariant that shows an error in case the invariant is not satisfies and hints the type system that the value is not null.

I like this, and it seems consistent with the rest.

> I also fixed the tests.

Actually, I think there is still one `!!` that can be converted to `?` in tests, is it not?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
dandominguez04: Yes, it does not cover all the behavior. `isTagMissing` covers `hasCanonicalTag` since is the same method as `hasCanonicalOfLexical` (just opposite answers), but does not cover `!isRenamingToLexicalTag`. `!isRenamingToLexicalTag` checks for the case in which the tag is not identical the current one but the lexical tag is the same. This would show the merge dialog unnecessarily. Fixed in 7bf0745b86e4630215e26b79498bc0b90ba0f196 (added `isTagConflict` to `TagsRepository`) and bbfc4c1f7e8385de71a4f9e08f6655d173824141 (used `isTagConflict` in ViewModel and fixed tests).

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
dandominguez04: I agree. Fixed in 573355dea3fbd6fe43e5718faa0da0fc7129c67e.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
dandominguez04: Fixed in b112ade6ed25423db196d9bdf6f5c0b33f60979d.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
dandominguez04: In `onCreate(...)` some views that are specific to the dialog (e,g, positive, negative buttons) are not available (they return `null` if we want to get them). The observers use these views specifically, therefore it would throw a null pointer exception. What I did change was the call to `viewModelo.start(...)`, I put it after I call the observers. Fixed in ce616f50086d096375c65e1db6cc66a47bb98ac2.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/TagDialogViewModelTest.kt
dandominguez04: Yes, those were missing tests. Added tests for `close` and `renameTag`. Fixed in e025a750967a98e21f9288b952c826817e6e67ac.

dandominguez04:
> In the end, what I try to avoid is to have the reviewer think harder than need be, trying to figure out if something was missed along the lines, matching the diffs mentally, and thus compromise their focus from the important bits of reviewing my code, where the value lies. Obviously, all that take more time from the author, but at the same time give time to the reviewer, which ends up being balanced with a positive effect, and the quality of the overall PR improved.

Thanks @ParaskP7  for the feedback. I will consider this to improve my workflow and make it easier for my coworkers. 

I have made some changes related to your comments. Let me know what do you think?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
ParaskP7: The extra `!!` was fixed in bbfc4c1f7e8385de71a4f9e08f6655d173824141. 🙏

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.java
ParaskP7: Brilliant, thanks for doing so @dandominguez04 !

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
ParaskP7: > isTagMissing covers hasCanonicalTag since is the same method as hasCanonicalOfLexical (just opposite answers).

This is true!

> ...but does not cover !isRenamingToLexicalTag. !isRenamingToLexicalTag checks for the case in which the tag is not identical the current one but the lexical tag is the same. This would show the merge dialog unnecessarily.

Nice finding!

> Fixed in 7bf0745 (added isTagConflict to TagsRepository) and bbfc4c1 (used isTagConflict in ViewModel and fixed tests).

🚀

ParaskP7: PS: I think missing this behavior changes is very easy when not doing these kind of refactoring, migration, modernization steps commit-by-commit, guarding those commits with tests, even TDD wise if possible. Thus, you are seeing me pushing towards this direction a bit too enthusiastically. 😅

ParaskP7: 👋 @dandominguez04 !

> isTagMissing covers hasCanonicalTag since is the same method as hasCanonicalOfLexical (just opposite answers).

About that, I still wonder why the initial author choose to use the `hasCanonicalOfLexical(...)` instead of `isTagMissing(...)`. Do you think the `isTagMissing(...)` is better suited here, thus you choose to change it or?

ParaskP7 (approved):
👋 @dandominguez04 !

I did another pass on your changes and resolved everything but one discussion, just so I get the change to discuss it will you. I am approving this PR nevertheless. Good job! 🌟

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagDialogViewModel.kt
dandominguez04: HI @ParaskP7 !

> PS: I think missing this behavior changes is very easy when not doing these kind of refactoring, migration, modernization steps commit-by-commit, guarding those commits with tests, even TDD wise if possible. Thus, you are seeing me pushing towards this direction a bit too enthusiastically.

Yes, I think the UI tests are missing some tests regarding encoding of tags and the canonical form of the tags. 

> About that, I still wonder why the initial author choose to use the hasCanonicalOfLexical(...) instead of isTagMissing(...). Do you think the isTagMissing(...) is better suited here, thus you choose to change it or?

Maybe the author chose that method to be more explicit on naming about the canonical form of the tag. I chose `isTagMissing(...)` because I did not want to blow the Repository with multiple methods that have the same behavior. I think `isTagMissing(...)` does the job here because we are looking to see if there another tag. The same method is used for validation when adding a new tag, thus, we have uniform behavior across screens.

ParaskP7: 👋 @dandominguez04 !

Thanks for your responses.

> Yes, I think the UI tests are missing some tests regarding encoding of tags and the canonical form of the tags.

True, but I also think that this might be tricky to test anyway.

> Maybe the author chose that method to be more explicit on naming about the canonical form of the tag. I chose isTagMissing(...) because I did not want to blow the Repository with multiple methods that have the same behavior. I think isTagMissing(...) does the job here because we are looking to see if there another tag. The same method is used for validation when adding a new tag, thus, we have uniform behavior across screens.

I wouldn't worry about blowing the `Repository`, it is anyway not that big atm, especially if the right naming is used. However, you other arguments might suffice for now, so I am okay to keeping it as is for now. In the future, we might merge those two similar utility methods into a unified one, with a more appropriate naming, or choose to rollback to using the original utility method.

But, for now we are okay, thanks for taking the time to think about it. 🙏

PR #1397, opened 2021-05-14
Add TagsViewModel to restructure UI and orchestration logic in TagsActivity

### LINKED ISSUE #1371: Add TagViewModel to restructure UI and orchestration logic in TagActivity
Parent: #1361

dandominguez04:
Fixes #1371

### Fix
This PR restructure the code of `TagsActivity` to use ViewModels (`TagsViewModel`)  in order to decouple data management logic from the view. 

#### Automatic Tests
1. Run `TagsActivityTest` inside `androidTest`. It requires to run an on emulator or physical device. 
2. Run `TagsViewModelTest` inside `test`

#### Manual Tests
1. Open Simplenote app
2. Tab on left menu
3. Tab on EDIT tags
4. Try deleting tags without and with notes. It should update the list
5. Try tab on tags to open a dialog to update the tag
6. Try tab on + button to add new tags
7. Try searching tags

### Review
This PR will be reviewed by @ParaskP7 .

### Release

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
@@ -284,7 +284,7 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
     @Override
     protected void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
         if (resultCode == RESULT_OK && requestCode == REQUEST_ADD_TAG) {
-            refreshTags();
+            viewModel.updateOnResult();
         }
 
         super.onActivityResult(requestCode, resultCode, data);
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
@@ -132,6 +132,8 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
                 }
 
                 Toast.makeText(TagsActivity.this, getString(R.string.add_tag), Toast.LENGTH_SHORT).show();
+            } else if (event instanceof TagsEvent.FinishEvent) {
+                finish();
             }
         });
     }
@@ -294,7 +296,7 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
         if (item.getItemId() == android.R.id.home) {
-            finish();
+            viewModel.close();
             return true;
         }
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
@@ -72,7 +72,6 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
     private ImageView mEmptyViewImage;
     private MenuItem mSearchMenuItem;
     private String mSearchQuery;
-    private TagsAdapter mTagsAdapter;
     private TextView mEmptyViewText;
     private boolean mIsSearching;
 
@@ -117,7 +116,6 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
         }
 
         mTagsList = findViewById(R.id.list);
-        mTagsAdapter = new TagsAdapter();
         mTagsList.setAdapter(tagItemAdapter);
         mTagsList.setLayoutManager(new LinearLayoutManager(TagsActivity.this));
         View emptyView = findViewById(R.id.empty);
@@ -317,7 +315,7 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
     protected void refreshTags() {
         Query<Tag> tagQuery = Tag.all(mTagsBucket).reorder().orderByKey().include(Tag.NOTE_COUNT_INDEX_NAME);
         Bucket.ObjectCursor<Tag> cursor = tagQuery.execute();
-        mTagsAdapter.swapCursor(cursor);
+      //  mTagsAdapter.swapCursor(cursor);
     }
 
     protected void refreshTagsSearch() {
@@ -326,7 +324,7 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
             .orderByKey().include(Tag.NOTE_COUNT_INDEX_NAME)
             .reorder();
         Bucket.ObjectCursor<Tag> cursor = tags.execute();
-        mTagsAdapter.swapCursor(cursor);
+//        mTagsAdapter.swapCursor(cursor);
     }
 
     private void setEmptyListImage(@DrawableRes int image) {
@@ -416,155 +414,5 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
     @Override
     public void onSyncObject(Bucket<Tag> bucket, String key) {
 
-    }
-
-    private static class RemoveTagFromNotesTask extends AsyncTask<Tag, Void, Void> {
-        private SoftReference<TagsActivity> mTagsActivityReference;
-
-        private RemoveTagFromNotesTask(TagsActivity context) {
-            mTagsActivityReference = new SoftReference<>(context);
-        }
-
-        @Override
-        protected Void doInBackground(Tag... removedTags) {
-            TagsActivity activity = mTagsActivityReference.get();
-            Tag tag = removedTags[0];
-
-            if (tag != null) {
-                Bucket.ObjectCursor<Note> cursor = tag.findNotes(activity.mNotesBucket, tag.getName());
-
-                while (cursor.moveToNext()) {
-                    Note note = cursor.getObject();
-                    List<String> tags = note.getTags();
-                    tags.remove(tag.getName());
-                    note.setTags(tags);
-                    note.save();
-                }
-
-                cursor.close();
-            }
-
-            return null;
-        }
-    }
-
-    private class TagsAdapter extends BaseCursorAdapter<TagsAdapter.ViewHolder> {
-        public class ViewHolder extends RecyclerView.ViewHolder implements View.OnClickListener {
-            private TextView mCount;
-            private TextView mTitle;
-
-            private ViewHolder(View itemView) {
-                super(itemView);
-
-                mTitle = itemView.findViewById(R.id.tag_name);
-                mCount = itemView.findViewById(R.id.tag_count);
-
-                ImageButton deleteButton = itemView.findViewById(R.id.tag_trash);
-                deleteButton.setOnClickListener(
-                    new View.OnClickListener() {
-                        @Override
-                        public void onClick(View view) {
-                            if (!hasItem(getAdapterPosition())) {
-                                return;
-                            }
-
-                            @SuppressWarnings("unchecked")
-                            final Tag tag = ((Bucket.ObjectCursor<Tag>) getItem(getAdapterPosition())).getObject();
-                            final int tagCount = mNotesBucket.query().where(TAGS_PROPERTY, Query.ComparisonType.EQUAL_TO, tag.getName()).count();
-
-                            if (tagCount == 0) {
-                                deleteTag(tag);
-                            } else if (tagCount > 0) {
-                                AlertDialog.Builder alert = new AlertDialog.Builder(new ContextThemeWrapper(TagsActivity.this, R.style.Dialog));
-                                alert.setTitle(R.string.delete_tag);
-                                alert.setMessage(getString(R.string.confirm_delete_tag));
-                                alert.setNegativeButton(R.string.no, null);
-                                alert.setPositiveButton(
-                                    R.string.yes,
-                                    new DialogInterface.OnClickListener() {
-                                        public void onClick(DialogInterface dialog, int whichButton) {
-                                            deleteTag(tag);
-                                        }
-                                    }
-                                );
-                                alert.show();
-                            }
-                        }
-                    }
-                );
-                deleteButton.setOnLongClickListener(
-                    new View.OnLongClickListener() {
-                        @Override
-                        public boolean onLongClick(View v) {
-                            if (v.isHapticFeedbackEnabled()) {
-                                v.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS);
-                            }
-
-                            Toast.makeText(TagsActivity.this, getString(R.string.delete_tag), Toast.LENGTH_SHORT).show();
-                            return true;
-                        }
-                    }
-                );
-
-                itemView.setOnClickListener(this);
-            }
-
-            @Override
-            public void onClick(View view) {
-                if (!hasItem(getAdapterPosition())) {
-                    return;
-                }
-
-                //noinspection unchecked
-                TagDialogFragment dialog = new TagDialogFragment(
-                    ((Bucket.ObjectCursor<Tag>) getItem(getAdapterPosition())).getObject(),
-                    mNotesBucket,
-                    mTagsBucket
-                );
-                dialog.show(getSupportFragmentManager().beginTransaction(), DIALOG_TAG);
-            }
-
-            private void deleteTag(Tag tag) {
-                tag.delete();
-                new RemoveTagFromNotesTask(TagsActivity.this).executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, tag);
-                AnalyticsTracker.track(
-                    AnalyticsTracker.Stat.TAG_MENU_DELETED,
-                    AnalyticsTracker.CATEGORY_TAG,
-                    "list_trash_button"
-                );
-            }
-        }
-
-        private TagsAdapter() {
-            super(null);
-        }
-
-        @NonNull
-        @Override
-        public TagsAdapter.ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
-            Context context = parent.getContext();
-            LayoutInflater inflater = LayoutInflater.from(context);
-            View contactView = inflater.inflate(R.layout.tags_list_row, parent, false);
-            return new TagsAdapter.ViewHolder(contactView);
-        }
-
-        @Override
-        public void onBindViewHolder(@NonNull TagsAdapter.ViewHolder holder, Cursor cursor) {
-            @SuppressWarnings("unchecked")
-            Tag tag = ((Bucket.ObjectCursor<Tag>)cursor).getObject();
-            holder.mTitle.setText(tag.getName());
-            final int tagCount = mNotesBucket.query().where(TAGS_PROPERTY, Query.ComparisonType.EQUAL_TO, tag.getName()).count();
-
-            if (tagCount > 0) {
-                holder.mCount.setText(String.valueOf(tagCount));
-            } else {
-                holder.mCount.setText("");
-            }
-        }
-
-        @Override
-        public void swapCursor(Cursor newCursor) {
-            super.swapCursor(newCursor);
-        }
     }
 }
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
@@ -42,6 +42,7 @@ import com.automattic.simplenote.utils.DrawableUtils;
 import com.automattic.simplenote.utils.HtmlCompat;
 import com.automattic.simplenote.utils.ThemeUtils;
 import com.automattic.simplenote.viewmodels.AddTagViewModel;
+import com.automattic.simplenote.viewmodels.TagsEvent;
 import com.automattic.simplenote.viewmodels.TagsViewModel;
 import com.automattic.simplenote.viewmodels.ViewModelFactory;
 import com.automattic.simplenote.widgets.EmptyViewRecyclerView;
@@ -107,18 +108,8 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
         mTagsList.setEmptyView(emptyView);
 
         mButtonAdd = findViewById(R.id.button_add);
-        mButtonAdd.setOnClickListener(
-            new View.OnClickListener() {
-                @Override
-                public void onClick(View v) {
-                    Intent intent = new Intent(TagsActivity.this, AddTagActivity.class);
-                    intent.putExtra(MorphSetup.EXTRA_SHARED_ELEMENT_COLOR_END, ThemeUtils.getColorFromAttribute(TagsActivity.this, R.attr.drawerBackgroundColor));
-                    intent.putExtra(MorphSetup.EXTRA_SHARED_ELEMENT_COLOR_START, ThemeUtils.getColorFromAttribute(TagsActivity.this, R.attr.fabColor));
-                    ActivityOptions options = ActivityOptions.makeSceneTransitionAnimation(TagsActivity.this, mButtonAdd, "shared_button");
-                    startActivityForResult(intent, REQUEST_ADD_TAG, options.toBundle());
-                }
-            }
-        );
+        mButtonAdd.setOnClickListener(v -> viewModel.addTag());
+
         mButtonAdd.setOnLongClickListener(
             new View.OnLongClickListener() {
                 @Override
@@ -133,7 +124,21 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
             }
         );
 
+        setObservers();
+
         refreshTags();
+    }
+
+    private void setObservers() {
+        viewModel.getEvent().observe(this, event -> {
+            if (event instanceof TagsEvent.AddTagEvent) {
+                Intent intent = new Intent(TagsActivity.this, AddTagActivity.class);
+                intent.putExtra(MorphSetup.EXTRA_SHARED_ELEMENT_COLOR_END, ThemeUtils.getColorFromAttribute(TagsActivity.this, R.attr.drawerBackgroundColor));
+                intent.putExtra(MorphSetup.EXTRA_SHARED_ELEMENT_COLOR_START, ThemeUtils.getColorFromAttribute(TagsActivity.this, R.attr.fabColor));
+                ActivityOptions options = ActivityOptions.makeSceneTransitionAnimation(TagsActivity.this, mButtonAdd, "shared_button");
+                startActivityForResult(intent, REQUEST_ADD_TAG, options.toBundle());
+            }
+        });
     }
 
     @Override
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
@@ -138,13 +138,14 @@ public class TagsActivity extends ThemedAppCompatActivity {
 
     private void setObservers() {
         viewModel.getUiState().observe(this, uiState ->
-            tagItemAdapter.submitList(uiState.getTagItems(), () -> {
-                if (uiState.getSearchUpdate()) {
-                    mTagsList.scrollToPosition(0);
-                    boolean isSearching = uiState.getSearchQuery() != null;
-                    checkEmptyList(isSearching);
-            }
-        }));
+                tagItemAdapter.submitList(uiState.getTagItems(), () -> {
+                    if (uiState.getSearchUpdate()) {
+                        mTagsList.scrollToPosition(0);
+                        boolean isSearching = uiState.getSearchQuery() != null;
+                        checkEmptyList(isSearching);
+                    }
+                })
+        );
 
         // Observe different events such as clicks on add tags, edit tags and delete tags
         viewModel.getEvent().observe(this, event -> {
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
@@ -41,7 +41,6 @@ import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.DrawableUtils;
 import com.automattic.simplenote.utils.HtmlCompat;
 import com.automattic.simplenote.utils.ThemeUtils;
-import com.automattic.simplenote.viewmodels.AddTagViewModel;
 import com.automattic.simplenote.viewmodels.TagsEvent;
 import com.automattic.simplenote.viewmodels.TagsViewModel;
 import com.automattic.simplenote.viewmodels.ViewModelFactory;
@@ -108,21 +107,11 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
         mTagsList.setEmptyView(emptyView);
 
         mButtonAdd = findViewById(R.id.button_add);
-        mButtonAdd.setOnClickListener(v -> viewModel.addTag());
-
-        mButtonAdd.setOnLongClickListener(
-            new View.OnLongClickListener() {
-                @Override
-                public boolean onLongClick(View v) {
-                    if (v.isHapticFeedbackEnabled()) {
-                        v.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS);
-                    }
-
-                    Toast.makeText(TagsActivity.this, getString(R.string.add_tag), Toast.LENGTH_SHORT).show();
-                    return true;
-                }
-            }
-        );
+        mButtonAdd.setOnClickListener(v -> viewModel.clickAddTag());
+        mButtonAdd.setOnLongClickListener(v -> {
+            viewModel.longClickAddTag();
+            return true;
+        });
 
         setObservers();
 
@@ -137,6 +126,12 @@ public class TagsActivity extends ThemedAppCompatActivity implements Bucket.List
                 intent.putExtra(MorphSetup.EXTRA_SHARED_ELEMENT_COLOR_START, ThemeUtils.getColorFromAttribute(TagsActivity.this, R.attr.fabColor));
                 ActivityOptions options = ActivityOptions.makeSceneTransitionAnimation(TagsActivity.this, mButtonAdd, "shared_button");
                 startActivityForResult(intent, REQUEST_ADD_TAG, options.toBundle());
+            } else if (event instanceof TagsEvent.LongAddTagEvent) {
+                if (mButtonAdd.isHapticFeedbackEnabled()) {
+                    mButtonAdd.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS);
+                }
+
+                Toast.makeText(TagsActivity.this, getString(R.string.add_tag), Toast.LENGTH_SHORT).show();
             }
         });
     }
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
@@ -194,9 +194,8 @@ public class TagsActivity extends ThemedAppCompatActivity {
     }
 
     private void showTagDialogFragment(TagsEvent.EditTagEvent event) {
-        TagsEvent.EditTagEvent editTagEvent = event;
         TagDialogFragment dialog = new TagDialogFragment(
-                editTagEvent.getTagItem().getTag(),
+                event.getTagItem().getTag(),
                 mNotesBucket,
                 mTagsBucket
         );
--- a/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
@@ -35,11 +35,13 @@ import androidx.recyclerview.widget.RecyclerView;
 import com.automattic.simplenote.analytics.AnalyticsTracker;
 import com.automattic.simplenote.models.Note;
 import com.automattic.simplenote.models.Tag;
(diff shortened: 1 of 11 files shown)


Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+    val uiState: LiveData<UiState> = _uiState
+
+    private val _event = SingleLiveEvent<TagsEvent>()
+    val event: LiveData<TagsEvent> = _event
+
+    private var jobTagsFlow: Job? = null
+
+    fun start() {
+        viewModelScope.launch {
+            val tagItems = withContext(Dispatchers.IO) { tagsRepository.allTags() }
```
ParaskP7: Suggestion (💡): This is quite the debatable opinion, so please take it with a grain of salt, I just want to listen your opinion as well. So, here I go. If it was me I would have moved all the context switching, aka `withContext(...)`, downstream and in this case into the repository.

This kind of pattern helps the caller of the API, in this case the caller of the repository to not have to deal with the internals of where is this piece of logic need to run. Instead the API designer should make sure that this is run with the most appropriate dispatcher, in this case the IO, thus making this case move from the main thread and into a background one.

By doing it this way you make it easier to the `ViewModel`, which will now not have to worry about any context switching. This way, the tests also benefit from not having to worry about it, thus making them more deterministic. For example, the bad practice of `Thread.sleep(...)` in unit tests, in this case, can be totally avoided. Obviously, sometimes you need to include, by design, context switching in your `ViewModel` and then you indeed need to write your tests in such a way to switch the main dispatchers with test dispatcher and avoid `Thread.sleep(...)` as well, but in this case, I think it is even easier because this is not even needed if those are moved to the `Repository` level. Wdyt?

Also, this for example, will remove the need to include `withContext(Dispatchers.IO)` every time the `tagsRepository.allTags()` is called, thus removing some duplication. The author of the `ViewModel` actually gets freed up, cognitively, into having to think about all that.

Let me know your thoughts and let's discuss. 🙏

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+            _uiState.value = UiState(tagItems)
+        }
+
+        jobTagsFlow = viewModelScope.launch {
+            tagsRepository.tagsChanged().collect {
```
ParaskP7, line 33: 1) Warning (⚠️): Consider adding tests for this `collect { ... }` logic.
2) Question (❓): Wouldn't the flow need to run on another context as well?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+        _uiState.value = _uiState.value?.copy(searchUpdate = true, searchQuery = null)
+    }
+
+    fun search(searchQuery: String) {
+        viewModelScope.launch {
+            updateUiState(searchQuery, true)
+        }
+    }
+
+    fun pause() {
```
ParaskP7, line 69: 1) Warning (⚠️): Consider adding tests for this function.
2) Question (❓): Do you think this comment is necessary?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    fun clickEditTag(tagItem: TagItem) {
+        _event.postValue(TagsEvent.EditTagEvent(tagItem))
+    }
+
+    fun clickDeleteTag(tagItem: TagItem) {
+        if (tagItem.noteCount > 0) {
+            _event.postValue(TagsEvent.DeleteTagEvent(tagItem))
+        } else {
+            deleteTag(tagItem)
```
ParaskP7, line 90: Warning (⚠️): Consider adding tests for this `deleteTag(...)` branch logic.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    fun longClickDeleteTag(view: View) {
+        _event.postValue(TagsEvent.LongDeleteTagEvent(view))
+    }
+
+    fun deleteTag(tagItem: TagItem) {
+        viewModelScope.launch(Dispatchers.IO) {
```
ParaskP7: Question (❓): Above, you choose to use `withContext(Dispatchers.IO) { ... }` instead of `viewModelScope.launch(Dispatchers.IO) { ... }`, but here you are doing it this way, why did you choose one over the other?

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
```diff
@@ -82,48 +119,95 @@ public void onCreate(Bundle savedInstanceState) {
...
+
+    private void setObservers() {
+        viewModel.getUiState().observe(this, uiState ->
+            tagItemAdapter.submitList(uiState.getTagItems(), () -> {
+                if (uiState.getSearchUpdate()) {
+                    mTagsList.scrollToPosition(0);
+                    boolean isSearching = uiState.getSearchQuery() != null;
+                    checkEmptyList(isSearching);
+            }
+        }));
```
ParaskP7: Minor (🔍): Consider reformatting this to:
```
        viewModel.getUiState().observe(this, uiState ->
                tagItemAdapter.submitList(uiState.getTagItems(), () -> {
                    if (uiState.getSearchUpdate()) {
                        mTagsList.scrollToPosition(0);
                        boolean isSearching = uiState.getSearchQuery() != null;
                        checkEmptyList(isSearching);
                    }
                })
        );
```

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/TagsViewModelTest.kt
```diff
@@ -0,0 +1,232 @@
...
+    private val fakeTagsRepository = mock(TagsRepository::class.java)
+    private val tagItems = mutableListOf<TagItem>()
+
+    @Before
+    fun setup() {
+        viewModel = TagsViewModel(fakeTagsRepository)
+    }
+
+    @Before
+    fun setupTags() {
```
ParaskP7: 1) Minor (🔍): Why using two `@Before` annotation instead having one entry point to it and then using inner calls to split setup into separate functions? To be honest, I didn't know that you can do something like that. 😅 Is that even deterministic, which one will run first, does that depend on the order of the defined annotated functions?
2) Question (❓): Why did you choose to setup the `tagItems` every time a test runs? For example, if it was me I would have move that to field declaration and made this list immutable as well, like:
```
    private val tagItems = listOf(
        TagItem(Tag("tag1"), 0),
        TagItem(Tag("tag2"), 2),
        TagItem(Tag("tag3"), 5),
        TagItem(Tag("tag4"), 10),
        TagItem(Tag("tag5"), 0),
    )
```
I understand that you need that to be mutable, just in this one test, and in that case I would just create another mutable inner to this single test variable that would contain all these `tagItems` values, but open it to mutability. Wdyt?

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
```diff
@@ -48,4 +55,71 @@ class SimperiumTagsRepository(
...
+            tags.remove(tag.name)
+            note.tags = tags
+            note.save()
+        }
+
+        cursor.close()
+    }
+
+    @ExperimentalCoroutinesApi
+    override suspend fun tagsChanged(): Flow<Boolean> = callbackFlow {
```
ParaskP7, line 79: Praise (❤️): This is a super interesting solution, using the `callbackFlow { ... }` trick like that, kudos. The only thing that would made me feel a bit uncomfortable is the fact that this is still an experimental API, but other than that, nice job.

ParaskP7 (changes requested):
👋 @dandominguez04  !

I have reviewed and tested this PR as per the instructions, everything works as expected, really nice work here as well, it is very impressive to see the diff here, the before and after, so much complexity has been reduced! 🌟 🌟 🌟 

I have left some warning (⚠️) comments, which are mainly test related, then some minor (🔍) ones as well, few suggestions (💡) and a couple of questions (❓) that I want to discuss with you. In addition to those, I have the below I would like to briefly discuss:
- You did a good job splitting the commits like you did. However, please try to avoid combining migration commits with pure refactor commits. It becomes harder to focus on the migration part when an extra refactor commit get squashed into it. Things can split away this way, don't you think?
- Although you did a good job splitting the commits, I would have loved if that was done feature-by-feature, followed by tests, or even better proceeded by tests. Otherwise it is very hard to follow the changes and I found myself trying to clue commits together in order to make sure I understood the full migration. Adding tests at the end do provide value, but adding tests as you progress with the migration makes it easier to follow, at least from the reviewer's point of view. Wdyt?
- Both the above made it really hard for me to follow the adapter migration. You did an excellent job with this migration, but the committing strategy didn't help me figure out how that was done. I was just presented with the result rather than the path to the migration. Then, I had to work it out myself, to make sure that nothing is missing and I still end up feeling that I might missed bits here and there, even after I was done reviewing this PR. Wdyt?
- Why did you choose the `org.mockito.kotlin` library? I am asking that because in all our other repos, or at least the `WPAndroid` one, we are using the `com.nhaarman.mockitokotlin2` instead. I would prefer to keep this consistent, but I am definitely not opposed to using a different library if that is for the best. For example, I would have personally used `io.mockk` for that matter, but I would have first explain the reasoning behind that, else I would have gone with the flow, using `com.nhaarman.mockitokotlin2` as a starting point. Wdyt?

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
```diff
@@ -48,4 +55,71 @@ class SimperiumTagsRepository(
...
+        tag.delete()
+    }
+
+    private fun deleteTagFromNotes(tag: Tag) {
+        val cursor = tag.findNotes(notesBucket, tag.name)
+
+        while (cursor.moveToNext()) {
+            val note = cursor.getObject()
+            val tags = note.tags
+            tags.remove(tag.name)
```
dmsnell: we need to keep in mind the distinction here between a tag's name and its `tagHash`

tags are stored in a note's model with the name, of which there might be multiple "lexical variations" that collapse to the same reference, or hash. for example, tags with different casings are identical as are tags whose only difference is their Unicode normal forms.

it would be nice, especially since we're jumping in between notes and tags here, coupling the two, to add a new method to `Note`, which would remove a tag and maintain all the appropriate logic.

```kt
while (cursor.moveToNext()) {
    var note = cursor.getObject()
    note.removeTag(tag.name)
    note.save()
}
```

```kt
// Note
fun removeTag(tag: Tag) {
    tags = tags.filter { name -> ! tag.sameAs(name) }
}
```

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
```diff
@@ -48,4 +55,71 @@ class SimperiumTagsRepository(
...
+    override suspend fun allTags(): List<TagItem> {
+        val tagQuery = Tag.all(tagsBucket).reorder().orderByKey().include(Tag.NOTE_COUNT_INDEX_NAME)
+        val cursor = tagQuery.execute()
+
+        return cursorToTagItems(cursor)
+    }
+
+    override suspend fun searchTags(query: String): List<TagItem> {
+        val tags = Tag.all(tagsBucket)
+                .where(Tag.NAME_PROPERTY, Query.ComparisonType.LIKE, "%$query%")
```
dmsnell, line 106: note that here too we will want to compare not the tag name but the tag hash. the `name` for the tag in the Tag bucket might be a different representation than the name given in the query even though they are the same tag. if we don't do this then we'll miss tags in our searches that should appear.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/TagsViewModelTest.kt
```diff
@@ -0,0 +1,232 @@
...
+    fun clickDeleteTagOfTagWithNotesShouldDeleteTagDirectly() = runBlockingTest {
+        fakeTagsRepository.stub {
+            onBlocking { deleteTag(any()) }.doReturn(Unit)
+        }
+
+        fakeTagsRepository.stub {
+            onBlocking { tagsChanged() }.doReturn(flow{ emit(true) })
+        }
+
+        val updatedTags = tagItems.filter { tagItem -> tagItem.tag.name != "tag2" }
```
dmsnell: this works in the tests because we're using lower-cased ASCII/latin1 tag names, but we should rely on the tag hashing mechanism to avoid hiding regressions in the tag comparison logic.

we should be able to click on a tag shown as `Tag2` and find that the `tag2` tag is also gone. this happens regularly with Simplenote due to how notes reference tags by name and not by reference.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/TagsViewModelTest.kt
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        viewModel.clickDeleteTag(tagItems[1])
+
+        fakeTagsRepository.stub {
+            onBlocking { allTags() }.doReturn(updatedTags)
+        }
+        viewModel.updateOnResult()
+
+        // Flay tests regarding runBlockingTest. I will include this sleep here for now
```
dmsnell: typo: Flay -> Flaky?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
dandominguez04: I completely agree with this point. It makes the ViewModel much cleaner and moves the responsibility of context switching where it should be. It also fixed our tests.

Fixed in 87b060c762f390a5421c6821e8c7dc96751c2621.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/TagsViewModelTest.kt
dandominguez04: Yes, typo. But we already removed the sleep call since we fixed the tests in 87b060c762f390a5421c6821e8c7dc96751c2621.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
dandominguez04: If we follow the pattern you described above, then the responsibility of changing the context to IO should be of the callee, not the caller. I will change the `launch` call to use the default dispatcher and change the context in the repo as we did with the `allTags` and `searchTags` methods. 

Fixed in cf6f8ce802019fbe99b0d1a09bc0729e3b1f4b68.

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
dandominguez04: Fixed in 165d851d1a7900f7406fa715f7bbd07a58967aea.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
ParaskP7: Great, thanks for making this change @dandominguez04 ! 🚀

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
ParaskP7: Thanks for the change @dandominguez04 ! 🚀

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
```diff
@@ -82,48 +119,96 @@ public void onCreate(Bundle savedInstanceState) {
...
         mTagsList = findViewById(R.id.list);
-        mTagsAdapter = new TagsAdapter();
-        mTagsList.setAdapter(mTagsAdapter);
+        mTagsList.setAdapter(tagItemAdapter);
         mTagsList.setLayoutManager(new LinearLayoutManager(TagsActivity.this));
         View emptyView = findViewById(R.id.empty);
         mEmptyViewImage = emptyView.findViewById(R.id.image);
         mEmptyViewText = emptyView.findViewById(R.id.text);
-        checkEmptyList();
+        checkEmptyList(false);
```
dmsnell: _don't_ check empty list?

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
```diff
@@ -82,48 +119,96 @@ public void onCreate(Bundle savedInstanceState) {
...
             }
+        });
+    }
+
+    private void showLongAddToast() {
+        if (mButtonAdd.isHapticFeedbackEnabled()) {
+            mButtonAdd.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS);
+        }
+
+        Toast.makeText(TagsActivity.this, getString(R.string.add_tag), Toast.LENGTH_SHORT).show();
```
dmsnell: in the past I found it helpful to add a class method for toasting. the toasts were all similar enough that they mostly just required passing the string. possibly the length.

```java
private void showLongAddToast() {
    if (mButtonAdd…) {
        …
    }

    toast(R.string.add_tag);
}

private void toast(@StringRes int resId) {
    toast(resId, Toast.LENGTH_SHORT);
}

private void toast(@StringReg int resId, @Duration int duration) {
    Toast.makeText(TagsActivity.this, resId, duration).show();
}
```

besides removing duplicated code, I felt like this baked it down to the essence of what we're doing and relegated the details to the implementation

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
```diff
@@ -48,4 +57,71 @@ class SimperiumTagsRepository(
...
+        return@withContext cursorToTagItems(cursor)
+    }
+
+    private fun cursorToTagItems(cursor: Bucket.ObjectCursor<Tag>): List<TagItem> {
+        return (1 .. cursor.count).map {
+            cursor.moveToNext()
+            val tag = cursor.`object`
+            val noteCount: Int = notesBucket
+                    .query()
+                    .where(Note.TAGS_PROPERTY, Query.ComparisonType.EQUAL_TO, tag.name)
```
dmsnell, line 120: unfortunately this is like the other places and we probably can't use the database query directly. if we don't compare tag hashes we will get the wrong counts. this has been an annoying bug in the app for a long time, so it's not like it's new, but also, we should avoid introducing new regressions.

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
dandominguez04: Check empty list while is not searching. Do you think is confusing?

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dandominguez04: @dmsnell thanks for the feedback. It is clear that there are some bugs regarding tag names and how we compare them and search them. All these queries were passed directly (copied) from the original activity to keep the same behavior while refactoring. Do you think we should make all these changes in this PR or make separate issues to introduce the changes?

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
dmsnell: > Do you think is confusing?

is it not confusing to you? `false` does not convey to me the part about `while is not searching`
I know it's common to use these flags, but I find _boolean blindness_ to be a common confusing style.
it's not like `checkEmptyList` itself is that clear, but is there a way we could remove the anxiety
introduced by an unexplained boolean?

dandominguez04: Yes, I think we can create a specific method for each behavior and let the caller decide which method to call. Allow me to introduce the change.

dmsnell: it looks like we're not even "checking" anything here, more like updating a label.

doodling…

```java
// as separate methods
public void labelNoTags() {
    setEmptyListImage(…);
    setEmptyListMessage(…);
}

public void labelNoTagSearchResults() {
    if (DisplayUtils…) { … } else { … }

    setEmptyListMessage(…);
}

// as a named value with inherent meaning

public enum TagListContext {
    AllTags,
    SearchResults
}

public void updateEmptyListLabel(TagListContext context) {
    switch (context) {
        case AllTags:
            setEmptyListImage(…);
            setEmptyListMessage(…);
            break;

        case SearchResults:
             …
    }
}
```

of course if we use the `enum` we wouldn't have to switch (no pun intended) from an `if` to a `switch`. we could leave everything else the same. thoughts?

```java
checkEmptyList(AllTags); // vs. checkEmptyList(false);
checkEmptyList(SearchResults); // vs. checkEmptyList(); or checkEmptyList(true);
```

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dmsnell: > All these queries were passed directly (copied) from the original activity to keep the same behavior while refactoring

Ugh. Bummer - sorry you have to deal with this!

> Do you think we should make all these changes in this PR or make separate issues to introduce the changes?

Like I said, this app has a lot of issues with tags. A separate PR is fine. Would it be much work to propose that change _before_ this refactor? It doesn't matter too much, but I prefer seeing that we fix something before moving it around to another place, leaving its trail of buggy code in the project history.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dandominguez04: @dmsnell do you have some examples with I can work and test. I think I understand the problem but I would like some samples you have had problems with before so I can add tests to be sure we are covering those cases.

I think something like `Tag2` and `tag2` works now because the `like` query in Simperium is case insensitive. However, I think the cases you mentioned related to Unicode normal forms might not pass.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dandominguez04: > Ugh. Bummer - sorry you have to deal with this! 

No problem 😅️. Let's fix them all!

> It doesn't matter too much, but I prefer seeing that we fix something before moving it around to another place, leaving its trail of buggy code in the project history.

We could do it before on a separate PR, although I am inclining on doing the changes in this PR since we have the code encapsulated in the SimperiumTagsRepository and the changes might not be that big (few lines of code). 

Let me see if the solution is simple enough to fix it in the repository.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dmsnell: We'll see if I can do this without GitHub mangling the text. (Edit: nope)

 - `Bu\u{0308}cher` and `B\u{00fc}cher` "Bücher" are the same tag
 - `q\u{0307}\u{0323}` and `q\u{0323}\u{0307}` "q̣̇" are the same tag

If these two cases are handled properly (even if it works with the query itself) then it's probably all fine. It _can_ be difficult to get the right test inputs since text entry systems do like to perform their own normalization (that is, even if copying and pasting, sometimes when you paste two different things they turn into the same thing, and it's hard to tell since they are both visually identical).

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
dandominguez04: I implemented them as separate methods. Fixed in e15ef97222b8dfc62bfe4f03b2dcf4b615ab63ec.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/TagsViewModelTest.kt
dandominguez04: > Why did you choose to setup the tagItems every time a test runs?

There is not a particular reason. I fixed it in ed892e53be8b775c162bc617388d563460b1cc03.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
dandominguez04: The test `clickDeleteTagOfTagWithNotesShouldDeleteTagDirectly` hits this branch.

ParaskP7: 👋 @dandominguez04 !

This is interesting, it's been some time now and I don't remember exactly, but can you please confirm that by running `Run test with Coverage`? I am pretty sure that when I did that I was seeing this branch, either the `if` or the `else` not being hit. 🤔

dandominguez04: You are correct. I was using the wrong tag in the test. Fixed in 4246fd70f037c3011491820632227f84f4dd806b.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
dandominguez04: > Wouldn't the flow need to run on another context as well?

Yes! They run inside a bucket, so it is safe to run on the `IO` dispatcher. Fixed in 7f19bb404e880a70dff2e6eb636efdbd88b120e6.

dandominguez04: > Consider adding tests for this collect { ... } logic.

The test added uses a `MutableSharedFlow` as a return value of `tagsChanged` to be able to manipulate the flow whenever we want to trigger the `collect` method. When we emit values to the flow, this triggers the collect method, which calls `allTags()`. Fixed in b05c4fac4952bf65ef17839ab177df2ddc0c316a.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
dandominguez04: > Consider adding tests for this function.

Use a similar test to the test for `collect { ... }` logic with the difference that when we call `pause`, any changes to the tags is not listened, therefore the `UiState` is not updated. Fixed in 79c13d4b073b2f1b200967ea0ee9e67b183ffe63.

dandominguez04: > Do you think this comment is necessary?

Since `callbackFlow` is a new library and it is not explicit and clear that `cancel()` eventually calls to `awaitClose()`, I think the comment is useful for any reader which has not used `callbackFlow` before.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
ParaskP7: Thanks for the fix! 🙏

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
ParaskP7: 👋 @dandominguez04 !

> Yes! They run inside a bucket, so it is safe to run on the IO dispatcher. Fixed in 7f19bb4.

What about using the `callbackFlow { ... }.flowOn(...)` instead?

> Fixed in b05c4fa.

Thanks for the fix! 🙏

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
ParaskP7: 👋 @dandominguez04 !

> Fixed in 79c13d4.

Thanks for the fix! 🙏 

> Since callbackFlow is a new library and it is not explicit and clear that cancel() eventually calls to awaitClose(), I think the comment is useful for any reader which has not used callbackFlow before.

This makes sense, let's keep it then! 👍 Maybe, one last suggestion then would be to move this comment(s) to a Javadoc on top of the `pause()` function itself. But, it is totally up to you at this point.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/TagsViewModel.kt
dandominguez04: Yes, I agree `flowOn(...)` is the right method. Fixed in 41920bd4502cab737ff7381edbe5982b5bee1fc3.

ParaskP7: 👋  @dandominguez04 !

> Yes, I agree flowOn(...) is the right method. Fixed in 41920bd.

Great, thanks for updating that!

ParaskP7 (approved):
👋 @dandominguez04 !

I have looked at all your updates, resolved all my individual comments and it all LGTM now, thanks for working on this! 🌟 

PS: This PR is now approved from my side, but please do check with @dmsnell before progressing with a merge. 🙏

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dandominguez04: Hi @dmsnell ,

I have tried both cases:

> Bu\u{0308}cher and B\u{00fc}cher "Bücher" are the same tag
imagen

 &nbsp;
 &nbsp; 
  

> q\u{0307}\u{0323} and q\u{0323}\u{0307} "q̣̇" are the same tag
imagen

It seems that both cases worked. However,

>  It can be difficult to get the right test inputs since text entry systems do like to perform their own normalization (that is, even if copying and pasting, sometimes when you paste two different things they turn into the same thing

Yes, this made me hesitate a little bit about the results, but let me add an issue to add tests for the `SimperiumTagsRepository` and allow me to investigate more how to format the inputs so we can test this properly (see issue #1403).

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/TagsViewModelTest.kt
dandominguez04: As I mentioned above, this behavior should be tested at the repository level. At the ViewModel level, the repo is mocked. The solution for #1403 should have the tests for edge cases on tag names.

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.java
dandominguez04: This is a good idea. I have added just one method since both toasts have the same duration, but these are functions that can be added as utility functions for all the project. Fixed in fc1bf536bd26c1eee274a2c11c2290bfda37568a.

```kotlin
 private void toast(@StringRes int resId) {
        Toast.makeText(TagsActivity.this, getString(resId), Toast.LENGTH_SHORT).show();
 }
```

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dandominguez04: Fixed in 0429e79402759be2ba4fd93d060e03173f8126bf. I noticed that when the tag is added, it adds the canonical form. For example, if `tag3` exists and I type `TAG3`, the app changes to `tag3`.

Simplenote/src/main/java/com/automattic/simplenote/models/Note.java
```diff
@@ -449,6 +449,14 @@ public void setPublished(boolean isPublished) {
         }
     }
 
+    public void removeTag(String tag) {
+        List<String> tags = getTags();
+        tags.remove(tag);
```
dmsnell: I'm sure you were planning on updating this as well, but we want to compare by tag hash here and not by tag name, or better, use some kind of `Tag.equals` call (which doesn't exist so probably best not to add in this PR)

```java
List<String> tags = getTags();
String tagHash = TagUtils.hashTag(tag);

for (String tagName : tags) {
    if (TagUtils.hashTag(tag) == tagHash) {
       tags.remove(tag);
    }
}

setTags(tags);
save();
```

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dmsnell: Good checking! If it's too hard _and_ especially if this is already broken, let's not worry too much about it. Otherwise, in a test (or for a quick check by modifying the code) we should be able to manually add a new tag with a hard-coded name, as we can do that in Java strings as well.

```java
TagUtils.createTagIfMissing(tagBucket, "q\u0307\u0323");
TagUtils.createTagIfMissing(tagBucket, "q\u0323\u0307");
```

and then we can perform a tag search and inspect the tag bucket/database to see if it created two entities, or maybe not 🤔 it's so confusing sometimes to intentionally inject bad data.

Simplenote/src/main/java/com/automattic/simplenote/models/Note.java
dandominguez04: There are actually some methods in the `Note` model we can use to check tags:

```java
   public boolean hasTag(String tag) {
        List<String> tags = getTags();
        String tagLower = tag.toLowerCase();
        for (String tagName : tags) {
            if (tagLower.equals(tagName.toLowerCase())) return true;
        }
        return false;
    }

    public boolean hasTag(Tag tag) {
        return hasTag(tag.getSimperiumKey());
    }
```

Let me check how I can use them to check the tags.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dmsnell: > it adds the canonical form. For example, if tag3 exists and I type TAG3, the app changes to tag3.

we've still been settling these decisions. the goal is to be able to prompt someone that what they are adding as a new tag already exists, and then to give them a choice to rename the existing tag or adopt the existing name.

we did decide that in the individual note view, when listing tags, that those should eventually all shows the canonical name as provided in the `Tag` entity itself, regardless of how it's written in the `Note`'s `tags` property.

Simplenote/src/main/java/com/automattic/simplenote/models/Note.java
dmsnell: hm. I would recommend against using that method due to it being inaccurate - in separate work one of us should update that. also, it seems silly to iterate over all the tags inside a loop where we're already iterating over the tags.

I only meant to suggest that we rely on the `tagHash` here instead of the name. the whole `Tag.equals()` idea is just that basically. it wouldn't be more robust than…

```
// TagUtils.java or Tag.java
public static boolean areEqual(String left, String right) {
  return hashTag(left) == hashTag(right);
}
```

keeping it local and _not_ abstracting is probably most helpful at this point since `Tag`, `TagUtil`, and `Bucket` are all so jumbled and coupled together. in an ideal world I think that the `Note` object would have something like a `TagSet` property which itself held `Tag` objects and this would all be properly abstracted away from the data. for now it's just a jumble that would be hard to untangle 🤷‍♂️

dandominguez04: Yeah, I misread your first comment sorry (which had the solution). Fixed in a07747c09d8a98574723b9022e697c1f5a81f9e6.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumTagsRepository.kt
dandominguez04: There is also the case of two different users adding lexical variations of the same tag before synchronizing. For example, user 1 adds `TAG3` and user 2 adds `tag3`. The note will keep the two tags I think.

Simplenote/src/main/java/com/automattic/simplenote/models/Note.java
dandominguez04: > in an ideal world I think that the Note object would have something like a TagSet property which itself held Tag objects and this would all be properly abstracted away from the data. for now it's just a jumble that would be hard to untangle

That would be interesting. It would be much simpler.

danilo04:
Hi @dmsnell 👋 . After talking with @ParaskP7, we decided to merge the PR to the feature branch and create separate issues for improvements on the tag management you mentioned.

PR #1412, opened 2021-07-12
Adding some UI tests for login & sync use cases

qcorbion:

### DIFF
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
@@ -1,0 +1,496 @@
+package com.automattic.simplenote.login;
+
+import static androidx.test.espresso.Espresso.onView;
+import static androidx.test.espresso.action.ViewActions.click;
+import static androidx.test.espresso.action.ViewActions.pressImeActionButton;
+import static androidx.test.espresso.action.ViewActions.swipeDown;
+import static androidx.test.espresso.action.ViewActions.swipeUp;
+import static androidx.test.espresso.action.ViewActions.typeText;
+import static androidx.test.espresso.matcher.ViewMatchers.hasDescendant;
+import static androidx.test.espresso.matcher.ViewMatchers.isAssignableFrom;
+import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static androidx.test.espresso.matcher.ViewMatchers.isRoot;
+import static androidx.test.espresso.matcher.ViewMatchers.withContentDescription;
+import static androidx.test.espresso.matcher.ViewMatchers.withId;
+import static androidx.test.espresso.matcher.ViewMatchers.withText;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.core.IsNull.notNullValue;
+import static org.junit.Assert.assertThat;
+import static tools.fastlane.screengrab.cleanstatusbar.BarsMode.TRANSPARENT;
+import static tools.fastlane.screengrab.cleanstatusbar.IconVisibility.SHOW;
+
+import android.content.Context;
+import android.content.Intent;
+import android.util.DisplayMetrics;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.ViewParent;
+
+import androidx.appcompat.widget.SearchView;
+import androidx.test.core.app.ApplicationProvider;
+import androidx.test.espresso.PerformException;
+import androidx.test.espresso.UiController;
+import androidx.test.espresso.ViewAction;
+import androidx.test.espresso.ViewInteraction;
+import androidx.test.espresso.action.ViewActions;
+import androidx.test.espresso.assertion.ViewAssertions;
+import androidx.test.espresso.contrib.RecyclerViewActions;
+import androidx.test.espresso.matcher.ViewMatchers;
+import androidx.test.espresso.util.HumanReadables;
+import androidx.test.espresso.util.TreeIterables;
+import androidx.test.filters.LargeTest;
+import androidx.test.platform.app.InstrumentationRegistry;
+import androidx.test.rule.ActivityTestRule;
+import androidx.test.runner.AndroidJUnit4;
+import androidx.test.uiautomator.By;
+import androidx.test.uiautomator.UiDevice;
+import androidx.test.uiautomator.Until;
+
+import com.automattic.simplenote.BuildConfig;
+import com.automattic.simplenote.NotesActivity;
+import com.automattic.simplenote.R;
+import com.google.android.material.textfield.TextInputLayout;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeoutException;
+
+import tools.fastlane.screengrab.Screengrab;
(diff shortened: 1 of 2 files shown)


pachlava:
Thank you for creating the PR @qcorbion! 👍  I'll be adding some review notes today, maybe not in one sitting 🙂

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
```diff
@@ -0,0 +1,496 @@
...
+                .setBarsMode(TRANSPARENT)
+                .enable();
+    }
+
+    @Test
+    public void loginTest() throws InterruptedException {
+        // Pre-checks if the state is dirty
+        enterThenDisablePasscodeIfNeeded();
+        dismissVerifyEmailScreenIfNeeded();
+        logoutIfNeeded();
```
pachlava, line 105: Given this sequence is shared in all four tests, what do you think of moving it into a method annotated with `@Before`, which serves the same purpose as `setUpWithError()` in Swift?

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
```diff
@@ -0,0 +1,496 @@
...
+import tools.fastlane.screengrab.locale.LocaleTestRule;
+
+@LargeTest
+@RunWith(AndroidJUnit4.class)
+public class LoginTests {
+    private static final int LAUNCH_TIMEOUT = 5000;
+
+    private static final String NOTE_FOR_EDITOR_SHOT_TITLE = "Lemon Cake & Blueberry";
+    private static final String NOTE_FOR_INTERLINKING_SHOT_TITLE = "# Colors";
+    private static final String NOTE_FOR_INTERLINKED_NOTE_SHOT_TITLE = "Blueberry Recipes";
```
pachlava, line 80: Since this file is based on `SceenshotTest`, it is natural to inherit the code, of course. But I would see value in keeping only the code that is relevant for the tests - this would improve the readability.

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
```diff
@@ -0,0 +1,496 @@
...
+
+    @Test
+    public void loginWithWrongPasswordTest() throws InterruptedException {
+        // Pre-checks if the state is dirty
+        enterThenDisablePasscodeIfNeeded();
+        dismissVerifyEmailScreenIfNeeded();
+        logoutIfNeeded();
+
+        // Login with incorrect password
+        loginWithSpecificCredentials(BuildConfig.SCREENSHOT_EMAIL, "wrongPassword");
```
pachlava, line 139: One of the key attributes of the test is expecting for something to happen. Without this, the test becomes a sequence of actions with no defined outcome. As it is, the test will pass if the sequence is completed, and will fail in the opposite case.

In this test, we log in with a wrong password, but do not really check what happens next? Do we see an error, or do we get logged in 🤔 

I would suggest making use of [Assert](https://junit.org/junit4/javadoc/latest/org/junit/Assert.html), allowing tests to pass or fail based on the expected outcome.

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
```diff
@@ -0,0 +1,496 @@
...
+    }
+
+    @Test
+    public void logoutTest() throws InterruptedException {
+        // Pre-checks if the state is dirty
+        enterThenDisablePasscodeIfNeeded();
+        dismissVerifyEmailScreenIfNeeded();
+        logoutIfNeeded();
+
+        // Login
```
pachlava, line 124: I would say, if the methods naming is descriptive and clean (and this is something we should aim for), the comments like this become a bit redundant 🤔

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
```diff
@@ -0,0 +1,496 @@
...
+        login();
+
+        // Waiting for the notes (to make sure the login was successful)
+        Thread.sleep(2000);
```
pachlava, line 111: While I'm not free of this sin myself 😅 , it's preferable to avoid hard-coded sleeps and wait (with a timeout) for something to appear on the screen instead, like `All Notes` navigation bar of the FAB button.

One of the issues with hardcoded sleeps, which I stumbled upon recently, is the variability of the required time for the sleep: the test can run perfectly on a high-end laptop, but when executed as a part of CI cycle, where we usually have VMs, the used amount of 😴 might not be enough, because the VM performance might be lower.

Simplenote/src/androidTest/java/com/automattic/simplenote/sync/SyncTests.java
```diff
@@ -0,0 +1,696 @@
...
+        onView(withContentDescription("More Options")).perform(click());
+
+        onView(withText("Publish")).perform(click());
+
+        onView(withContentDescription("Navigate up")).perform(click());
+
+        waitForViewMatching(withText(R.string.all_notes), 500);
+
+        waitForViewMatching(allOf(withId(R.id.note_title), withText(title)), 5000);
+    }
```
pachlava, line 149: It looks like this method does more than its name implies. For such a scenario, I would have the code broken up into three methods instead (naming is just an example):

- `createNoteFromNoteList(...)`
- `publishCurrentNote()`
- `leaveNoteEditor()`

This would allow having more granular tests and steps, which you can still freely combine into some mega-methods.

One of the things that happens once the note is published is generation of a URL, which can be used to open the published note in the browser. So, automating this test would definitely require checking that the URL is generated and can be opened in the browser. Unfortunately, note publishing does not work in the dev-DB, and I forgot to let you know about this earlier 🤦. Sorry for this confusion.

pachlava (commented):
I left some review notes, and I think at this point it makes sense to work on them and the infrastructure, rather than automating more tests.

One of the points I'd like to add here, is that since `SyncTests` and `LoginTests` share a lot of code, it would make sense to move out the shared parts into a separate helper file, which then will be utilized by both. Of course, ideally, the same can be done for the original `ScreenshotTest`, but since it's out of our scope, let's not touch it.

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
qcorbion: It's a good idea actually

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
qcorbion: Agreed, seems that I completely forgot to check that the login failed

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
qcorbion: This code is a copy/paste/hack of ScreenshotTest. I'm going to clean it better. Thanks.

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
qcorbion: Ahah you're right 🤣

Simplenote/src/androidTest/java/com/automattic/simplenote/login/LoginTests.java
qcorbion: Fully agreed, I'm going to change the logic here

Simplenote/src/androidTest/java/com/automattic/simplenote/sync/SyncTests.java
qcorbion: I'll split it up. Thanks for the URL generation, I didn't know that.

PR #1414, opened 2021-07-12
Fix list of searched tags after leaving and entering the app

### LINKED ISSUE #1413: List of search tags is cleared after leaving and entering the app
Parent: #1361

When the list of tags is filtered when a user searches with a term, the filtered list is cleared (show the complete list of tags) when the user enters the app again after it locks the screen. 

**Note:** this bug is present just in branch `feature/modernized_tags`.

### Expected
The filtered list of tags is kept after entering the app again:

### Observed
The filtered list of tags is cleared even though the search term is still in the search bar:

### Reproduced

1. Open the menu
2. Tap on Edit (Tags)
3. Tap on the Search icon (lens icon)
4. Introduce a search term that filters the list of tags
5. Tap on the home button
6. Enter Simplenote again

Device Make: Samsung
Device Model: Galaxy A72
Android Version: 11
Simplenote Version: 2.18

danilo04:
Fixes #1413 

### Fix

The previous implementation of the method `start()` setup the initial `UiState` with all tags and empty search terms. It also starts listening to tag changes from the tag's bucket. The `start()` method was called in `onResume`, thus if the user leaves the app and comes back (the activity is restarted), the `start()` method was called every time, cleaning the `UiState` to its initial state. This created a problem when the user searched tags and left the app and come back because the search term stayed in the search bar but showed the complete list of tags (the expected behavior is to show the filtered list of tags).

This PR separates the setup of the initial `UiState` (method `start()`) and start listening to tag changes (method `startListeningTagChanges()`). The method `start()` is now called in the `onCreate()` callback to avoid cleaning the `UiState `when the user returns an activity (activity is restarted).

### Test

1. Open the menu
2. Tap on Edit (Tags)
3. Tap on the Search icon (lens icon)
4. Introduce a search term that filters the list of tags
5. Tap on the home button
6. Enter Simplenote again
7. The list of search tags should be filtered

### Review

This PR will be reviewed by @ParaskP7

### Release

### DIFF
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -121,6 +121,9 @@ dependencies {
     // Support for ViewModels and LiveData
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.3.1'
     implementation 'androidx.lifecycle:lifecycle-livedata-ktx:2.3.1'
+    // Support for androidx Fragments
+    implementation "androidx.lifecycle:lifecycle-viewmodel-savedstate:2.3.1"
+
     // Google Play Services
     implementation 'com.google.android.gms:play-services-wearable:17.0.0'
     // Third Party
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/integration/TagsActivityTest.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/integration/TagsActivityTest.kt
@@ -1,6 +1,7 @@
 package com.automattic.simplenote.integration
 
 import android.widget.EditText
+import androidx.lifecycle.Lifecycle
 import androidx.recyclerview.widget.RecyclerView
 import androidx.test.core.app.ActivityScenario
 import androidx.test.espresso.Espresso.onView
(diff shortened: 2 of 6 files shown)


ParaskP7 (approved):
👋 @danilo04 !

I have reviewed and tested this PR as per the instructions, everything works as expected, great find and fix! 🌟 🌟 🌟 

Praise (❤️): I love the fact that you used a TDD approach to first create a UI test and verify that the bug exists before jumping into fixing it.
Suggestion (💡): I have noticed the structure of your commit messages and I wonder if you would like a couple of tips related to that.
- In your IDE settings, go to `Version Control` -> `Commit`.
- Find the `Commit message inspections` section.
- Enable the `Blank line between subject and body` warning.
- Enable the `Limit body line` warning and set the right margin to `72`. Also, I suggest enabling the `Show right margin` and `Wrap when typing reaches right margin`. This will really help you switch lines automagically when typing a commit message.
- Enable the `Limit subject line` warning and set the right margin to `72` as well.
- Enable the `Spelling` to a `typo` severity.

PS: The `72` character is the default suggested line length of a commit message. Consider this is as an industry standard.
PSS: Also, consider not using code related words in your commit title (ie. `RecyclerViewItemCountAssertion`). A title should be kept simple, understandable and readable by all. As far as using code related words in your commit description, consider enclosing them with a quote, like `'RecyclerViewItemCountAssertion'` so that the eye can immediately separate the code related words from the normal ones.

Feel free to skip my suggestions if all the above sound too much, I just thought to share what I am doing as well when I am developing, that is, in relation to commit messages. 🙏

danilo04:
Hi 👋  @ParaskP7. 

Thanks for the review and your comments. I have activated the 72 line restrictions. 

>  Also, consider not using code related words in your commit title (ie. RecyclerViewItemCountAssertion)

Got it.

PR #1416, opened 2021-07-14
Implements In App Account Deletion

### LINKED ISSUE #1415: In App Account Deletion
Adds the ability to delete your own account from the Simplenote Android app. 

### What
Users can delete their accounts within the Simplenote Android app. Details of this project can be found at: p2XJRt-3bK-p2. The last iteration of designs can be found at: p2XJRt-3cd-p2.

### Why
Feature parity with iOS app which is required by Apple Store's new policies that require apps on the app store that have accounts to allow them to be deleted from within the app.

### How
image

danilo04:
Fixes #1415 

### Fix
Implement a feature that let users delete their account from within the app. This feature sends a request to the Simplenote server which sends an email to the user. The user can then delete the account using the email with the required information. If the request fails because of Internet connection or server issues, then we show an error message to the user, otherwise, we show a received request message. The videos below show both flows:

**Light Mode**
| Successful Flow      | Error Flow |
| ----------- | ----------- |
|  |       |

**Dark Mode**
| Successful Flow      | Error Flow |
| ----------- | ----------- |
|   |      |

More details about the project: p2XJRt-3bK-p2
Last Iteration of designs: p2XJRt-3cd-p2

### Test

**Note: use Simperium key for production and create accounts for tests. You can create accounts with the following format: youremail+n@automattic.com where n is an integer number (e.g. danilo.dominguez+1@automattic.com). Follow the instructions in the email to setup password.**

1. Open the left menu by tapping on the top-left corner icon
2. Tap on Settings
3. Scroll down and Tap on the preference `Delete Account`
4. That will open a dialog. Tap on `DELETE ACCOUNT` button
5. Open your email and proceed to delete account
6. Open Simplenote and it should logout automatically

### Review

This PR can be reviewed by @ParaskP7  or @hichamboushaba.

### Release
`RELEASE-NOTES.txt` was updated in fd9ac10108a556a42ce1b2dbe33c3babb63d90d5 with:
> Added In App Account Deletion


SylvesterWilmott:
Thanks @danilo04 

This looks great and works as expected. Design-wise theres just a few things:

- In dark mode, the red text in the buttons in the settings and dialog should be red20 to provide adequate contrast. This is the general rule we use in dark mode ie. all text should use the `20` variant of the light mode's counterpart.
- Can we update the copy of the error copy to - `An error occurred. Please, try again. If the problem continues, contact us at support@simplenote.com for help.`
- I noticed that in the "Request Received" dialog, my email address was broken (ie. it flowed onto two lines), is this a consequence of it being too long to fit on one line or do we need to make sure that the dialog doesn't break this particular word?

danilo04:
Hi @SylvesterWilmott, thank you for the review. 

> In dark mode, the red text in the buttons in the settings and dialog should be red20 to provide adequate contrast. This is the general rule we use in dark mode ie. all text should use the 20 variant of the light mode's counterpart.

Fixed in 572d4dd46cc26b211b7fabee210c777d67b82de8 and 2bd355e31edae1ac5f486a1f38b4c673fbc5f6de. 

> Can we update the copy of the error copy to - An error occurred. Please, try again. If the problem continues, contact us at support@simplenote.com for help.

Fixed in c2aa7aa5345a218fa7bf33727771a1cdb07772de.

> I noticed that in the "Request Received" dialog, my email address was broken, is this a consequence of it being too long to fit on one line or do we need to make sure that the dialog doesn't break this particular word?

Let me investigate this case because it is breaking just the email in any case. If you rotate your phone, the email can be put in the next line, but it is still breaking it.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -116,6 +130,18 @@ public boolean onPreferenceClick(Preference preference) {
             }
         });
 
+        Preference deleteAppPreference = findPreference("pref_key_delete_account");
+        int preferenceLayout = ThemeUtils.isLightTheme(requireActivity()) ?
+                R.layout.preference_red_light : R.layout.preference_red_dark;
```
hichamboushaba: Just a suggestion, I didn't test it to see if it works, why not specify the layout directly in `preferences.xml` using the `android:layout` option, to have a separate layout for dark them maybe you can define it in the `layout-night` folder, or define the colors in the `values-night` folder, and use the same layout.
WDYT?

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -304,6 +330,185 @@ public boolean onPreferenceClick(Preference preference) {
...
+                        if (response.isSuccessful()) {
+                            AppLog.add(Type.ACCOUNT, "Request to delete account was successful");
+
+                            showDeleteAccountConfirmationDialog();
+                        } else {
+                            AppLog.add(Type.ACCOUNT, "Request to delete account had an " +
+                                    "error. Error code: " + response.code());
+
+                            showDialogDeleteAccountError();
+                        }
```
hichamboushaba: just a thought, I don't think it's a good idea to expose OkHttp's CallBack here, it's making the fragment responsible for network logic (response.isSuccessFull).
I would prefer to have some kind of interface as listener for the account deletion, with two simple methods: `onSuccess` and `onFailure`, which we can then pass to `makeDeleteAccountRequest`, and in `AccountNetworkUtils` we would use the `Callback` to handle the logic of success and failure and forward it to our listener.

hichamboushaba (approved):
Good job @danilo04, the code looks code, and the change works as expected.

I left some comments for you to consider, but I'm pre-approving as they are not blockers.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
danilo04: Yes, I think that makes the code much cleaner and removes code from the fragment that should be in the layout. It also makes it simpler and flexible to add more themes to the app. Fixed in c177fab7e3144f5469c27ed2a048105783961dcc.

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
danilo04: HI @hichamboushaba 👋 . Yes, I agree, it is not the responsibility of the fragment. I have added a handler in 28c1446a4d80a59936bc5de194bd53560cc0a070 and move the fragment to a weak reference in the handler (dd8199e4f88400f2d43936ddc55bfbeb5f1d887c) to avoid memory leaks.

danilo04:
Hi @SylvesterWilmott 👋 . I have added track events for delete account in cc253a776c4772647e72f7afdcabd389d6ad9a17.

SylvesterWilmott (approved):
Approving the design on this since I believe the email breaking will be tackled separately. Thanks @danilo04

danilo04:
I am going to merge the PR since it has been approved.
@SylvesterWilmott, I have added the issue regarding how the email is shown in two lines: #1427.

PR #1418, opened 2021-07-16
Add final details to Import Data Feature

### LINKED ISSUE #1417: Final details for Import Data Feature
This issue is to keep track of the final details for merging `Import Data` feature (PR #1333) to `develop`. After the merge, we need to add and revise few details to make the feature ready to release. 

- [x] Add `AppLog` to the import process
- [x] Add Feature to `RELEASE-NOTES.txt`
- [x] More tests
- [x] Review UI messages and design

danilo04:
Fixes #1417

### Fix
This PR is a follow-up on #1333 to add final details to the feature of importing notes in simplenote-android. 

| Importing one note (Marknown)   | Importing Several Notes (JSON) |
| ----------- | ----------- |
|  |   |

### Test

1. Go to the top-left corner button and tab to open the menu
2. Tap on `Settings`
3. Scroll down and tab on `Import data`
4. Select a file with extensions `.txt`, `.md` or `.json`. The `.json` should be exported from a Simplenote app.
5. Go to All Notes and verify that the note or notes were imported

### Review

### Release

`RELEASE-NOTES.txt` was updated in ba7a22929cfc9c857d06427bb6a97b8a3feef2d3 with:
> Added support to import notes


Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -475,17 +476,36 @@ public int compare(String text1, String text2) {
 
     private void importData(Uri uri) {
         try {
-            Importer.fromUri(this, uri);
+            AppLog.add(Type.IMPORT, "Importing notes from " + uri + ".");
+
+            FragmentActivity activity = getActivity();
+            if (activity == null) {
```
ParaskP7, line 488: Question (❓): Although this case will not happen that ofter, when or if it does, shouldn't we show a `toast` of some kind, just it happens when an exception is thrown? This way the user will know what happen. Otherwise, this code branch will silently complete without any message, leaving the user confused. Wdyt?

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
```diff
@@ -475,17 +476,36 @@ public int compare(String text1, String text2) {
...
+            }
+
+            Importer.fromUri(activity, uri);
             toast(R.string.import_message_success);
+
+            AppLog.add(Type.IMPORT, "Notes imported correctly!");
         } catch (Importer.ImportException e) {
             switch (e.getReason()) {
                 case FileError:
+                    AppLog.add(Type.IMPORT, "File error while importing note. Exception: "
```
ParaskP7: Minor (🔍): Consider making the `AppLog` a one-liner since it actually fits.

PS: The same applies for those other two `AppLog` lines further below.

ParaskP7 (commented):
👋 @danilo04 !

I have reviewed and tested this PR as per the instructions, everything works as expected (ish), good job! 🌟 

The one thing I couldn't make working is importing from `.txt` or `.md` files. Is that working in general? If so, how can I import from a `.txt` or `.md` file? I managed to export and import from the generated `.json` file, but when I tried to manually create a `.txt` or `.md` note files on my PC, I then couldn't import them. Those file types weren't actually available for selection, they were shown but disabled within the system's file chooser. Can you share you thoughts on that and help me understand what/if I am doing something wrong? Also, to my understanding you can't export in `.txt` or `.md` format, right?

Also, I have left one question (❓) and a minor (🔍) comment for you to consider. I am **NOT** going to approve this PR for now, not until we discuss this `.txt` and `.md` file import issue I am having, but I have the feeling that none is blocking. I am just trying to be extra careful.

SylvesterWilmott:
Design-wise this looks good, thanks @danilo04 

One request, can we add a tracks event to the import button in the settings?

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
danilo04: 👋  @ParaskP7. Yes, this is an edge case that will rarely happen. When it happens, is because some weird callback interleaving created by either device rotation or closing the app. The problem of showing a `Toast` here is that if the activity is `null` for the fragment, how we obtain the application context to show the `Toast`. Do you know if `requireContext()` does not throw an exception if activity is `null`?

ParaskP7: 👋 @danilo04 !

Actually, I was overthinking it and you're right, when and if that happens, there are going to be bigger problems than just showing a toast, at which case I am now thinking that logging would be enough so that we at least know that this happened. Please ignore my comment, I think we are good if we leave that as is for now.

🙇

Simplenote/src/main/java/com/automattic/simplenote/PreferencesFragment.java
danilo04: @ParaskP7 I am using 100 characters line length (default) and if I put the message, it overflows the 100 characters. Do you have another value (maybe 120)?

ParaskP7: Oh yea, `120` is the default in general and we are using that on `WPAndroid` and `WCAndroid` as far as I am aware. I would recommend maybe having `120` on Simplenote for consistency purposes as well. 😊

danilo04: Sounds good, thanks! Fixed in de71d5b472ca3060559429ddeb42176241e00101.

danilo04:
Hi 👋  @SylvesterWilmott. I have added tracks in 75baffef90f25e48b53ca8e166c473cab5623bd2. I track two events, when the user clicks the button and when the file is loaded (since the user can fall back and don't import anything). Please let me know if those events are OK.

ParaskP7 (approved):
👋 @danilo04 !

Your updates, including `tracks`, LGTM, good job! 👍

pachlava:
@danilo04 Just leaving a note here that I've tested the import as a part of 2.19-rc-1 beta testing, and so far found no real issues 🎉 . What I've tried:

✅  Importing a valid Simplenote-generated .json is handled correctly (tags, notes and trash are filled properly)
✅  Import of usual .txt and .md works
✅  Import of corrupted .json (a proper json with messed structure, or an image file renamed to .json) is rejected gracefully
✅  Import of image files renamed to .txt or .md results in notes having raw content of images, which is fine, I guess.

I, however, have one question about this scenario:

❓Import of extra large .txt works per se. I've tried [1.14 Mb Moby Dick](https://github.com/ErikSchierboom/sentencegenerator/blob/master/samples/moby-dick.txt), it was imported and could be opened. It took about 3 minutes to open on a mid/low-range Android device. The issue is that this note was never synced, due to 413 error (from what I see in Electron). This opens a larger topic [to discuss](p1628679573003000-slack-C036Y8QL4): should all Simplenote clients have a limitation on a note size.

danilo04:
Thanks for the tests @pachlava. I will join the discussion there.

Issue #1353, opened 2021-03-31
Email verification: Address showing up as null, account may not exist?


codebykat:
A user reported (3868456-zen) that they are prompted to verify their email when opening the app, but their email is shown as the word "Null".

They also say they don't recall signing up for the app / setting a password, and I don't see an account for the email address they contacted us from. They say they've been using the app for many years and have never had to sign in.

I've asked them for more information in the ticket but I wonder if it's possible we still have some folks using the account without having created an account - I found https://github.com/Automattic/simplenote-android/pull/432 in the history but can't tell if we still allow account-less usage for those very old accounts.

pachlava:
@codebykat please also see #1353 - `null` email address is mentioned there as well

codebykat:
@pachlava This is 1353 - what did you intend to link? I searched other issues for "null" and I'm not seeing anything.

pachlava:
@codebykat whoops 🤦 , I meant #1343, sorry

codebykat:
The user got back to us to confirm that they do not have an account at all. 😱

rossanafmenezes:
Got another case here 3938469-zen just FYI

User mentioned it's an old account that did not require login. So probably from back when you could use SN on Android without an account.

PR #1423, opened 2021-07-23
Fix for email verification screen appears on accounts verified

### LINKED ISSUE #1343: Email verification screen appears for some verified accounts
While working on the new signup flow, we noticed that the email verification screen appears in some cases where the account should be already verified:
1. Accounts using the new signup flow.
2. WordPress.com login.
3. Login after install.

Steps to reproduce:
1. Click on signup.
2. Enter an email.
3. Check the email inbox to continue signup.
4. Complete the account creation.
5. Notice the email verification screen.
The issue occurs randomly, so you may need to repeat the above steps multiple times.

danilo04:
Fixes #1343, #1353

### Fix
Fixed the bug in which the email verification screen appears even though the account is verified. To verify whether the account was verified or not, we need the `Account` bucket. This bug was related to a race condition in which the application tried to check if the account wasn't verified before the `Account` bucket was synchronized.

To fix the issue, we wait until the account bucket is synchronized using the `OnNetworkChangeListener` listener with `ChangeType = INDEX`. By using this listener, we make sure that we are checking whether the account is verified when the
`Account` bucket is synchronized. 

### Test

**Note:** to facilitate triggering the bug, disable WiFi and allow just 2G connections on your phone. This can help to trigger the race condition.

Try testing with the previous version and this version:

1. Go to login
2. Tap on Login with Email and introduce your credentials
3. If the account is verified, the email verification screen should not be used. 

### Review

This PR would be preferably reviewed by @dmsnell since it requires knowledge of the `Simperium` listeners and logic.

### Release

`RELEASE-NOTES.txt` was updated in dce439e5bf1d95d18cf3566a1213d66aff1619f8 with:
> Fixed email verification screen appears with verified accounts

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -75,6 +75,7 @@ public class Simplenote extends Application implements HeartbeatListener {
     private static final long HEARTBEAT_TIMEOUT =  WebSocketManager.HEARTBEAT_INTERVAL * 2;
 
     private Activity mCurrentActivity;
+    private ReviewAccountVerifier mReviewAccountVerifier;
 
     private static Bucket<Account> mAccountBucket;
     private static Bucket<Preferences> mPreferencesBucket;
@@ -128,7 +129,8 @@ public class Simplenote extends Application implements HeartbeatListener {
             mTagsBucket = mSimperium.bucket(tagSchema);
             mPreferencesBucket = mSimperium.bucket(new Preferences.Schema());
             mAccountBucket = mSimperium.bucket(new Account.Schema());
-            mAccountBucket.addOnNetworkChangeListener(new ReviewAccountVerifier(this));
+            mReviewAccountVerifier = new ReviewAccountVerifier(this);
+            mAccountBucket.addOnNetworkChangeListener(mReviewAccountVerifier);
             // Every time a note changes or is deleted we need to reindex the tag counts
             mNotesBucket.addListener(new NoteTagger(mTagsBucket));
         } catch (BucketNameInvalid e) {
@@ -240,6 +242,10 @@ public class Simplenote extends Application implements HeartbeatListener {
     public String getUserEmail() {
         User user = mSimperium.getUser();
         return user != null ? user.getEmail() : null;
+    }
+
+    public ReviewAccountVerifier getReviewAccountVerifier() {
+        return mReviewAccountVerifier;
     }
 
     public void showReviewVerifyAccount(boolean hasSentEmail) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -36,7 +36,7 @@ import com.automattic.simplenote.utils.AppLog.Type;
 import com.automattic.simplenote.utils.CrashUtils;
 import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.PrefUtils;
-import com.automattic.simplenote.utils.ReviewAccountVerifier;
+import com.automattic.simplenote.utils.AccountVerificationWatcher;
 import com.automattic.simplenote.utils.SyncWorker;
 import com.simperium.Simperium;
 import com.simperium.android.AndroidClient;
@@ -75,7 +75,7 @@ public class Simplenote extends Application implements HeartbeatListener {
     private static final long HEARTBEAT_TIMEOUT =  WebSocketManager.HEARTBEAT_INTERVAL * 2;
 
     private Activity mCurrentActivity;
-    private ReviewAccountVerifier mReviewAccountVerifier;
+    private AccountVerificationWatcher mAccountVerificationWatcher;
 
     private static Bucket<Account> mAccountBucket;
     private static Bucket<Preferences> mPreferencesBucket;
@@ -129,8 +129,8 @@ public class Simplenote extends Application implements HeartbeatListener {
             mTagsBucket = mSimperium.bucket(tagSchema);
             mPreferencesBucket = mSimperium.bucket(new Preferences.Schema());
             mAccountBucket = mSimperium.bucket(new Account.Schema());
-            mReviewAccountVerifier = new ReviewAccountVerifier(this);
-            mAccountBucket.addOnNetworkChangeListener(mReviewAccountVerifier);
+            mAccountVerificationWatcher = new AccountVerificationWatcher(this);
+            mAccountBucket.addOnNetworkChangeListener(mAccountVerificationWatcher);
             // Every time a note changes or is deleted we need to reindex the tag counts
             mNotesBucket.addListener(new NoteTagger(mTagsBucket));
         } catch (BucketNameInvalid e) {
@@ -239,8 +239,8 @@ public class Simplenote extends Application implements HeartbeatListener {
         return user != null ? user.getEmail() : null;
     }
 
-    public ReviewAccountVerifier getReviewAccountVerifier() {
-        return mReviewAccountVerifier;
+    public AccountVerificationWatcher getReviewAccountVerifier() {
+        return mAccountVerificationWatcher;
     }
 
     public void showReviewVerifyAccount(boolean hasSentEmail) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -130,9 +130,8 @@ public class Simplenote extends Application implements HeartbeatListener {
             mPreferencesBucket = mSimperium.bucket(new Preferences.Schema());
             mAccountBucket = mSimperium.bucket(new Account.Schema());
             // Setup Account Verification Watcher to listen to network changes on the account bucket
-            mAccountVerificationWatcher = new AccountVerificationWatcher(this);
             VerificationListener verificationListener = new VerificationListener();
-            mAccountVerificationWatcher.addVerificationStateListener(verificationListener);
+            mAccountVerificationWatcher = new AccountVerificationWatcher(this, verificationListener);
             mAccountBucket.addOnNetworkChangeListener(mAccountVerificationWatcher);
             // Every time a note changes or is deleted we need to reindex the tag counts
             mNotesBucket.addListener(new NoteTagger(mTagsBucket));
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -128,7 +128,6 @@ public class Simplenote extends Application implements HeartbeatListener {
             mTagsBucket = mSimperium.bucket(tagSchema);
             mPreferencesBucket = mSimperium.bucket(new Preferences.Schema());
             mAccountBucket = mSimperium.bucket(new Account.Schema());
-            // Setup Account Verification Watcher to listen to network changes on the account bucket
             mAccountBucket.addOnNetworkChangeListener(
                     new AccountVerificationWatcher(this, new VerificationListener())
             );
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -1,5 +1,9 @@
 package com.automattic.simplenote;
 
+import static com.automattic.simplenote.models.Preferences.PREFERENCES_OBJECT_KEY;
+import static com.simperium.android.AsyncAuthClient.USER_ACCESS_TOKEN_PREFERENCE;
+import static com.simperium.android.AsyncAuthClient.USER_EMAIL_PREFERENCE;
+
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.Application;
@@ -32,12 +36,12 @@ import com.automattic.simplenote.models.NoteCountIndexer;
 import com.automattic.simplenote.models.NoteTagger;
 import com.automattic.simplenote.models.Preferences;
 import com.automattic.simplenote.models.Tag;
+import com.automattic.simplenote.utils.AccountVerificationWatcher;
 import com.automattic.simplenote.utils.AppLog;
 import com.automattic.simplenote.utils.AppLog.Type;
 import com.automattic.simplenote.utils.CrashUtils;
 import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.PrefUtils;
-import com.automattic.simplenote.utils.AccountVerificationWatcher;
 import com.automattic.simplenote.utils.SyncWorker;
 import com.simperium.Simperium;
 import com.simperium.android.AndroidClient;
@@ -54,10 +58,6 @@ import java.util.Calendar;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
-
-import static com.automattic.simplenote.models.Preferences.PREFERENCES_OBJECT_KEY;
-import static com.simperium.android.AsyncAuthClient.USER_ACCESS_TOKEN_PREFERENCE;
-import static com.simperium.android.AsyncAuthClient.USER_EMAIL_PREFERENCE;
 
 public class Simplenote extends Application implements HeartbeatListener {
     public static final String DELETED_NOTE_ID = "deletedNoteId";
@@ -76,7 +76,6 @@ public class Simplenote extends Application implements HeartbeatListener {
     private static final long HEARTBEAT_TIMEOUT =  WebSocketManager.HEARTBEAT_INTERVAL * 2;
 
     private Activity mCurrentActivity;
-    private AccountVerificationWatcher mAccountVerificationWatcher;
 
     private static Bucket<Account> mAccountBucket;
     private static Bucket<Preferences> mPreferencesBucket;
@@ -130,9 +129,9 @@ public class Simplenote extends Application implements HeartbeatListener {
             mPreferencesBucket = mSimperium.bucket(new Preferences.Schema());
             mAccountBucket = mSimperium.bucket(new Account.Schema());
             // Setup Account Verification Watcher to listen to network changes on the account bucket
-            VerificationListener verificationListener = new VerificationListener();
-            mAccountVerificationWatcher = new AccountVerificationWatcher(this, verificationListener);
-            mAccountBucket.addOnNetworkChangeListener(mAccountVerificationWatcher);
+            mAccountBucket.addOnNetworkChangeListener(
+                    new AccountVerificationWatcher(this, new VerificationListener())
+            );
             // Every time a note changes or is deleted we need to reindex the tag counts
             mNotesBucket.addListener(new NoteTagger(mTagsBucket));
         } catch (BucketNameInvalid e) {
@@ -239,10 +238,6 @@ public class Simplenote extends Application implements HeartbeatListener {
     public String getUserEmail() {
         User user = mSimperium.getUser();
         return user != null ? user.getEmail() : null;
-    }
-
-    public AccountVerificationWatcher getReviewAccountVerifier() {
-        return mAccountVerificationWatcher;
     }
 
     private void showUnverifiedAccount() {
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -7,7 +7,6 @@ import android.content.ComponentCallbacks2;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.res.Configuration;
-import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
@@ -36,7 +35,6 @@ import com.automattic.simplenote.utils.AppLog;
 import com.automattic.simplenote.utils.AppLog.Type;
 import com.automattic.simplenote.utils.CrashUtils;
 import com.automattic.simplenote.utils.DisplayUtils;
-import com.automattic.simplenote.utils.NetworkUtils;
 import com.automattic.simplenote.utils.PrefUtils;
 import com.automattic.simplenote.utils.SyncWorker;
 import com.simperium.Simperium;
@@ -52,11 +50,9 @@ import org.wordpress.passcodelock.AppLockManager;
 
 import java.util.Calendar;
 import java.util.HashMap;
-import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-import static com.automattic.simplenote.models.Account.KEY_EMAIL_VERIFICATION;
 import static com.automattic.simplenote.models.Preferences.PREFERENCES_OBJECT_KEY;
 import static com.simperium.android.AsyncAuthClient.USER_ACCESS_TOKEN_PREFERENCE;
 import static com.simperium.android.AsyncAuthClient.USER_EMAIL_PREFERENCE;
@@ -76,6 +72,8 @@ public class Simplenote extends Application implements HeartbeatListener {
     private static final String AUTH_PROVIDER = "simplenote.com";
     private static final String TAG_SYNC = "sync";
     private static final long HEARTBEAT_TIMEOUT =  WebSocketManager.HEARTBEAT_INTERVAL * 2;
+
+    private Activity mCurrentActivity;
 
     private static Bucket<Account> mAccountBucket;
     private static Bucket<Preferences> mPreferencesBucket;
@@ -242,36 +240,8 @@ public class Simplenote extends Application implements HeartbeatListener {
         return user != null ? user.getEmail() : null;
     }
 
-    private void checkReviewAccountOrVerifyEmail(final Activity activity) {
-        if (isFirstLaunch() || !NetworkUtils.isNetworkAvailable(this) || mHasShownReviewOrVerify) {
-            // Show nothing on first launch, no network connection, or has already shown review/verify.
-            return;
-        }
-
-        Account account;
-
-        try {
-            account = mAccountBucket.get(KEY_EMAIL_VERIFICATION);
-        } catch (BucketObjectMissingException bucketObjectMissingException) {
-            // Show review account when entity does not exist.
-            showReviewAccountOrVerifyEmail(activity, false);
-            return;
-        }
-
-        if (mSimperium == null || mSimperium.getUser() == null || mSimperium.getUser().getEmail() == null) {
-            // Show nothing when Simperium, user, or email cannot be retrieved.
-            return;
-        }
-
-        String email = mSimperium.getUser().getEmail();
-
-        if (account.hasVerifiedEmail(email)) {
-            // Show nothing when email has been verified.
-            return;
-        }
-
-        // Show verify email when email has been sent to account address.  Show review account otherwise.
-        showReviewAccountOrVerifyEmail(activity, account.hasSentEmail(email));
+    public void showReviewVerifyAccount(boolean hasSentEmail) {
+        showReviewAccountOrVerifyEmail(mCurrentActivity, hasSentEmail);
     }
 
     private void showReviewAccountOrVerifyEmail(final Activity activity, boolean hasSentEmail) {
@@ -409,11 +379,12 @@ public class Simplenote extends Application implements HeartbeatListener {
 
         @Override
         public void onActivityCreated(@NonNull Activity activity, Bundle savedInstanceState) {
-            checkReviewAccountOrVerifyEmail(activity);
+
         }
 
         @Override
         public void onActivityStarted(@NonNull Activity activity) {
+            mCurrentActivity = activity;
         }
 
         @Override
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -36,6 +36,7 @@ import com.automattic.simplenote.utils.AppLog.Type;
 import com.automattic.simplenote.utils.CrashUtils;
 import com.automattic.simplenote.utils.DisplayUtils;
 import com.automattic.simplenote.utils.PrefUtils;
+import com.automattic.simplenote.utils.ReviewAccountVerifier;
 import com.automattic.simplenote.utils.SyncWorker;
 import com.simperium.Simperium;
 import com.simperium.android.AndroidClient;
@@ -127,6 +128,7 @@ public class Simplenote extends Application implements HeartbeatListener {
             mTagsBucket = mSimperium.bucket(tagSchema);
             mPreferencesBucket = mSimperium.bucket(new Preferences.Schema());
             mAccountBucket = mSimperium.bucket(new Account.Schema());
+            mAccountBucket.addOnNetworkChangeListener(new ReviewAccountVerifier(this));
             // Every time a note changes or is deleted we need to reindex the tag counts
             mNotesBucket.addListener(new NoteTagger(mTagsBucket));
         } catch (BucketNameInvalid e) {
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -159,11 +159,6 @@ public class Simplenote extends Application implements HeartbeatListener {
         mHeartbeatHandler.postDelayed(mHeartbeatRunnable, HEARTBEAT_TIMEOUT);
     }
 
-    private boolean isFirstLaunch() {
-        // NotesActivity sets this pref to false after first launch
-        return PrefUtils.getBoolPref(this, PrefUtils.PREF_FIRST_LAUNCH, true);
-    }
-
     public static boolean analyticsIsEnabled() {
         if (mPreferencesBucket == null) {
             return true;
--- a/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
@@ -15,6 +15,7 @@ import android.util.Log;
 import androidx.annotation.IdRes;
 import androidx.annotation.NonNull;
 import androidx.appcompat.app.AppCompatDelegate;
+import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
 import androidx.work.BackoffPolicy;
 import androidx.work.Constraints;
@@ -86,7 +87,6 @@ public class Simplenote extends Application implements HeartbeatListener {
     private Handler mHeartbeatHandler;
     private Runnable mHeartbeatRunnable;
     private Simperium mSimperium;
-    private boolean mHasShownReviewOrVerify;
     private boolean mIsInBackground = true;
 
     public void onCreate() {
@@ -129,7 +129,10 @@ public class Simplenote extends Application implements HeartbeatListener {
             mTagsBucket = mSimperium.bucket(tagSchema);
             mPreferencesBucket = mSimperium.bucket(new Preferences.Schema());
             mAccountBucket = mSimperium.bucket(new Account.Schema());
+            // Setup Account Verification Watcher to listen to network changes on the account bucket
             mAccountVerificationWatcher = new AccountVerificationWatcher(this);
+            VerificationListener verificationListener = new VerificationListener();
+            mAccountVerificationWatcher.addVerificationStateListener(verificationListener);
             mAccountBucket.addOnNetworkChangeListener(mAccountVerificationWatcher);
             // Every time a note changes or is deleted we need to reindex the tag counts
             mNotesBucket.addListener(new NoteTagger(mTagsBucket));
@@ -243,8 +246,31 @@ public class Simplenote extends Application implements HeartbeatListener {
         return mAccountVerificationWatcher;
     }
 
-    public void showReviewVerifyAccount(boolean hasSentEmail) {
-        showReviewAccountOrVerifyEmail(mCurrentActivity, hasSentEmail);
+    private void showUnverifiedAccount() {
+        showReviewAccountOrVerifyEmail(mCurrentActivity, false);
+    }
+
+    private void showWaitingOnEmailConfirmation() {
+        showReviewAccountOrVerifyEmail(mCurrentActivity, true);
+    }
+
+    private void dismissReviewAccountDialog() {
+        FragmentManager fragmentManager;
+        if (mCurrentActivity instanceof NotesActivity) {
+            fragmentManager = ((NotesActivity) mCurrentActivity).getSupportFragmentManager();
+        } else if (mCurrentActivity instanceof NoteEditorActivity) {
+            fragmentManager = ((NoteEditorActivity) mCurrentActivity).getSupportFragmentManager();
+        } else {
+            return;
+        }
+
+        for (Fragment fragment : fragmentManager.getFragments()) {
+            if (fragment instanceof FullScreenDialogFragment) {
+                ((FullScreenDialogFragment) fragment).dismiss();
+
+                return;
+            }
+        }
     }
 
     private void showReviewAccountOrVerifyEmail(final Activity activity, boolean hasSentEmail) {
@@ -271,8 +297,6 @@ public class Simplenote extends Application implements HeartbeatListener {
             .setViewContainer(container)
             .build()
             .show(fragmentManager, FullScreenDialogFragment.TAG);
-
-        mHasShownReviewOrVerify = true;
     }
 
     private class ApplicationLifecycleMonitor implements Application.ActivityLifecycleCallbacks, ComponentCallbacks2 {
@@ -435,6 +459,24 @@ public class Simplenote extends Application implements HeartbeatListener {
         @Override
         public void onUpdate(String entityId, Calendar lastSyncTime, boolean isSynced) {
             mPreferences.edit().putLong(entityId, lastSyncTime.getTimeInMillis()).apply();
+        }
+    }
+
+    private class VerificationListener implements AccountVerificationWatcher.VerificationStateListener {
+        @Override
+        public void onUpdate(AccountVerificationWatcher.Status status) {
+            switch (status) {
+                case VERIFIED:
+                    dismissReviewAccountDialog();
+                    return;
+
+                case SENT_EMAIL:
+                    showWaitingOnEmailConfirmation();
+                    return;
+
+                case UNVERIFIED:
+                    showUnverifiedAccount();
+            }
         }
     }
 }
--- a/Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
@@ -1,14 +1,11 @@
 package com.automattic.simplenote.utils;
 
+import static com.automattic.simplenote.models.Account.KEY_EMAIL_VERIFICATION;
+
 import com.automattic.simplenote.Simplenote;
 import com.automattic.simplenote.models.Account;
 import com.simperium.client.Bucket;
 import com.simperium.client.BucketObjectMissingException;
-
-import java.util.ArrayList;
-import java.util.List;
-
-import static com.automattic.simplenote.models.Account.KEY_EMAIL_VERIFICATION;
 
 /**
  * Monitors account verification status by watching the `account` bucket and notifies of status and changes.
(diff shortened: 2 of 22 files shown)


Simplenote/src/main/java/com/automattic/simplenote/utils/AuthUtils.java
```diff
@@ -27,6 +27,9 @@ public static void logOut(Simplenote application) {
         application.getTagsBucket().reset();
         application.getPreferencesBucket().reset();
 
+        // We have to verify the account for the next user
+        application.getReviewAccountVerifier().clearStatus();
```
dmsnell: I see this may be a bigger problem with the design, but I don't like that we have to reset all these objects at logout. I wish instead they were destroyed and recreated when logging in. given that every other bucket is reset here I don't think it's a barrier to this PR, but wanted to ear-mark it for the future.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
```diff
@@ -0,0 +1,72 @@
...
+    public void onNetworkChange(Bucket<Account> bucket, Bucket.ChangeType type, String key) {
+        // If we already know the answer, we avoid computing it again
+        if (currentStatus != Status.UNKNOWN) {
+            return;
+        }
+
+        String email = simplenote.getUserEmail();
+        // When a network change of type INDEX is received, it means that the channel finished indexing the bucket, thus
+        // we can check whether the account is verified or not
+        if (type == Bucket.ChangeType.INDEX && email != null) {
```
dmsnell: it would be nice to abort early here since we're only interested in this one case and we don't need the nesting.

```java
if (type != Bucket.ChangeType.INDEX || email == null) {
  return;
}

…
```

Simplenote/src/main/java/com/automattic/simplenote/utils/AuthUtils.java
danilo04: I think this is the consequence of the current architecture which setups mostly everything when the application (`Simplenote`) is created. Thus, the application keeps the `state` that should be cleaned during logout.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: Fixed in 6bd292286dc5eaf53e8cdc8430b8e0f252558a70.

Simplenote/src/main/java/com/automattic/simplenote/utils/AuthUtils.java
danilo04: @dmsnell I will take this into account as part of the codebase improvement project.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
```diff
@@ -0,0 +1,74 @@
...
+import static com.automattic.simplenote.models.Account.KEY_EMAIL_VERIFICATION;
+
+/**
+ * When a user login into their account, Simplenote starts indexing the different buckets including Bucket<Account>.
+ * We setup a OnNetworkChangeListener on the account bucket to listen when the bucket is indexed. One of the parameters
+ * of the callback onNetworkChange called is Bucket.ChangeType. If this parameter is INDEX, then we know that the
+ * account bucket was indexed (synchronized). When the account bucket is synchronized, we can check whether the account
+ * was verified or not
+ */
+public class ReviewAccountVerifier implements Bucket.OnNetworkChangeListener<Account> {
```
dmsnell: The name here is curious to me. I don't know if I'm missing something, but it seems like the purpose of this class is to watch for verification of an account and report that. To that end I might expect to see something like `AccountVerificationWatcher` or `AccountVerificationListener`

Any thoughts on the name or explanation?

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
```diff
@@ -0,0 +1,74 @@
...
+
+import static com.automattic.simplenote.models.Account.KEY_EMAIL_VERIFICATION;
+
+/**
+ * When a user login into their account, Simplenote starts indexing the different buckets including Bucket<Account>.
+ * We setup a OnNetworkChangeListener on the account bucket to listen when the bucket is indexed. One of the parameters
+ * of the callback onNetworkChange called is Bucket.ChangeType. If this parameter is INDEX, then we know that the
+ * account bucket was indexed (synchronized). When the account bucket is synchronized, we can check whether the account
+ * was verified or not
+ */
```
dmsnell: In light of the implementation of the class, do you think all of this belongs here as class documentation? Could we explain the behavior and intention of the class differently? It reads like we're explaining the implementation of the class but not its purpose.

```java
/**
 * Monitors account verification status by watching the `account` bucket and notifies of status and changes.
 * 
 * When connecting this monitor will wait until we have received positive confirmation from the
 * server that we're sync'd before reporting the verification status because the account might
 * have (and probably was) updated from another device.
 */
```

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
```diff
@@ -0,0 +1,74 @@
...
+    public ReviewAccountVerifier(Simplenote simplenote) {
+        this.simplenote = simplenote;
+    }
+
+    public void clearStatus() {
+        currentStatus = Status.UNKNOWN;
+    }
+
+    @Override
+    public void onNetworkChange(Bucket<Account> bucket, Bucket.ChangeType type, String key) {
```
dmsnell: this comes from my own personal style but I like to avoid implementing these interfaces directly on classes that contain their own behavior. that is, by removing the boundary between the account review behavior and the network listener behavior. this is a small class so it won't make too much of an impact here, but for example, in [SyncTimes](https://github.com/Automattic/simplenote-android/blob/develop/Simplenote/src/main/java/com/automattic/simplenote/SyncTimes.java#L60) I created a listener as a property of the class and [connected its listener](https://github.com/Automattic/simplenote-android/blob/develop/Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java#L126) vs. connecting itself as `this`. I found this quite helpful in separating the code into pieces that felt independent from each other.

```java
mVerificationWatcher = new AccountVerificationWatcher(this);
mAccountBucket.addOnNetworkChangeListener(mVerificationWatcher.bucketListener);
```

this kind of separation duly pays off when we want to have a class listen to more than one bucket but has its own benefits otherwise by splitting the "new event from the bucket" logic from the "state of the verification status" logic.

what do you think about this? I'm curious what you think about it.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
```diff
@@ -0,0 +1,74 @@
...
+        // If we already know the answer, we avoid computing it again
```
dmsnell: for boot this would be good enough but we probably also want to monitor the bucket for changes. it's definitely possible to de-verify an account so we might get change events of `MODIFY` or `REMOVE` that would update the verification status.

it's also a trivial computation so there's no need to avoid it.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
```diff
@@ -0,0 +1,74 @@
...
+        if (currentStatus != Status.UNKNOWN) {
+            return;
+        }
+
+        String email = simplenote.getUserEmail();
+        if (type != Bucket.ChangeType.INDEX || email == null) {
+            return;
```
dmsnell: see previous note: we _do_ want to monitor other changes which could update the verification status. we might have started the app without verification, shown the screen, but while the screen is open we get an update indicating that the account is now verified. let's listen for those other events.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        // When a network change of type INDEX is received, it means that the channel finished indexing the bucket, thus
+        // we can check whether the account is verified or not
```
dmsnell: side note: it's possible at other times, even _most_ times to check on the verification status. what's important for us here is that we don't check the status before _the first index_ operation or before we positively synchronize with the server. I believe that the Android library _always_ indexes at boot but this is strictly not required by the Simperium protocol.

the primary thing we want to avoid is nagging people when they've already verified because that's annoying and confusing, which is why we wait until we can confidently say "we have all recent changes to the account bucket and the value we read is as good as we can expect."

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
```diff
@@ -0,0 +1,74 @@
...
+                    public void run() {
+                        simplenote.showReviewVerifyAccount(hasSentEmail);
+                    }
+                });
+            } else {
+                currentStatus = Status.VERIFIED;
+            }
+        } catch (BucketObjectMissingException e) {
+            AppLog.add(AppLog.Type.SYNC, "Account for email verification is missing");
+        }
```
dmsnell: superficially here it'd be nice if we limited the `try`/`catch` to surrounding the statements that could crash, also of early-aborting where possible to avoid needless nesting. in this case it would be something like the following…

```java
Account account;
try {
    account = bucket.get(KEY_EMAIL_VERIFICATION);
  AppLog.add(…);
  return;
}

if (account.hasVerifiedEmail(email)) {
  currentStatus = Status.VERIFIED;
  return;
}

currentStatus = Status.UNVERIFIED;
…
```

dmsnell: on a deeper level I think there are a few challenges with this section of code and the way we're coupling the behavior of the verification monitor to the broader application, especially the `looper`/`handler` lines.

what we've done is push the implementation of the app's logic into this helper class and in doing so overlooked an opportunity to abstract: to define some new names of terms and behaviors. if we go back and ask ourselves what we're want to do, essentially, then I think it might lead us to a few changes that would clear up the code _and_ decouple these bits.

if I can reference `SyncTimes` again we did the same kind of thinking there and [created a new interface](https://github.com/Automattic/simplenote-android/blob/68efd052b29eb88a9e2bbe4476700d91acab5a5e/Simplenote/src/main/java/com/automattic/simplenote/SyncTimes.java#L122-L125) for parent/host/calling classes to implement. by doing this and keeping the internal details of the modules decoupled we retain control in "the caller" (in this case, the `Simperium` class).

```java
interface VerificationStateListener {
    void onUpdate(Status status);
}
```

if we think about the verification status though `onUpdate` we can see that we don't actually need to export boolean states for `hasSentEmail` or probably even track `mHasShownReviewOrVerify`. let's consider a subtle change in our `Status` enum to focus on the verification process more than the data in the `account` bucket.

```java
public enum Status {
    Verified,
    SentEmail,
    Unverified
}
```

as far as we care, we only need to know these states. we can track `Unknown` but it's also a value we don't need to export, so maybe it's possible to use a `null` for the internal status tracker inside our monitor. I don't know, `Unknown` isn't a bad value.

```java
// Simplenote.java

mVerificationWatcher = new AccountVerificationWatcher(this);
mAccountBucket.addOnNetworkChangeListener(mVerificationWatcher.bucketListener);
mVerificationWatcher.addListener(this.verificationListener);

private class VerificationListener implements AccountVerificationWatcher.VerificationStateListener {
    @Override
    public void onUpdate(AccountVerificationWatcher.Status status) {
        switch (status) {
            case VERIFIED:
                closeNag();
                return;

            case SENT_EMAIL:
                showWaitingOnEmailConfirmation();
                return;

            case UNVERIFIED:
                showUnverifiedWarning();
                return;
        }
    }
}
```

it might seem verbose, but it lets us treat each class independently and it removes our need to track state. we know that `onUpdate` won't be called unless the status _changes_ and that also means it won't be called until we get our first update from the server.

with this alternate design I think it will be clearer to watch the flow of data in the bucket and make it easier not to introduce another bug because of updates coming across after (or before) indexing operations complete.

dmsnell (commented):
Left a few comments but the big one I think worth most consideration is the coupling we have between `Simplenote.java` and `ReviewAccountVerifier`. I'd really like to get away from some of this deep coupling that's somewhat prevalent within the application because I think it tends to muddy the control and data flow. You have to reach into `Simplenote.java` to then reach into `ReviewAccountVerifier.java` only then to know to reach back into `Simplenote.java` to understand what happens when the `account` bucket changes.

If you are willing to consider that and respond or attempt to change the design then I think we have some things to review that are less impactful but important from a correctness perspective: we're only watching for one account verification change when in fact we should be continuously looking for changes.

This is looking good and I appreciate your care when putting this together. It was particularly good seeing how you wrapped `showReviewVerifyAccount` so that we could avoid coupling in the `mCurrentActivity` into the update function. It would pay off if we push that value a bit further by separating the app update from the insides of the verification monitor class.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: Agree. Fixed in 79d4813bf85e462852237f08a6c4800f50a713d5.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: I changed it to `AccountVerificationWatcher`. Fixed in 79d4813bf85e462852237f08a6c4800f50a713d5.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: > superficially here it'd be nice if we limited the try/catch to surrounding the statements that could crash, also of early-aborting where possible to avoid needless nesting. in this case it would be something like the following…

I personally don't like this style because it becomes too choppy. The first check `if (type != Bucket.ChangeType.INDEX || email == null)` we cut it earlier because it is more of a constraint, but if we split the algorithm in parts, IMO It makes the logic behind the method harder to understand. So, I prefer the method to flow as it is right now since it just has two paths (low cyclomatic complexity) and makes the logic simpler.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: Question(❓ ) @dmsnell : if we get a `REMOVE` event, does the account will be available if we try to get it with:
```java
Account account = bucket.get(KEY_EMAIL_VERIFICATION);
```

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: > on a deeper level I think there are a few challenges with this section of code and the way we're coupling the behavior of the verification monitor to the broader application, especially the looper/handler lines.

One way we can solve this in a simple way is moving the handler code inside `Simplenote` which has the knowledge about the current activity and can use the activity to call the method `runOnUIThread` instead. However, I see your previous comments and this one have some concerns about the whole architecture and coupling. Let me address those comments before making this change.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: Agree. Let me just be sure the events we are going to handle and make the complete change.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
dmsnell: > does the account will be available if we try to get it with:

I think you're asking if the verification entity will still be available in the bucket if there's a `REMOVE` event. if we _do_ get that event then we'd still have to check. if `key` is the same as `KEY_EMAIL_VERIFICATION` then no, we wouldn't expect the property to exist and we would immediately know that we are now `UNVERIFIED`. if it's any other value then we can ignore the event because it was some other entity that was removed.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
dmsnell: > I personally don't like this style because it becomes too choppy.

That's fair. I think our Android code is full of nesting but that's uncommon in our other projects. Most of the time we prefer early-aborting to avoid nesting.

> if we split the algorithm in parts, IMO It makes the logic behind the method harder to understand

just be careful not to make adverse decisions because of some change we might make in the future but have no plans to make. there are ways of handling changes when they come, in a case like this many times we find that instead of adding more nesting levels, what we end up discovering is that we have a dispatcher and multiple methods to handle each dispatch.

> So, I prefer the method to flow as it is right now since it just has two paths (low cyclomatic complexity)

This nested structure has higher cyclomatic complexity three paths. This is why I like avoiding the nesting and early-aborting, because the control flow isn't deceiving as it can be in cases like this 😉. We know that once we hit that `return` statement we're done. Here, we have to remember that the `try`/`catch` interacts with the conditionals.

If we narrow the `try`/`catch` so that it only wraps `bucket.get()` then we avoid increasing the cyclomatic complexity but we do have to remember to `return` from the `catch` clause or else we'll continue to run the verification check even if the entity is missing.

> makes the logic simpler

I'm not sure this true. It is probably more familiar to your style, but in order to simply the logic we have to remove branches or variables. in this case I'm pretty sure it's mostly a style preference apart from the fact that we have to remember to skip to the end of the `try`/`catch` to see if there's any further processing in the method (whereas with a `return` we can know without looking any further, "this is where execution ends.")

> One way we can solve this in a simple way is moving the handler code inside Simplenote which has the knowledge about the current activity and can use the activity to call the method runOnUIThread instead. 

If that means adding the listener code in `Simperium.java` then that's definitely not what I'm recommending. I like that you created a separate class/object to watch the verification status and that is a positive move for the project. We're already way too-coupled in places and our classes typically do way too much. Isolating behaviors into independent and semantically-narrow classes is a good thing here.

> concerns about the whole architecture and coupling. Let me address those comments before making this change.

👍 awesome. I think that with some rethinking of the `AccountVerificationWatcher` interface we will end up in an easier spot; we won't have to answer some of the current questions about architecture because they will be solved by construction for us.

danilo04: > If that means adding the listener code in Simperium.java then that's definitely not what I'm recommending.

No, this is not what I meant. But is futile to discuss it if we are going to change the architecture. 

> We know that once we hit that return statement we're done. Here, we have to remember that the try/catch interacts with the conditionals.

We could talk ages about style, but in the end, it is getting subjective since we do not have a cognitive study to support our claims 🤣 . 

> We're already way too-coupled in places and our classes typically do way too much. Isolating behaviors into independent and semantically-narrow classes is a good thing here.

I definitely agree with your sentiment and I think the architecture you presented for account verification is great (let me prepare the changes, I will have them shortly). I just want to be careful of applying half-patches improvements that when if thought better, we could make something really simpler with higher quality. For example, you mentioned after logout we should not make all those calls manually (they should be part of the process itself). In that case, we could go a try to move that code, but if we focus on designing a holistic solution for session management, we might end up with a better solution. There are some spots that need some work and I really appreciate that from this small PR you have identified more than two spots for improvements. But I think we could handle those cases in separate PRs.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: > if key is the same as KEY_EMAIL_VERIFICATION then no, we wouldn't expect the property to exist and we would immediately know that we are now UNVERIFIED

That defines the logic for this event type. Thanks!

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
dmsnell: > We could talk ages about style, but in the end, it is getting subjective since we do not have a cognitive study to support our claims 🤣 .

Note that I'm not saying one is better than the other. It's not a subjective claim I'm making: I'm simply pointing out that with early-abort patterns we are able to safely turn our minds off _at the point where we return_ because it's impossible for there to be continuing behaviors whereas with the nesting we aren't allowed to do that - we have to continue scanning until the end of the block to check if there is more code that might still run.

> I just want to be careful of applying half-patches improvements that when if thought better, we could make something really simpler with higher quality.

I'm not talking about restructuring the app in this PR. I'm saying instead that we should not feel bound to a pattern that's prevalent in the app which has caused headaches, that we are free to make new things with a better design than existing things have.

What are the half-patches you're talking about?

> For example, you mentioned after logout we should not make all those calls manually (they should be part of the process itself).

Specifically in that comment I mentioned that it's not work for this PR since it involves a larger redesign.

> In that case, we could go a try to move that code, but if we focus on designing a holistic solution for session management, we might end up with a better solution. There are some spots that need some work and I really appreciate that from this small PR you have identified more than two spots for improvements. But I think we could handle those cases in separate PRs.

I'm confused on what you're attempting to say here. Can you reword this or maybe be more specific about what you're talking about? In my own suggestions I'm _only_ talking about the design points of the account verification watcher and not of changing anything else in the app. If I've communicated doing broader work as a requirement here that was me miscommunication; I mention parts of the app as _motivation_ for design on this change, not as a dependency before it.

dmsnell:
@danilo04 hiya 👋 I'm adding this comment here to try and help avoid getting off-track in our discussion in the code. this is good work! there are two basic things I am recommending for the PR:

 - I think it's worth our time not to add a new module that strongly couples cross-cutting concerns. specifically that we hook into the new `AccountVerificationWatcher` from `Simplenote` so that the call to show the verification nag remains in `Simplenote` and `AVW` remains agnostic about what happens _when_ verification status changes. this can happen in different ways, `onUpdate` is one possible way to do that.
 - we need to track the `account` bucket and watch for changes since we expect there to be times when the verification status changes while the app is open and syncing

you don't need to feel bound to the style comments. while I'm not going to ignore the observation or discussion (I'm happy to hold that conversation) it's not central to the needs of this PR.

Simplenote/src/main/java/com/automattic/simplenote/utils/ReviewAccountVerifier.java
danilo04: > I'm not talking about restructuring the app in this PR. I'm saying instead that we should not feel bound to a pattern that's prevalent in the app which has caused headaches, that we are free to make new things with a better design than existing things have.

Agree. I think the improvements you presented leave the PR in a better position from a design point of view.

> In my own suggestions I'm only talking about the design points of the account verification watcher and not of changing anything else in the app.
> you don't need to feel bound to the style comments. while I'm not going to ignore the observation or discussion (I'm happy to hold that conversation) it's not central to the needs of this PR.

Yes, I think this is the part the confuses me a little bit about your review. I think I made them too much part of the conversation which got mixed with the rest of the comments that define the behavior and better architecture.

danilo04:
Hi 👋  @dmsnell. I have changed the architecture of the account verification dialog in 569bf83. Thanks for the suggestions. Let me know if you find something to change.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -437,4 +461,22 @@ public void onUpdate(String entityId, Calendar lastSyncTime, boolean isSynced) {
             mPreferences.edit().putLong(entityId, lastSyncTime.getTimeInMillis()).apply();
         }
     }
+
+    private class VerificationListener implements AccountVerificationWatcher.VerificationStateListener {
+        @Override
+        public void onUpdate(AccountVerificationWatcher.Status status) {
+            switch (status) {
+                case VERIFIED:
+                    dismissReviewAccountDialog();
```
dmsnell, line 470: are these wrapper functions necessary or could we inline their calls, given that this all happens from inside `Simplenote` now?

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -437,4 +461,22 @@ public void onUpdate(String entityId, Calendar lastSyncTime, boolean isSynced) {
...
+                    return;
+
+                case SENT_EMAIL:
+                    showWaitingOnEmailConfirmation();
+                    return;
+
+                case UNVERIFIED:
+                    showUnverifiedAccount();
```
dmsnell, line 471: are these wrapper functions necessary or could we inline their calls, given that this all happens from inside `Simplenote` now?

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -18,57 +18,63 @@
...
 
     public AccountVerificationWatcher(Simplenote simplenote) {
         this.simplenote = simplenote;
     }
 
-    public void clearStatus() {
-        currentStatus = Status.UNKNOWN;
+    public void addVerificationStateListener(VerificationStateListener listener) {
+        listeners.add(listener);
     }
```
dmsnell, line 44: I should have thought about this yesterday, but _if_ you find this a bit overbearing with the listeners it's probably good enough to store just a single listener and pass it in when instantiating this thing. that would result in far less plumbing.

```java
private VerificationStateListener listener;

public AccountVerificationWatcher(Simplenote simplenote, VerificationStateListener listener) {
    this.simplenote = simplenote;
    this.listener = listener;
}
```

it seems unlikely we'll be adding more than one.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -18,57 +18,63 @@
...
+        for (VerificationStateListener listener : listeners) {
+            listener.onUpdate(status);
         }
+    }
+
+    @Override
+    public void onNetworkChange(final Bucket<Account> bucket, Bucket.ChangeType type, String key) {
+        // If the account is removed, the status is changed to UNVERIFIED immediately since the account will not be
+        // available
+       if (type == Bucket.ChangeType.REMOVE) {
```
dmsnell: I am about 60% confident that this change isn't saying "the bucket has been removed" but rather "the entity with key `key` has been removed." it's possible that someone might add or remove another entity in this bucket. currently inside the `account` bucket we use the entity whose entity id is `email-verification` to store this information. that would make this check look a little different…

```java
if (type == Bucket.ChangeType.REMOVE && key == KEY_EMAIL_VERIFICATION) {
    notifyListeners(Status.UNVERIFIED);
    return;
}
```

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -18,57 +18,63 @@
...
+            notifyListeners(Status.UNVERIFIED);
+           return;
+       }
```
dmsnell: nitpick: looks like the indentation is off by one here

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -18,57 +18,63 @@
...
 
         String email = simplenote.getUserEmail();
-        if (type != Bucket.ChangeType.INDEX || email == null) {
+        if ((type != Bucket.ChangeType.INDEX && type != Bucket.ChangeType.MODIFY) || email == null) {
```
dmsnell: unfortunately we want to monitor for `INSERT` here too as well. we can safely ignore `RESET`. we might want to change these conditionals as a result.

```java
if (type == Bucket.ChangeType.RESET || email == null) {
    return;
}

// get the entity

// set verification based on presence and validity of entity
```

dmsnell: I'm worried we'll have to track the status to avoid notifying on changes if we haven't changed status. if we have to avoid double-sending the notification then I guess we go back to what you had, but I think we can still focus on the `onChange` interface to keep that state-tracking in this module, to avoid exposing `UNKNOWN`

dmsnell:
When it comes time to test we can probably do so most easily with `curl` and some calls directly to Simperium. Hit me up on Slack and we can discuss the details. ✅

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
danilo04: @dmsnell Question(❓ ): currently in every update of `onNetworkChange` I see that `key` is `null` (mostly for `RESET` and `INDEX` events). Maybe because those are general events instead of events on particular keys such as `REMOVE`?

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
danilo04: I would leave them since they make the code clearer, but I see your point since they are just private methods and not part of the public interface.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
danilo04: > unfortunately we want to monitor for INSERT here too as well. we can safely ignore RESET. we might want to change these conditionals as a result.

Question(❓ ): Will the `key` will not be `null` for `INSERT` as well as for `REMOVE` (question above)?

danilo04: > I'm worried we'll have to track the status to avoid notifying on changes if we haven't changed status. if we have to avoid double-sending the notification then I guess we go back to what you had, but I think we can still focus on the onChange interface to keep that state-tracking in this module, to avoid exposing UNKNOWN

We can keep the status (`currentState`) and just notify the listener when the status changes: `if (newState != currentState)` with `currentState = UNKNOWN` as default value.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
dmsnell: > since they make the code clearer

:trollface: well, they make me jump around the file to see what they do 🤷‍♂️

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
dmsnell: that's right. `key` will be a string value, the `entity_id` of an entity, for those network events which refer to an entity. these would be `REMOVE`, `INSERT`, `MODIFY`. `RESET` is kind of a made-up event that's not coming from the network but is part of the Android Simperium library, used to reset bucket storage because of the aforementioned issues where these aren't being destroyed on signet.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
dmsnell: > Will the key will not be null for INSERT as well as for REMOVE (question above)?

Easy to test 😄 

But no they won't be `null` - they will have the `entity_id` as a string value. Wouldn't make much sense to always and only pass `null` as an argument I suppose…

danilo04: Yes, that is what I thought later, put a listener on the notes bucket and observe the behavior on`INSERT` and `MODIFY`. 

PS: I didn't see `INSERT` nor `MODIFY` events in notes. When do those events happen?
PS2: I triggered modifying the note in another device. It is clear to me know

danilo04: > Wouldn't make much sense to always and only pass null as an argument I suppose…

The interface is handling two types of events, one for the whole bucket and one for specific keys. We could improve the interface and apply the `Interface Segregation Principle` since this interface may be doing more than one thing 😜

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
dmsnell: thumbs down?

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
dmsnell: > The interface is handling two types of events, one for the whole bucket and one for specific keys.

Yes, this is a weakness of the current design. It's actually because of issues like this why I put my attention into code reviews to encourage that we deeply consider the new interfaces we add so as to make a positive turn in the app's design.

danilo04:
Hi 👋  @dmsnell. I have added the requested changes as discussed. Also, one way I am using for testing is injecting network changes like this: 

```java
@Override
    public void onNetworkChange(final Bucket<Account> bucket, Bucket.ChangeType type, String key) {
        // If the key for email verification is removed, the status is changed to UNVERIFIED immediately
       if (type == Bucket.ChangeType.REMOVE && KEY_EMAIL_VERIFICATION.equals(key)) {
            notifyListener(Status.UNVERIFIED);
           return;
       }

        String email = simplenote.getUserEmail();
        if (!isValidChangeType(type, key) || email == null) {
            return;
        }

        Account account;
        try {
            // When a network change of type INDEX or MODIFY is received, it means that the account bucket finished
            // indexing or there were some changes in the account. In both cases, we need to check for the account status
            account = bucket.get(KEY_EMAIL_VERIFICATION);
            AppLog.add(AppLog.Type.SYNC, "Account for email verification is missing");
            return;
        }

        boolean hasVerifiedEmail = type == Bucket.ChangeType.INDEX; //account.hasVerifiedEmail(email);
        if (hasVerifiedEmail) {
            updateState(Status.VERIFIED);
        } else {
            Status statusUpdate = account.hasSentEmail(email) ? Status.SENT_EMAIL : Status.UNVERIFIED;
            updateState(statusUpdate);
        }

        Handler handler = new Handler(Looper.getMainLooper());
        handler.postDelayed(new Runnable() {
            @Override
            public void run() {
                onNetworkChange(bucket, Bucket.ChangeType.MODIFY, KEY_EMAIL_VERIFICATION);
            }
        }, 4000);
    }
```

If I want to test two different behaviors, I just change the equality in `boolean hasVerifiedEmail = type == Bucket.ChangeType.INDEX;`

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
danilo04: Double 👍  to make it even 🤣

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -29,33 +27,44 @@
...
+            notifyListener(newState);
         }
     }
 
+    private boolean isValidChangeType(Bucket.ChangeType type, String key) {
+        return type == Bucket.ChangeType.INDEX ||
+                (type == Bucket.ChangeType.INSERT && KEY_EMAIL_VERIFICATION.equals(key)) ||
+                (type == Bucket.ChangeType.MODIFY && KEY_EMAIL_VERIFICATION.equals(key));
+
+    }
```
dmsnell, line 50: is it really "valid" that we care about? all changes are "valid" because they are coming from the Simperium library. maybe we're looking for something like `affectsVerification` or `involvesVerification`?

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -29,33 +27,44 @@
...
+            currentState = newState;
+            notifyListener(newState);
         }
     }
 
+    private boolean isValidChangeType(Bucket.ChangeType type, String key) {
+        return type == Bucket.ChangeType.INDEX ||
+                (type == Bucket.ChangeType.INSERT && KEY_EMAIL_VERIFICATION.equals(key)) ||
+                (type == Bucket.ChangeType.MODIFY && KEY_EMAIL_VERIFICATION.equals(key));
+
```
dmsnell: nitpick: added space here

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -29,33 +27,44 @@
...
+        if (newState != currentState) {
+            currentState = newState;
+            notifyListener(newState);
         }
     }
 
+    private boolean isValidChangeType(Bucket.ChangeType type, String key) {
+        return type == Bucket.ChangeType.INDEX ||
+                (type == Bucket.ChangeType.INSERT && KEY_EMAIL_VERIFICATION.equals(key)) ||
+                (type == Bucket.ChangeType.MODIFY && KEY_EMAIL_VERIFICATION.equals(key));
```
dmsnell, line 49: this is fine, but it's also reducible. we know that if we have a matching `key` then it means that the verification object was involved in the change. if not, the only bucket-wide change we care about is `INDEX`

```java
return KEY_EMAIL_VERIFICATION.equals(key) || type == Bucket.ChangeType.INDEX;
```

not important, just wanted to point it out. yeah I know, this also treats `REMOVE` as impacting `verification`.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
danilo04: I would say that is valid inside the context of `AccountVerificationWatcher`, so we care about this event.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
danilo04: you mean this?
```java
return type ...
       (type ...
```

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -29,33 +27,44 @@
...
         this.simplenote = simplenote;
+        this.listener = listener;
     }
 
-    public void addVerificationStateListener(VerificationStateListener listener) {
-        listeners.add(listener);
+    private void notifyListener(Status status) {
+        if (listener != null) {
+            listener.onUpdate(status);
+        }
```
dmsnell: since we set this in the constructor do we need the `null` check?

if we do, would it help to add `@NonNull VerificationStateListener listener` as the constructor arg? We use that around the app already.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
```diff
@@ -19,6 +16,7 @@
  */
 public class AccountVerificationWatcher implements Bucket.OnNetworkChangeListener<Account> {
     public enum Status {
+        UNKNOWN,
```
dmsnell: Nothing that this exposes `UNKNOWN` to the listeners. While it's not a problem to ignore that, how do you feel about introducing `UNKNOWN` and exposing it vs. storing `null` for "unknown" and only exposing the known states?

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
dmsnell: okay. "valid" is so meaning-devoid I usually like to question it. "what is 'valid'?" I only meant to point out that every call to this function is valid in the sense that they are legitimate calls, which is what I jumped to when I read "valid"

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
dmsnell: oops, extra newline is what I meant.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
danilo04: Yeah, let me add that annotation to remove the null check.

dmsnell:
thanks for going back and removing the `addListener` code I suggested 🙃 I like the simplification. also, please know that whenever I do recommend things, even actual code examples, I almost never check if they are right or work. I usually use them as recommendations for avenues to follow, to illustrate some concept.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
danilo04: Exahustiveness in pattern matching (if conditions in Java) is important and as you mention, it is not checked in the listener. Let me use `null` which does not affect the rest of the code.

danilo04: Fixed in 5d421ed.

Simplenote/src/main/java/com/automattic/simplenote/utils/AccountVerificationWatcher.java
danilo04: Fixed in 5d421ed.

dmsnell:
@danilo04 I added e66dca7 and 86af0f1a because I didn't think they were worth asking you to change when I could just do it myself. please feel free to wipe out that commit if you don't like what I did.

 - fixed some small insignificant whitespace issues
 - turned `isValidChangeType` into a static method since it doesn't need to be an instance method

danilo04:
Hi 👋  @dmsnell. I have tested the last version with different interleaving of events on the bucket including:


| Initial State      | Event |  Expected Behavior  | 
| ----------- | ----------- | ----------- |
| UNVERIFIED  | INDEX (login)       | show review account dialog |
| VERIFIED  | INDEX (login)       | don't show anything |
| UNVERIFIED  | MODIFY/INSERT (correct email)       | close review account dialog if it is showing or don't do anything otherwise |
| VERIFIED  | MODIFY/INSERT (different email)       | show review account dialog |
| UNVERIFIED  | DELETE     | don't show anything  |
| VERIFIED  |  DELETE    | show review account dialog |

I have fixed two bugs related to how to dismiss the dialog. Please, let me know if there is something else to change or test.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -242,36 +241,35 @@ public String getUserEmail() {
...
     }
 
-    private void checkReviewAccountOrVerifyEmail(final Activity activity) {
-        if (isFirstLaunch() || !NetworkUtils.isNetworkAvailable(this) || mHasShownReviewOrVerify) {
-            // Show nothing on first launch, no network connection, or has already shown review/verify.
-            return;
-        }
+    public AccountVerificationWatcher getReviewAccountVerifier() {
+        return mAccountVerificationWatcher;
+    }
```
dmsnell: this function is never called, right? we don't need it?

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
```diff
@@ -129,6 +129,10 @@ public void run() {
             mTagsBucket = mSimperium.bucket(tagSchema);
             mPreferencesBucket = mSimperium.bucket(new Preferences.Schema());
             mAccountBucket = mSimperium.bucket(new Account.Schema());
+            // Setup Account Verification Watcher to listen to network changes on the account bucket
+            VerificationListener verificationListener = new VerificationListener();
+            mAccountVerificationWatcher = new AccountVerificationWatcher(this, verificationListener);
+            mAccountBucket.addOnNetworkChangeListener(mAccountVerificationWatcher);
```
dmsnell: given that we don't use any of these apart from starting the whole chain, would you consider not storing them as variables?

```java
mAccountBucket.addOnNetworkChangeListener(new AccountVerificationWatcher(this, new VerificationListener()));
```

danilo04: Fixed in f7793d3831cc70952383bbf2c9307d41394b1839.

Simplenote/src/main/java/com/automattic/simplenote/Simplenote.java
danilo04: Yes, it is dead code. Fixed in f7793d3831cc70952383bbf2c9307d41394b1839.

dmsnell (approved):
Thanks for all the good work on this @danilo04 

There's nothing left I see to mention on this review. It's ready for all the testing, which hopefully will be smooth since you already tested it.

PR #1430, opened 2021-08-02
Move secrets debug keystore and Google upload credentials outside of the repository

mokagio:
Counter part to https://github.com/wordpress-mobile/WordPress-Android/pull/14995 and https://github.com/woocommerce/woocommerce-android/pull/4512

Moves the low-hanging fruits in terms of secrets outside of the repository:

- Fastlane's `google-upload-cerdentials.json`
- The debug keystore

## To test

Checkout this branch and run `./gradlew configureApply`, then `ls ~/.configure/simplenote-android/secrets`, you should see:

```
➜ ls ~/.configure/simplenote-android/secrets

debug_a8c.keystore             google-upload-credentials.json
```

Also ensure that `debugStoreFile` in the local `gradle.properties` has been updated to the new path from `~/.configure/simplenote-android/secrets/`.

**Individual tests**

**`debug_a8c.keystore`**: This file is used to sign the debug build. We can verify it works locally by applying the following patch:

```diff
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -171,7 +171,7 @@ android.applicationVariants.all { variant ->
 
 if (project.hasProperty("debugStoreFile")) {
     def sharedDebugStore = file(project.debugStoreFile.replaceFirst("^~", System.getProperty("user.home")))
-    if (sharedDebugStore.exists()) {
+    // if (sharedDebugStore.exists()) {
         android {
             signingConfigs {
                 debug {
@@ -179,5 +179,5 @@ if (project.hasProperty("debugStoreFile")) {
                 }
             }
         }
-    }
+    // }
 }
```

After the patch, you should:

1. Run `./gradlew configureApply` to ensure the `gradle.properties` local secret is updated.
2. Remove the keystore with `rm ~/.configure/simplenote-android/secrets/debug_a8c.keystore`. 
3. Run `./gradlew assembleDebug` this should fail because the keystore can't be found and ensure that the repo configuration reads the keystore from the new location.
4. Finally, run `./gradlew configureApply` to restore the secret and `./gradlew assembleDebug` should now pass.

**`google-upload-credentials.json`**: This file is used by the `upload_to_play_store` Fastlane action. I checked the action [documentation](https://docs.fastlane.tools/actions/upload_to_play_store/) and couldn't find a way to perform a dry-run. I saw there is a `validate_only` option, but the documentation says "Only validate changes with Google Play rather than actually publish," which doesn't make it clear to me whether data would be uploaded or not.


.circleci/config.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - android/save-gradle-cache
       - android/save-test-results
   Lint:
-    executor: 
+    executor:
```
mokagio, line 23: I opened the file to look for the step that copies the debug keystore like WordPress and Woo do, but couldn't find it.

My editor removed these trailing whitespaces and I thought I'd live the change in for the better 😄

.configure
```diff
@@ -1,7 +1,7 @@
 {
   "project_name": "simplenote-android",
   "branch": "trunk",
-  "pinned_hash": "22b584c6e99946ef9fb20ce62b8f729c5bea16b7",
+  "pinned_hash": "1fa6315925665d577d1570adb56906c4ed47bc2c",
```
mokagio, line 4: This updates `gradle.properties` to include the new location of the debug keystore.

I left a comment about this [here](https://github.com/woocommerce/woocommerce-android/pull/4512#issuecomment-890732951), too.

fastlane/Fastfile
```diff
@@ -43,6 +43,8 @@ SUPPORTED_LOCALES = [
   { glotpress: "zh-tw", google_play: "zh-TW",  promo_config: {} },
 ].freeze
 
+UPLOAD_TO_PLAY_STORE_JSON_KEY = File.join(Dir.home, '.configure', 'simplenote-android', 'secrets', 'google-upload-credentials.json')
```
mokagio, line 46: I didn't really need to extract the path in this constant, because it's used only once. Still, it keeps the code below clearer and it will make it easier to reuse in the future.

AliSoftware (approved):
- ✅  Checkout this branch and run `./gradlew configureApply`, then `ls ~/.configure/simplenote-android/secrets`, you should see: […]
- ✅ ensure that `debugStoreFile` in the local (Olivier: `Simplenote/`)`gradle.properties` has been updated to the new path 
:shipit:

PR #1433, opened 2021-08-04
Fix out of memory error on CI

loremattei:
### Fix
This PR raises the amount of memory available to the JVM in order to fix an out of memory error during the build.

### Test
Check CI is green.

### Review

### Release


danilo04 (approved):
LGTM!

PR #1436, opened 2021-08-05
Update the plus icon to add new notes

### LINKED ISSUE #1434: Update new note icon
Update the icon to add a new note using new designs.

Internal link: pd8BQf-p-p2

### What
Change the plus icon to add new notes for the icon shown below:

android-1

### Why
Part of project to improve Simplenote

### How
Replace `plus` icon by the new note icon.

danilo04:
Fixes #1434 

### Fix

Update the icon to add a new note using new designs.

### Test

1. Go to main screen (list of notes)
2. Tap on add note which should have the new icon

**Phone**

| Light      | Dark |
| ----------- | ----------- |
|      |    |

**Table**

| Light      | Dark |
| ----------- | ----------- |
|   |  |

### Review

Only one designer is required to review these changes. @SylvesterWilmott can review this PR.

### Release

`RELEASE-NOTES.txt` was updated in f3023a97d57dcba5ce9a3c52505a7ad3d640452d with:
> Changed the new note icon

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+2.20
+-----
+* Changed the new note icon [https://github.com/Automattic/simplenote-android/pull/1436]
+
 2.19
 -----
 * Added support to import notes [https://github.com/Automattic/simplenote-android/pull/1333]
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,13 +1,10 @@
-2.20
------
-* Changed the new note icon [https://github.com/Automattic/simplenote-android/pull/1436]
-
 2.19
 -----
 * Added support to import notes [https://github.com/Automattic/simplenote-android/pull/1333]
 * Fixed crash when searching notes in tablets [https://github.com/Automattic/simplenote-android/pull/1425]
 * Added In App Account Deletion [https://github.com/Automattic/simplenote-android/pull/1416]
 * Fixed email verification screen appears with verified accounts [https://github.com/Automattic/simplenote-android/pull/1423]
+* Changed the new note icon [https://github.com/Automattic/simplenote-android/pull/1436]
 
 2.18
 -----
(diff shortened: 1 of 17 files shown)


SylvesterWilmott:
Looks good! Thanks, @danilo04

SylvesterWilmott (approved):

ParaskP7 (approved):
👋 @danilo04 !

PS: The only minor (🔍) comment I can add to it is that I would consider splitting the `Refactor` from the `Feature` commits as much as possible. For example this 1201df959c40d3588e1da7d40e12dd67134ee863 commit of yours, on some file changes, like the `note_list_widget_dark_black.xml`, contain both changes, which might make it difficult for a reviewer to verify correctness. As such, I would suggest to either don't do the `Refactor`, or do it in a separate commit, before committing the actual `Feature` related change. Wdyt?

danilo04:
Howdy @ParaskP7. Thanks for the feedback. I think that was something made by Android studio. Since it didn't change anything for the fragment layout, I assumed it was the same for the other layouts. I will need to double-check next time on simple changes to see if it didn't change anything.

ParaskP7:
👋 @danilo04 !

Anytime! 🚀 

> Howdy @ParaskP7. Thanks for the feedback. I think that was something made by Android studio. Since it didn't change anything for the fragment layout, I assumed it was the same for the other layouts. I will need to double-check next time on simple changes to see if it didn't change anything.

Yes, AS will do that for you if you have the below settings enabled, see `Reformat code` or/and `Rearrange code`, this will do that for you automagically on this specific file/section that you just modified:

image

So, what I usually do is:
1. Being brave and commit anyway, with those settings on, But then, I double check my this commit from the `Log` to make sure that it actually reflects what I wanted, without any additional changes. If there are changes that I don't want, I undo the commit.
2. Being careful and run a reformat/rearrange manually to see if that will cause any automatic reformatting/re-arrangement. If so, then I have two choices:
    - Either keep those automatic changes but add a `reformat` commit on them separately, or
    - Undo those automatic changes and commit only my `feature` change by disabling the reformat/rearrange checks for this commit only.

Does the above process help you a bit?

danilo04:
Nice @ParaskP7! Let me see how I incorporate this to my workflow.

PR #1443, opened 2021-08-11
Rewrite tag-related activities and fragments in Kotlin

### LINKED ISSUE #1373: Rewrite Tag related screens ( AddTagActivity, TagDialogFragment, TagActivity) in Kotlin
Parent: #1361

danilo04:
Fixes #1373

### Fix
Rewrite tag-related activities and fragments in Kotlin. This includes `TagsActivity`, `AddTagActivity` and `TagDialogFragment`. 

### Test

**Search**
1. Go to left panel menu
2. Tap on `Edit Tags`
3. Search for tags and verify the tags shown are the right results
4. Search for tags that don't exists, the text `No Tags Found` should be shown

**Add Tag**
1. Tap in the plus button (bottom right corner)  to add new tags
2. Add a new tag that is not in the list of tags and save
3. Tap in the plus button (bottom right corner) and try to add a tag that already exists
4. An error message with tag that already exists should be shown
5. Add an space 
6. An error message with tags cannot contain spaces should be shown

**Edit Tag**
1. Tap in one of the tags in the list of tags
2. Add a new tag that is not in the list of tags and save
3. Tap into another tag and try to add a tag that already exists
4. A Tag conflict message shown be shown
5. Tag on merge to merge the tag

**Delete Tag**
1. Tap on the trash icon of one of the tags in the list
2. If the tag does not have a note associated, the tag will be deleted immediately
3. If the tag has at least one note associated, a dialog asking if the user is sure to delete the tag should be shown
4. Tap `Yes` and the tag should be deleted

### Review

### Release


Simplenote/src/androidTest/java/com/automattic/simplenote/integration/TagsActivityTest.kt
```diff
@@ -30,6 +31,15 @@ import org.junit.runner.RunWith
 class TagsActivityTest : BaseUITest() {
```
ParaskP7, line 31: Praise (❤️): I love the 2 new empty tags related tests that you added based on the bug that you found, kudos!

Simplenote/src/main/java/com/automattic/simplenote/utils/BaseExt.kt
```diff
@@ -0,0 +1,21 @@
...
+import androidx.core.content.ContextCompat
+
+fun AppCompatActivity.toast(@StringRes resId: Int, length: Int = Toast.LENGTH_SHORT) {
+    Toast.makeText(this, getString(resId), length).show()
+}
+
+fun AppCompatActivity.getColorStr(@ColorRes color: Int): String {
+    val hintColor = ContextCompat.getColor(this, color)
+    val hintHexColor = hintColor.toHexString().replace("ff", "")
+    return "#$hintHexColor"
```
ParaskP7, line 16: Question (❓): Is this `#` character a new addition, it didn't exist before, right?

Simplenote/src/main/java/com/automattic/simplenote/utils/BaseExt.kt
```diff
@@ -0,0 +1,21 @@
...
+}
+
+fun Int.toHexString(): String {
+    return Integer.toHexString(this)
+}
```
ParaskP7: Minor (🔍): Consider adding a new line at the end of this file. Actually, GitHub also warns you about it, nice! 😄 

PS: Same apply to all other files. Actually, consider adding that setting to your AS so that it is done for you automatically from now on.

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.kt
```diff
@@ -0,0 +1,130 @@
...
+    private val viewModel: AddTagViewModel by viewModels()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        ThemeUtils.setTheme(this)
+        super.onCreate(savedInstanceState)
+        val binding: ActivityTagAddBinding = ActivityTagAddBinding.inflate(layoutInflater)
+
+        binding.setObservers()
+        binding.setupLayout()
+        binding.setupViews()
```
ParaskP7, line 34: Praise (❤️): I love what you did with the binding extension functions and using this pattern to build the UI, we are doing the same on `WordPress` so it is great to see it happening on `Simplenote` as well, kudos!

PS: Same applies to all the other places you did the same!

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        viewModel.start()
+
+        setContentView(binding.root)
+    }
+
+    private fun ActivityTagAddBinding.setupViews() {
+        title.text = getString(R.string.add_tag)
+        tagInput.doAfterTextChanged { s -> viewModel.updateUiState(s.toString()) }
```
ParaskP7, line 43: Praise (❤️): I love what you did with this KTX extension function and for managing to remove the `TextWatcher`. The whole class is so much cleaner now, kudos!

PS: Same applies to all the other places you did the same!

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.kt
```diff
@@ -0,0 +1,151 @@
...
+     */
+    private fun showDialogErrorConflict(canonical: String, tagOld: String) {
+        binding.message.text = getString(R.string.dialog_tag_conflict_message, canonical, tagOld, canonical)
+        binding.message.visibility = View.VISIBLE
+        binding.inputTagName.visibility = View.GONE
+
+        dialogEditTag.setTitle(R.string.dialog_tag_conflict_title)
+
+        val positiveButton = dialogEditTag.getButton(DialogInterface.BUTTON_POSITIVE)
+        positiveButton.text = getString(R.string.dialog_tag_conflict_button_positive)
```
ParaskP7, line 113: Question (❓): What if we were explicit with the `positiveButton` visibility as well, just like we do for the `negativeButton` and `neutralButton`?

PS: Same apply for the `showDialogRenameTag()` function below.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.kt
```diff
@@ -0,0 +1,151 @@
...
+        showDialogRenameTag()
+    }
+
+    private fun startUiState() {
+        viewModel.start(tag)
+        binding.inputTagName.editText?.setText(tag.name)
+    }
+
+    /**
+     * Change the dialog layout to show a conflict message in which the user can merge two tags or cancel the merge
```
ParaskP7: Minor (🔍): Consider ending the KDoc sentence with a dot (`.`).

PS: Same apply for the showDialogRenameTag() function below.

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.kt
```diff
@@ -0,0 +1,246 @@
...
+class TagsActivity : ThemedAppCompatActivity() {
+    private val viewModel: TagsViewModel by viewModels()
+
+    private var _binding: ActivityTagsBinding? = null
+    private val binding get() = _binding!!
+
+    public override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        _binding = ActivityTagsBinding.inflate(layoutInflater)
```
ParaskP7: Warning (⚠️): For activities you don't need to do the binding backing field trick so that to then make the binding null during `onDestroy()`. This is only required on fragment and in case you need that field since you need access to it from outside the `onCreate(...)`, just like you did for `TagDialogFragment`.

Actually, for `AddTagActivity` you did exactly what I am suggesting, so I am not sure why you didn't do the same for this activity as well.

As such, this can become `val binding: ActivityTagsBinding = ActivityTagsBinding.inflate(layoutInflater)` and the binding fields can be removed. Doing so will propagate further refactoring down the line, mainly making some more functions an extension function of the binding.

ParaskP7 (changes requested):
👋 @danilo04 !

I have reviewed and tested this PR as per the instructions, everything works as expected (ish), amazing job! 🌟 🌟 🌟 

The only problem I found during testing was during the `Edit Tag` testing, where a tag, that is not associated with any note, wouldn't update itself. Or two conflicting tags, that are not associated with any note, wouldn't merge as well. Can you double check that? Maybe this is a known bug that is already existing out in the wild as well. 🤷 

I have left a warning (⚠️), a couple of questions (❓), some minor (🔍) comments for you to consider. I am **NOT** going to approve this PR atm, since I would like for us to figure out what is wrong with the `Edit Tag` testing and maybe apply a fix on the my warning comment.

In addition, I would like to offer a couple suggestions (💡) for you to consider for when converting a Java file to Kotlin:
- Preparing the Java file, prior hitting the `Convert to Kotlin` button, might help you create a much better Kotlin file. For example, adding `@Nullable` or `NonNull` annotation wherever possible can end up helping you enormously. Try to find out what kind of refactoring or preparation you can do before hitting the `Convert to Kotlin` button.
- When converting to `ViewBinding` and removing all fields at the same time can end up being more complicated than expected. When doing that, do a commit-by-commit field conversion, with the first commit being the binding setup. For example, if that was done on `TagsActivity` you might have saved yourself the trouble of forgetting the `list.setEmptyView(empty.root)` since it would have been more obvious that this is missing.

PS: Last but not least, consider renaming the `AddActivityViewModelTest` test class to `AddTagViewModelTest`. The file name and class name differ and as such the main class with the test class differ, which makes it harder for the AS to associate the main to the test class.

Simplenote/src/main/java/com/automattic/simplenote/utils/BaseExt.kt
danilo04: Yes, I added it because if you get a color using `getString(...)`, it will include `#`, but since we are getting the color using `getColor(...)`, it does not include it.

Simplenote/src/main/java/com/automattic/simplenote/AddTagActivity.kt
danilo04: Yes, I saw that in some files we discussed before and I thought that is a great way to reduce boilerplate and reduce the line length.

Simplenote/src/main/java/com/automattic/simplenote/utils/BaseExt.kt
ParaskP7: Ah, nice, thanks for the clarification Danilo! 🙏

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.kt
danilo04: Good catch! Fixed in b41944abc5f2a0b8a0ac1c4fca8e9c567e8010d9.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.kt
danilo04: For the moment is not needed since the positive button is always visible, but to avoid any bugs in future changes, I agree that making it explicit would make this code less error-prone. Fixed in 045bddff1a9612962d432649dced6d1fddc31c4e.

Simplenote/src/main/java/com/automattic/simplenote/utils/BaseExt.kt
danilo04: 🤯 , did not know about this. Thanks! Fixed in 55059c868d009e4e830bd484b0b2ae4d2e091c7a.

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.kt
danilo04: Good point! Fixed in d3c92932c209f99075c7bc3ead2cbfcec4a0bcf6.

danilo04:
Howdy @ParaskP7 👋 . Thanks for the thorough review. I have added some changes answering your comments.

> The only problem I found during testing was during the Edit Tag testing, where a tag, that is not associated with any note, wouldn't update itself. Or two conflicting tags, that are not associated with any note, wouldn't merge as well. Can you double check that? 

This is a bug that already exists in the application. I am keeping it for now since there is a community PR that I think touches this bug. 

> Preparing the Java file, prior hitting the Convert to Kotlin button, might help you create a much better Kotlin file. For example, adding @Nullable or NonNull annotation wherever possible can end up helping you enormously. 

Good point! I will use this for future conversions.

> When converting to ViewBinding and removing all fields at the same time can end up being more complicated than expected. When doing that, do a commit-by-commit field conversion, with the first commit being the binding setup. 

Yes, I agree. Although I was using the tests during the refactoring, some refactors were merged together and that complicated the whole process (I think you could see that in the commit history). I need to be more careful for sure. 

> Last but not least, consider renaming the AddActivityViewModelTest test class to AddTagViewModelTest
Fixed in 4c72de558edac2616f48d7b66ceaa27d50e8a963.

Simplenote/src/main/java/com/automattic/simplenote/TagDialogFragment.kt
ParaskP7: You are a rockstar, thank you! 🙏

Simplenote/src/main/java/com/automattic/simplenote/TagsActivity.kt
ParaskP7: Amazing, many thanks Danilo! 🕺 💃

ParaskP7:
> Howdy @ParaskP7 👋 . Thanks for the thorough review. I have added some changes answering your comments.

👋 @danilo04 !

Thanks so much for applying all these changes! 🙏 

> This is a bug that already exists in the application. I am keeping it for now since there is a community PR that I think touches this bug.

Great, thanks for the clarification on that! 🙏 

> Good point! I will use this for future conversions.

🥇 🙏 

> Yes, I agree. Although I was using the tests during the refactoring, some refactors were merged together and that complicated the whole process (I think you could see that in the commit history). I need to be more careful for sure.

🥇 🙏 

> Fixed in 4c72de5.

👍

ParaskP7 (approved):
👋 @danilo04 !

I did another round on your latest changes and it LGTM, thanks so much for working on this PR, it is wild to see how much the tags screen is improved! ❤️ 

PS: Feel free to merge this PR when you are ready.

PR #1450, opened 2021-08-18
Handle back press button when NoteEditorActivity is the Task Root

### LINKED ISSUE #1429: Widget: Back button doesn't return to Simplenote but to home screen
When using an Android widget, the Back button inside Simplenote doesn't return to the list with all notes if the app is opened for the first time from the widget. Instead, tapping the back button will redirect users to the home screen.

### Expected
Tapping the back button should return to the list of notes in Simplenote

### Observed
- On existing notes, the Back button returns the user to the Home Screen. If you want to stay on Simplenote and edit another note, for instance, there is no way
- If you tap the + (new note) button instead to create a new note, then tap back, you're redirected to the All Notes screen in Simplenote
- You can only return to the All Notes screen if Simplenote was already opened

### Reproduced
1. Make sure that Simplenote is closed
2. Go to the home screen and add a Simplenote Widget.
2. Tap on a note from the widget list to open Simplenote
3. Tap the Back button in Simplenote

#### Reference 4184686-zd-woothemes

danilo04:
Fixes #1429

The editor can be the task root when it comes from an action on a widget. In these cases, instead of going to the home screen, the notes activity is started. The toolbar's back button also follows this behavior.

### Fix

Tapping the back button after entering a note through a widget, should return to the list of notes in Simplenote.

**Without Passcode**

https://user-images.githubusercontent.com/195721/129981459-b9490a6d-8729-4398-991e-5e37ab0d37e6.mp4

**With Passcode**

https://user-images.githubusercontent.com/195721/129981490-667f720a-e145-4f2a-a29a-60e067fda05b.mp4

### Test

1. Add a widget to the home screen with the list of notes
2. Close the app permanently
3. Go to the widget and tap on one of the notes in the list
4. Tap on the back button (top of the toolbar and also the Android back button)
5. The list of notes should be shown

### Review

### Release
`RELEASE-NOTES.txt` was updated in b1a42e7f1723214b7d84d2fa0b4afc1028bbd7dc with:
> Fixed back button return to home screen when a note is accessed from a widget

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,7 @@
 2.20
 -----
-* Removed sort bar from notes list but still accessible from settings [1438](https://github.com/Automattic/simplenote-android/pull/1438)
-* Fixed back button return to home screen when a note is accessed from a widget [1450](https://github.com/Automattic/simplenote-android/pull/1450)
+* Removed sort bar from notes list but still accessible from settings [#1438](https://github.com/Automattic/simplenote-android/pull/1438)
+* Fixed back button return to home screen when a note is accessed from a widget [#1450](https://github.com/Automattic/simplenote-android/pull/1450)
 
 2.19
 -----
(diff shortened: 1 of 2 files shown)


ParaskP7 (approved):
👋 @danilo04 !

PS: Feel free to merge this PR yourself when you are ready.

PR #1454, opened 2021-08-20
Move scroll setup from view tree updates after the note content refreshes

### LINKED ISSUE #1338: [Search] Scroll position of a note with matching keyword in not persistent
This comes from Google Play [review](https://link.reviewbot.io/review/hWBC6R_QD):

> The SEARCH function was the main advantage of this app.When we searched for a word, the result was positioned on the 1st occurrence of the word and proposed the other occurrences. No other competing note app offers this. This was decisive in my choice of app. Since the last update (Feb 2021), if there is only 1 occurrence, the cursor is placed at the end of the note.

After some testing it definitely looks like there is an inconsistency in this place. However, the scroll position can be both correct and 'bottom of the note' for both cases of multiple or single match.

### Expected
If a note contains a search match, tapping a note should show the match. Or, at least, be opened with a consistent scroll position.

### Observed
The scroll position can be both correct and 'bottom of the note' for both cases of multiple or single match.

### Reproduced
1. Have a note that contains some unique keyword (only once, only in one note)
2. Start search against this keyword
3. Tap the note ~> you might firstly see the keyword shown on screen
4. Go back to note list and tap the note again ~> you might see the note scrolled to bottom.

https://user-images.githubusercontent.com/73365754/110334627-a67b9c80-802b-11eb-9ee6-c25d524e220f.mp4

Device Make: Samsung
Device Model: A31
Android Version: 10
Simplenote Version: 2.17-rc-1

danilo04:
Fixes #1338

The initial scroll position is set based on the last scroll position the user was or on the first keyword found if entering from search results. To set the scroll position, the app used a `ViewTreeObserver` with and set up a `OnGlobalFocusChangeListener` callback listener. The problem with this approach is that this callback gets called several times even after the text has been rendered. Thus, the scroll position for a search was overridden by the scroll position saved in preferences.

### Fix
We now change the scroll setup to after the note contents are refreshed. This happens when the note is loaded and set in the EditText. We post a task on the root view to be sure that the scroll setup is done after the text has been rendered. Also, when there is a configuration change, the note is not refreshed, but the view is re-rendered and the scroll position is lost. We setup the scroll as well when there is a configuration change.

### Test
**Note:** have a long note where the same term appears in different parts of the note

1. Have a note that contains some unique keyword (only once, only in one note)
2. Start search against this keyword
3. Tap the note ~> you might firstly see the keyword shown on screen
4. Go back to note list and tap the note again.
5. Tap to the next keyword
6. Rotate phone

Video from issue:

https://user-images.githubusercontent.com/73365754/110334627-a67b9c80-802b-11eb-9ee6-c25d524e220f.mp4

Fixed:

### Review

### Release

`RELEASE-NOTES.txt` was updated in af046a10e74c078bc37f60700d2e2ece5c4e8fff with:
> Fixed scroll position of a note with matching keyword in not persistent


planarvoid (approved):
The solution works and it scrolls to the target word correctly 👍 

Generally I think this could be really simplified but I don't see how to do it without big changes to the codebase so I'm approving this PR.

PR #1459, opened 2021-08-24
Modernize Tag Related Screens and Surrounding Code

danilo04:
Fixes #1361

### Fix

This PR brings the feature branch `feature/modernize_tags` to `develop`. This branch has the following:

- Added support to write Kotlin in the app
- UI tests for Tag related screens
- Added Jetpack components, specially ViewModels
- Rewrite Tag related screens in Kotlin and re-architecture them using MVVM
- Added Dagger/Hilt for Dependency Injection

The changes were reviewed in the following **PRs**:

#### Tests
- #1376 
- #1380 
- #1384 
- #1391 

#### Restructure Tag related screens to using viewModels and MVVM
- #1393 
- #1396 
- #1397 
- #1414 

#### Other tasks

- #1426 
- #1443 

### Test

The whole application but mostly the tag related screens. This include:

1. Open the left panel by taping on the menu (top left corner)
2. Tap on `Edit Tags`
3. Add tags
4. Edit tags
5. Merge tags. This is done by editing a tag and using the name of another tag
6. Delete tags

### Review

One developer is required to review these changes.

### Release

These changes do not require release notes since they don't include any behavior changes.

PR #1481, opened 2021-09-08
Refactor Tag Management in NoteEditorFragment

### LINKED ISSUE #1478: Refactor Tag Management in NoteEditorFragment and add new behavior
Parent #1473.

Refactor Tag Management in NoteEditorFragment and add new behavior. When a tag is added, we have to validate whether it is a collaborator. If it is a collaborator, it should be guided to the collaborator UI. Also, the list of tags should filter collaborators.

### LINKED ISSUE #1480: Refactor tags views to filter out collaborators. This includes: `TagsActivity` and the list of tags shown on the left panel menu.
Parent #1473.

Refactor tags views to filter out collaborators. This includes: `TagsActivity` and the list of tags shown on the left panel menu.

danilo04:
Fixes #1478.

### Fix

This PR refactors tag management in the `NoteEditorFragment`. Since `NoteEditorFragment` is one of the more complex classes in the project, we will refactor it at different stages to use the new architecture added in #1459. In this project, we are going to extract the tag management logic into a `ViewModel` to leverage `Repositories` and `Use Cases`. 

The refactor also adds new behavior to support the collaboration UI described in #1480 (see Details section). The concrete change is that we are not going to show emails (collaborators) as tags. They will be shown in their own UI. This behavior should be implemented in all screens that show a list of tags. 

Thus, this PR adds the use cases `GetTagsUseCase` and `ValidateTagUseCase` which will be used as the single source of truth to get tags (and filter collaborators) and validate a tag. Both use cases will be used in several places in the app.

Another important change in this PR is that the method `saveNote()` in `NoteEditorFragment` does not update tags. It used the list of tags on the UI (`Chip` views) to save the tags, but since that list does not include collaborators (emails), it was deleting them when the note was closed. The tags are updated correctly in all the other actions we included. 

| Before      | After |
| ----------- | ----------- |
|       |        |

### Test

**Note:** important to test in tablets also since the parent activity changes when in landscape mode in tablets. Use <= 2.21 to test whether the emails are added as tags.

1. Go to a note
2. Scroll down to the tags field
3. Add a tag and tap on `Done` button in the keyboard. For example `tag1`
4. Add an email. A `Toast` showing `<email> added as a collaborator.` should be shown
5. Delete a tag
6. Write to add a tag but do not tap on the button `Done`. For example `tag2`
7. Tap on the note editor. The tag should be added with whatever the tag field had
8.  Write to add a tag but do not tap on the button `Done`. For example `tag3`
9. Tab on the back button on top-left corner
10. Go to the same note again
11. The tag `tag3` should be in the list of tags

### Review

### Release


SylvesterWilmott:
Looks good!

I noticed that email tags are also displayed in the tags list in the navigation drawer. Since this change will be paired with a dedicated UI for collaboration, managing collaborators will occur in the dedicated UI. This means we should not show email tags in the tags list.

danilo04:
@SylvesterWilmott Yes, that will be part of the next PR (collaboration UI) or maybe a separate PR. I have added the infrastructure (`GetTagsUseCase`) to reuse it and get the correct tags both in the Tags screen and the navigation drawer.

SylvesterWilmott:
Thanks for clarifying @danilo04 

In that case, it LGTM

Simplenote/src/main/java/com/automattic/simplenote/models/Note.java
```diff
@@ -342,13 +342,30 @@ public boolean hasTag(Tag tag) {
         return tagList;
     }
 
+    public void addTag(String tagName) {
```
ParaskP7, line 345: Question (❓): Is adding a test a possibility for this method?

PS: The same applies to `removeTag(...)` further below.

Simplenote/src/main/java/com/automattic/simplenote/usecases/GetTagsUseCase.kt
```diff
@@ -0,0 +1,24 @@
+package com.automattic.simplenote.usecases
+
+import com.automattic.simplenote.models.Note
+import com.automattic.simplenote.models.TagItem
+import com.automattic.simplenote.repositories.CollaboratorsRepository
+import com.automattic.simplenote.repositories.TagsRepository
+import javax.inject.Inject
+
+/**
+ * Since the collaborators are saved as tags, we have to filter them before returning the tags
```
ParaskP7: Minor (🔍): Consider avoiding active voice and prefer passive voice instead. For example, this KDoc sentence can become: `Since the collaborators are saved as tags, those need to be filtered before returning the tags.`

PS: Don't forget the period at the end... OCD me! 😅

Simplenote/src/main/java/com/automattic/simplenote/usecases/GetTagsUseCase.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+class GetTagsUseCase @Inject constructor(
+    private val tagsRepository: TagsRepository,
+    private val collaboratorsRepository: CollaboratorsRepository) {
```
ParaskP7: Question (❓): I saw you corrected this format, why did you do that? To my understanding, the correct format for this syntax is indeed the below:
```
class GetTagsUseCase @Inject constructor(
    private val tagsRepository: TagsRepository,
    private val collaboratorsRepository: CollaboratorsRepository
) {
```

Simplenote/src/main/java/com/automattic/simplenote/usecases/ValidateTagUseCase.kt
```diff
@@ -0,0 +1,44 @@
+package com.automattic.simplenote.usecases
+
+import com.automattic.simplenote.repositories.CollaboratorsRepository
+import com.automattic.simplenote.repositories.TagsRepository
+import javax.inject.Inject
+
+
+sealed class TagValidationResult {
```
ParaskP7: Suggestion (💡): Consider making this sealed class a part of `ValidateTagUseCase` to make it explicit that this should be used as part of that use case.

Simplenote/src/main/java/com/automattic/simplenote/usecases/ValidateTagUseCase.kt
```diff
@@ -0,0 +1,44 @@
...
+    object TagTooLong : TagValidationResult()
+    object TagIsCollaborator : TagValidationResult()
+    object TagExists : TagValidationResult()
+}
+
+class ValidateTagUseCase @Inject constructor(
+    private val tagsRepository: TagsRepository,
+    private val collaboratorsRepository: CollaboratorsRepository) {
+
+    fun isTagValid(tagName: String): TagValidationResult {
```
ParaskP7: Suggestion (💡): Consider a more Kotlin-ish, kind-of more readable, alternative here:
```
    fun isTagValid(tagName: String) = when {
        tagName.isEmpty() -> TagValidationResult.TagEmpty
        tagName.contains(EMPTY_STRING) -> TagValidationResult.TagWithSpaces
        collaboratorsRepository.isValidCollaborator(tagName) -> TagValidationResult.TagIsCollaborator
        !tagsRepository.isTagValid(tagName) -> TagValidationResult.TagTooLong
        !tagsRepository.isTagMissing(tagName) -> TagValidationResult.TagExists
        else -> TagValidationResult.TagValid
    }
```

PS: Note the `EMPTY_STRING` constant that I extracted instead of using the hardcoded `" "`. I also usually suggest a utils with such strings so that they are being later-on reused across the app.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/NoteEditorViewModel.kt
```diff
@@ -0,0 +1,75 @@
...
+            Stat.EDITOR_TAG_REMOVED,
+            AnalyticsTracker.CATEGORY_NOTE,
+            "tag_removed_from_note"
+        )
+    }
+
+    data class UiState(val tags: List<String>)
+}
+
+sealed class NoteEditorEvent {
```
ParaskP7: Suggestion (💡): Consider making this sealed class a part of `NoteEditorViewModel` to make it explicit that this should be used as part of that use case. Just like what you did for the `UiState` data class.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -551,6 +560,32 @@ public void onClick(View view) {
         return mRootView;
     }
 
+    private void setObservers() {
+        viewModel.getUiState().observe(this, new Observer<NoteEditorViewModel.UiState>() {
```
ParaskP7: Suggestion (💡): Consider refactor the anonymous class to lambdas on this whole fragment.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -1125,22 +1160,13 @@ private int newCursorLocation(String newText, String oldText, int cursorLocation
...
     @Override
     public void onTagAdded(String tag) {
-        if (mNote == null || !isAdded()) {
+        // When a tag is added, it triggers an event that sends an empty tag. For those cases or if the note is null,
+        // we avoid updating the UI and the note
+        if (mNote == null || !isAdded() || TextUtils.isEmpty(tag)) {
             return;
         }
 
-        if (mNote.getTagString() != null && tag.length() > mNote.getTagString().length()) {
```
ParaskP7, line 1132: Question (❓): Where is this logic now?

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -1125,22 +1160,13 @@ private int newCursorLocation(String newText, String oldText, int cursorLocation
...
-            AnalyticsTracker.track(
-                EDITOR_TAG_ADDED,
-                CATEGORY_NOTE,
-                "tag_added_to_note"
-            );
-        }
-
-        mNote.setTagString(mNote.getTagString() + String.valueOf(SPACE) + tag);
```
ParaskP7, line 1140: Question (❓): Where is this logic now?

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -1944,27 +1953,15 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
             chip.setOnCloseIconClickListener(new View.OnClickListener() {
                 @Override
                 public void onClick(View view) {
-                    mTagChips.removeView(view);
```
ParaskP7, line 1947: Question (❓): Where is this logic now?

PS: I saw a commit where you described why that is not needed, but I just want to make sure I am not missing anything. 🙏

Simplenote/src/test/java/com/automattic/simplenote/repositories/CollaboratorsRepositoryTest.kt
```diff
@@ -0,0 +1,38 @@
+package com.automattic.simplenote.repositories
+
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class CollaboratorsRepositoryTest {
```
ParaskP7: Minor (🔍): Consider renaming that to `SimperiumCollaboratorsRepositoryTest` since the test suite is applicable to the implementing class and not its interface.

Simplenote/src/test/java/com/automattic/simplenote/repositories/CollaboratorsRepositoryTest.kt
```diff
@@ -0,0 +1,38 @@
...
+    fun validEmailsShouldBeValidCollaborator() {
+        assertTrue(collaboratorsRepository.isValidCollaborator("name1@test.com"))
+        // Email with + sign
+        assertTrue(collaboratorsRepository.isValidCollaborator("name12.lastname+503@test.com"))
+        // IP address
+        assertTrue(collaboratorsRepository.isValidCollaborator("name12_lastname+503@192.168.43.54"))
+        assertTrue(collaboratorsRepository.isValidCollaborator("name12-lastname+503@192.168.43.54"))
+        // Email top level domain
+        assertTrue(collaboratorsRepository.isValidCollaborator("name@example.co.jp"))
+
```
ParaskP7: Minor (🔍): Consider removing the empty line from here.

Simplenote/src/test/java/com/automattic/simplenote/usecases/GetTagsUseCaseTest.kt
```diff
@@ -0,0 +1,81 @@
...
+
+@ExperimentalCoroutinesApi
+class GetTagsUseCaseTest {
+    @get:Rule val rule = InstantTaskExecutorRule()
+
+    private val tagsRepository: TagsRepository = mock(TagsRepository::class.java)
+    private val mockBucket: Bucket<*> = mock(Bucket::class.java)
+    private lateinit var collaboratorsRepository: CollaboratorsRepository
+
+    private lateinit var getTagsUseCase: GetTagsUseCase
```
ParaskP7: Suggestion (💡): Where writing tests, consider using `lateinit` and the `@Before` block as little as possible, especially with Kotlin. Converting to the below might be a better alternative:
```
    private val collaboratorsRepository = SimperiumCollaboratorsRepository()

    private val getTagsUseCase = GetTagsUseCase(tagsRepository, collaboratorsRepository)

...

    @Before
    fun setup() {
        // Set mock bucket to avoid NPE
        tagItems.forEach { it.tag.bucket = mockBucket }
    }
```

Note, with this alternative all fields are now `val` and this can make the test more robust.

PS: Same applies to `ValidateTagUseCaseTest` test suite:
```
    private val validateTagUseCase = ValidateTagUseCase(tagsRepository, SimperiumCollaboratorsRepository())
```
PSS: Same applies to `NoteEditorViewModelTest` test suite:
```
    private val collaboratorsRepository = SimperiumCollaboratorsRepository()
    private val getTagsUseCase = GetTagsUseCase(tagsRepository, collaboratorsRepository)
    private val validateTagUseCase = ValidateTagUseCase(tagsRepository, collaboratorsRepository)
    private val viewModel = NoteEditorViewModel(getTagsUseCase, validateTagUseCase)

    private val note = Note("key1").also {
        it.content = "Hello World"
        it.tags = listOf("tag1", "tag2", "name@test.com")
        it.bucket = mockBucket
    }

    @Before
    fun setup() {
        whenever(tagsRepository.isTagValid(any())).thenReturn(true)
        whenever(tagsRepository.isTagMissing(any())).thenReturn(true)
    }
```

Simplenote/src/test/java/com/automattic/simplenote/usecases/GetTagsUseCaseTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        // Set mock bucket to avoid NPE
+        tagItems.forEach { it.tag.bucket = mockBucket }
+
+        collaboratorsRepository = SimperiumCollaboratorsRepository()
+        getTagsUseCase = GetTagsUseCase(tagsRepository, collaboratorsRepository)
+    }
+
+    @Test
+    fun allTagsShouldFilterCollaborators() = runBlockingTest {
+        tagsRepository.stub {
```
ParaskP7: Minor (🔍): Consider making this a one-liner.

Simplenote/src/test/java/com/automattic/simplenote/usecases/GetTagsUseCaseTest.kt
```diff
@@ -0,0 +1,81 @@
...
+            onBlocking { allTags() }.doReturn(tagItems)
+        }
+
+        val tagItemsExpected = listOf(
```
ParaskP7: Suggestion (💡): What if this was made more explicit and less hardcoded, in terms of removing duplicates as well:
```
        val tagItemsExpected = tagItems.toMutableList()
        tagItemsExpected.removeAt(3) // TagItem(Tag("name@example.co.jp"), 2)
        tagItemsExpected.removeAt(3) // TagItem(Tag("name1@test.com"), 0)
```

Simplenote/src/test/java/com/automattic/simplenote/usecases/GetTagsUseCaseTest.kt
```diff
@@ -0,0 +1,81 @@
...
+            TagItem(Tag("tag1"), 0),
+            TagItem(Tag("tag2"), 2),
+            TagItem(Tag("tag3"), 5),
+            TagItem(Tag("name@test"), 1),
+            TagItem(Tag("あいうえお@example.com"), 1),
+        )
+        tagItemsExpected.forEach { it.tag.bucket = mockBucket }
+
```
ParaskP7: Minor (🔍): Consider removing the extra line here.

Simplenote/src/test/java/com/automattic/simplenote/usecases/ValidateTagUseCaseTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    @Before
+    fun setup() {
+        val collaboratorsRepository = SimperiumCollaboratorsRepository()
+        validateTagUseCase = ValidateTagUseCase(tagsRepository, collaboratorsRepository)
+    }
+
+    @Test
+    fun emptyTagShouldReturnTagEmpty() {
+        val result = validateTagUseCase.isTagValid("")
```
ParaskP7, line 18: Minor (🔍): Consider adding a new line to split the `act` from the `assert`.

PS: This applies to all the tests below as well.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/NoteEditorViewModelTest.kt
```diff
@@ -0,0 +1,87 @@
...
+    @Test
+    fun addTagShouldUpdateUiState() {
+        viewModel.addTag("tag3", note)
+
+        assertEquals(listOf("tag1", "tag2", "tag3"), viewModel.uiState.value?.tags)
+        assertEquals(listOf("tag1", "tag2", "name@test.com", "tag3"), note.tags)
+    }
+
+    @Test
+    fun addCollaboratorShouldNotUpdateUiStateAndNote() {
```
ParaskP7: Suggestion (💡): Consider splitting this test into two. Imagine that you actually added a commit which altered this test at some point. This could have been a hint that the test shouldn't have been modified and instead a new test should have been added. This is the single test assertion principle that I personally try to abide by.

PS: Else you might want to add the `AndShouldSendTagAsCollaborator` to the test function name to make sure this is being conveyed.
PSS: Same applies to the `addInvalidTagShouldNotUpdateUiStateAndNote()` test below.

ParaskP7 (commented):
👋 @danilo04 !

I have reviewed and tested this PR as per the instructions, everything works as expected, great job! 🌟 🌟 

I have left some questions (❓), few suggestions (💡) and a couple of minor (🔍) comments for you to consider. I am **NOT** going to approve this PR anyway, since I would like to get a response to my questions first, but nothing is blocking tbh, thus I am not requesting for changes, just commenting for now.

1. Praise (❤️): You did such a good job with using the new architecture to kiss start this new feature, really great job there!
2. Praise (❤️): I love how you are testing everything, keep it up! Maybe we should be starting to adding small Espresso tests on the `NoteEditorFragment` at some point, or at least its setup, just to have this layer tested as well. This way, while developing new features on this screen, we could make sure that nothing bad/unwanted gets slipped into production.
3. Praise (❤️): Your commits and commit messages are excellent and they really helped me during code review, kudos!

Simplenote/src/main/java/com/automattic/simplenote/usecases/GetTagsUseCase.kt
danilo04: Yes, that is correct. Fixed in 9083c3548d99aa7aaf6f32bba5bbd0ee1abedcec.

Simplenote/src/main/java/com/automattic/simplenote/models/Note.java
danilo04: I moved `NoteTest` to unit tests folder in 5841aea39303eb607e2f1afec5496f4521cdb8f7, added tests for `addTag` in e1af60d07f05024ab2bb3c58227ce16b14fde223, and tests for `removeTag` in e1af60d07f05024ab2bb3c58227ce16b14fde223.

Simplenote/src/main/java/com/automattic/simplenote/usecases/GetTagsUseCase.kt
danilo04: Fixed in 6af8e566cd5baacf4c77476d66e2449fb527e592.

Simplenote/src/main/java/com/automattic/simplenote/usecases/ValidateTagUseCase.kt
danilo04: Good suggestion. Namespaces in this case make it easier to grasp. Fixed in 3c77f44a90fa27076dc0c59f58c8114363644806.

Simplenote/src/main/java/com/automattic/simplenote/usecases/ValidateTagUseCase.kt
danilo04: This version looks more compact and easier to understand. The only thing that worried me a little bit was that the semantics of `when` assures the execution order is the same as the one defined. From [docs](https://kotlinlang.org/spec/expressions.html#when-expressions):

> When entries are checked and evaluated in their order of appearance. 

Fixed in 91e2a21d0f1d04f071c0dc1a9cab002bb9953136.

Simplenote/src/test/java/com/automattic/simplenote/usecases/GetTagsUseCaseTest.kt
danilo04: Fixed in 17d50ea901fb4c8fe0eae7d559191dac68dadff2.

Simplenote/src/test/java/com/automattic/simplenote/usecases/GetTagsUseCaseTest.kt
danilo04: Fixed in e3698a945dc8cc6c4b73e7bf365505cd310f1b27.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
danilo04: Fixed in 3d782f89995527dfe08a1af016df47c8483036cb.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
danilo04: In `NoteEditorViewModel` as part of `addTag()`.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
danilo04: In `NoteEditorViewModel` as part of `addTag()`.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
danilo04: This call is redundant because after the tag is deleted, all the chips are updated with the new tags:
```java
 private void setChips(List<String> tags) {
        mTagPadding.setVisibility(tags.size() > 0 ? View.VISIBLE : View.GONE);
        mTagChips.setVisibility(tags.size() > 0 ? View.VISIBLE : View.GONE);
        mTagChips.setSingleSelection(true);
        mTagChips.removeAllViews();
```

So the view (chip) was deleted, but all the chips were also deleted later when the new tags arrived.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/NoteEditorViewModel.kt
danilo04: Fixed in e151762b10b2043819d05d4697bf73eaf9f405b1.

Simplenote/src/test/java/com/automattic/simplenote/repositories/CollaboratorsRepositoryTest.kt
danilo04: Fixed in 3e4ec541822f5dc2abd85333b4c54d99c19bc921.

Simplenote/src/test/java/com/automattic/simplenote/repositories/CollaboratorsRepositoryTest.kt
danilo04: Fixed in 6193f9d9fa45c1e27853a2cc88230487cb99d330.

Simplenote/src/test/java/com/automattic/simplenote/usecases/GetTagsUseCaseTest.kt
danilo04: Thank you for the advice on this. So much cleaner. Fixed in d0d0ef96b77bcf3ff5ded82ecdad0ed9e05affc0.

Simplenote/src/test/java/com/automattic/simplenote/usecases/ValidateTagUseCaseTest.kt
danilo04: Fixed in 1d64e357ca5ba95237a80edf14800224be9f5d24.

Simplenote/src/test/java/com/automattic/simplenote/usecases/GetTagsUseCaseTest.kt
danilo04: Fixed in 1d64e357ca5ba95237a80edf14800224be9f5d24.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/NoteEditorViewModelTest.kt
danilo04: > This is the single test assertion principle that I personally try to abide by.

Interesting, I haven't heard about it before. I will take a closer look into it to see how we can write more maintainable tests now that I think of. Fixed in e6f2a578815e0f72fb8e839a5044a9db23370c65.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
danilo04: `addTag` updates `UiState` which updates the list of chips. that is why `updateTagList();` is not called.

danilo04:
Thank you for the review @ParaskP7. I have made the changes and commented on each of your questions and considerations. Let me know if there is something else you see that requires some change.

> You did such a good job with using the new architecture

Thank you! The architecture was shaped with the help of both Vojta and you.

> Maybe we should be starting to adding small Espresso tests on the NoteEditorFragment at some point

I tried, but it is more complicated than other fragments. It is tightly coupled with `NotesActivity` and requires lots of setup before we could start testing it. 

> Your commits and commit messages are excellent 

Thank you!

Simplenote/src/main/java/com/automattic/simplenote/models/Note.java
ParaskP7: This is awesome Danilo, thanks so much for adding those tests! 🥇

Simplenote/src/main/java/com/automattic/simplenote/usecases/ValidateTagUseCase.kt
ParaskP7: This is awesome Danilo, thank you so much for taking it a step further, thinking about and then verifying that the execution order is the same. 🥇 

PS: Also, 🥇 for extracting the `SPACE_SPRING` constant and making it a utils string! ❤️

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/NoteEditorViewModelTest.kt
ParaskP7: Oh well, this is not a principle per se, rather a rule of sorts. Doing a quick Google search will give you the: 

`"One assertion per test" is a wise rule to keep in mind, because it helps you have tests that fail for a specific reason, and drives you to focus on a specific behavior at a time. ... in his great writing “Test Desiderata”, this test is still “specific”, because, if it fails, “the cause of the failure should be obvious”.`

Thank you so much for even considering it! 🙇 

PS: To me, when I write tests, I write them in such a way that all test get to be closed to modification. So, if a logic gets updated, say something is added, I then want to add a new test, else, if say something gets removed, then I want to remove a test, but never modify it by removing or adding a line within an existing test, thus keeping all tests closed to modification, and only allowing deletion when necessary. Does this makes sense?

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
ParaskP7: Thanks for the confirmation Danilo! 🙏

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
ParaskP7: ❓ What I meant, is the `mNote.setTagString(...)` logic where it gets to be concatenated with the `SPACE` etc. 🤔 What am I missing here, I am probably too stupid to understand it completely, so I do apologise for being persistent on that? 🙏

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
ParaskP7: Ah, this is true, thanks so much for clarifying that for me Danilo! 🥇

ParaskP7 (approved):
👋 @danilo04 !

I have reviewed your updates and everything LGTM, such a good job, thank you! 🌟 🌟

PS: I have only one last (pending) question I want to discuss a bit, just in case, but I am not going to hold this PR any longer, thus I am approving it, but letting you do the merge when you see fit.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
danilo04: Howdy 👋  @ParaskP7. No problem at all, I may have introduced a bug, so it is important to discuss it. 

> What I meant, is the `mNote.setTagString(...)` logic where it gets to be concatenated with the SPACE etc. 

The implementation of tag management in `NoteEditorFragment` handle 4 main events and used two sources of truth for tags: 1) the note object `mNote` through `Note.getTagString()`, and 2) the tags in the UI through the method `NotEditorFragment.getNoteTagsString()` (number of `Chip` views on the `ChipGroup`). The 4 main events are the following: 

1. `LOAD`: when the fragment is loaded, the tags are obtained from the note. Use `Note.getTagString()` as source of truth
2. `NOTE UPDATE`: when the note is updated from another device for example. Used `Note.getTagString()` as source of truth
3. `ADD TAG`: when a tag is added (UI event) through the callback `onTagAdded()`. Used `Note.getTagString()` as source of truth and `mNote.setTagString(mNote.getTagString() + String.valueOf(SPACE) + tag)` to update the state
4. `REMOVE TAG`: when a tag is removed (UI event). This event used `NotEditorFragment.getNoteTagsString()` as source of truth. 
5. `TAG FIELD LOSE FOCUS`: when the tag field loses focus. If the field had a valid tag, the tag was added and used `Note.getTagString()` as source of truth. Otherwise, it used the `NotEditorFragment.getNoteTagsString()` as source of truth to set the tags (which was redundant since the tags did not change in this case).

The whole add space and split by space was needed because the method `NotEditorFragment.getNoteTagsString()` returned a `String` of tags separated by space, instead of returning a list of tags. This drove the whole architecture to use `String` as list of tags and add or split/join by `space` in several parts of the code. 

With the refactor, we drive the UI from the lower layers (`ViewModel` -> `UseCase` -> `Repository` -> `Note`). We have only one source of truth which is the `Note` modified by the `UseCase` to filter collaborators out. 

We also reduced the complexity because the fragment just deals with calling the `ViewModel` in each of the 4 events previously mentioned and reacting to change on the `UiState` to update the tags (`Chip` views). Now the 4 events are handled as follows:

1. `LOAD`: call `viewModel.update(mNote)` which updates the `UiState` that is handled by the fragment to update the tags (`Chip` views)
2. `NOTE UPDATE`: call `viewModel.update(mNote)` which updates the `UiState` that is handled by the fragment to update the tags (`Chip` views)
3. `ADD TAG`: call `viewModel.addTag(tag, mNote)` which updates the tags in the note and the `UiState` that is handled by the fragment to update the tags (`Chip` views)
4. `REMOVE TAG`: call `viewModel.removeTag(tag, mNote)` which updates the tags in the note and the `UiState` that is handled by the fragment to update the tags (`Chip` views)
5. `TAG FIELD LOSE FOCUS`: If the field had a valid tag, call `viewModel.addTag(tag, mNote)` which updates the tags in the note and the `UiState` that is handled by the fragment to update the tags (`Chip` views). Otherwise, it doesn't do anything.

I hope this explanation does not bring more confusion. Let me know if something isn't clear or where to expand.

ParaskP7: ❤️ Thank you so much for digging so deep to explain all that to me, much much much appreciate! Everything, seems like they are in order. Thank YOU! 🙇

Issue #1473, opened 2021-09-03
Add a dedicated UI to collaboration


danilo04:
Currently, only iOS and Web have a dedicated UI while Android and macOS uses the tag field to add a collaborator. 

Internal link for desgin specification: pd8BQf-ai-p2
Internal link for project: p2XJRt-3i1-p2

### What
Add a UI to add and remove collaborators from notes. 

### Why
Feature parity with iOS client.

#### Tasks

- [x] #1478
- [x] #1479
- [x] #1480  
- [x] #1496
- [x] #1508

Details:

- Any collaborator may only see collaborators they themselves have added.
- If a collaborator trashes the note:
     * All collaborators added by them including themselves are removed from the note.
     * Restoring the note does not restore collaborators.
     *.Restoring the note does not restore themselves as a collaborator to the owner’s note.
- If the collaborator (User A) removes another collaborator (User B) then:
     * User B still sees the note in their note list.
     * The note is no longer a collaborative note and changes will not be synced with the master note.
- If a note is published, it does not appear as published for collaborators and may be published independently by them.
- Currently, on Android, it is possible to add a collaborator by adding their email as a tag on the note. We should keep this functionality with some small changes to the way in which is behaves.
     * Email tags are “consumed”, once added they disappear from the tags field.
     * A toast verifies that the email was added as a collaborator.
     * Email tags do not appear in the tags list.
- To help differentiate between a normal note and a collaborative note, a collaboration status indicator has been added to the designs.

Collaboration flow:

![blueprint](https://user-images.githubusercontent.com/195721/132068110-55a77994-3d0b-4e66-9bfe-8f18d8531516.png)

PR #1491, opened 2021-09-13
Add collaboration UI

### LINKED ISSUE #1479: Add Collaborators UI. This includes: list of collaborators and add Collaborator dialog.
Parent #1473.

Add Collaborators UI. This includes: list of collaborators and add Collaborator dialog.

danilo04:
Fixes #1479 

### Fix

This PR adds a UI for adding/removing collaborators to a note. Sorry for the size of the PR but I couldn't break more the changes to bring collaboration to Android (they already broken in 4 different issues/PRs https://github.com/Automattic/simplenote-android/issues/1473). 

**Note:** I have stressed tested the UI by adding and removing collaborators as fast as I could and there are some cases in which collaborators are deleted from the list but then added. I presume that this is mostly due to a race condition on the `Simperium` library. I will track the issue in https://github.com/Simperium/simperium-android/issues/238.

### Test

1. Go to a note
2. Tap on the menu on the top right corner (dot ...)
3. Tap on collaborators
4. Tap on Add Collaborator and type anything that is not a valid email. A message invalid collaborator should be shown.
5. Write a valid email address and tap on  `Accept`. Should update the layout to list of collaborators
6. Tap on the cross icon to remove the collaborator. Should update the layout to show the initial message
7. Tap on Add Collaborator and enter an email address from another valid simplenote account
8. The note should be shared to the other account

### Review


Simplenote/src/main/java/com/automattic/simplenote/di/ThreadModule.kt
```diff
@@ -0,0 +1,23 @@
...
+import dagger.hilt.components.SingletonComponent
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import javax.inject.Named
+
+const val IO_THREAD = "IO_THREAD"
+
+@Module
+@InstallIn(SingletonComponent::class)
+abstract class ThreadModule {
```
ParaskP7, line 15: Question (❓): I am not sure about that, but why not just use `Dispatchers.IO` directly to avoid this kind of extra complexity? For example, within `SimperiumTagsRepository`, this is what was done.

PS: This is totally unrelated to this PR, but `SimperiumTagsRepository` is untested, maybe we should create an issue to make sure that to cover this repository with adequate tests as well. Wdty?

Simplenote/src/main/java/com/automattic/simplenote/repositories/CollaboratorsRepository.kt
```diff
@@ -1,8 +1,31 @@
...
+
 interface CollaboratorsRepository {
+
     /**
      * Check whether the [collaborator] is a valid under the logic defines for Simplenote clients.
      */
     fun isValidCollaborator(collaborator: String): Boolean
+
+    /**
+     * Get a list of collaborators for a given [noteId]
```
ParaskP7: Minor (🔍): This is so minor that could disappear by itself 😅 but try to add a period (`.`) at the end of the sentence so that it reads as expected without implying continuation.

PS: The same applies everywhere else within this or other docs with similar issues.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepository.kt
```diff
@@ -14,4 +31,83 @@ class SimperiumCollaboratorsRepository @Inject constructor() : CollaboratorsRepo
...
+        withContext(ioDispatcher) {
+            return@withContext when(val result = getNote(noteId)) {
+                is Either.Left -> result.l
+                is Either.Right -> {
+                    val note = result.r
+                    note.removeTag(collaborator)
+                    CollaboratorsActionResult.CollaboratorsList(filterCollaborators(note))
+                }
+            }
+        }
```
ParaskP7, line 67: Suggestion (💡): The `withContext(ioDispatcher) { ... }` block is necessary because of the `getNote(noteId)`. What if that was pushed down to the `getNote(noteId)` itself so that all three usages of it, or any future usages are free of doing that explicitly. Wdyt?

```
    override suspend fun getCollaborators(
        noteId: String
    ) = when (val result = getNote(noteId)) {
        is Either.Left -> result.l
        is Either.Right -> CollaboratorsActionResult.CollaboratorsList(filterCollaborators(result.r))
    }

    override suspend fun addCollaborator(
        noteId: String,
        collaborator: String
    ) = when (val result = getNote(noteId)) {
        is Either.Left -> result.l
        is Either.Right -> {
            val note = result.r
            note.addTag(collaborator)
            CollaboratorsActionResult.CollaboratorsList(filterCollaborators(note))
        }
    }

    override suspend fun removeCollaborator(
        noteId: String,
        collaborator: String
    ) = when (val result = getNote(noteId)) {
        is Either.Left -> result.l
        is Either.Right -> {
            val note = result.r
            note.removeTag(collaborator)
            CollaboratorsActionResult.CollaboratorsList(filterCollaborators(note))
        }
    }

    ...

    private suspend fun getNote(noteId: String) = withContext(ioDispatcher) {
        return@withContext try {
            val note = notesBucket.get(noteId)
            when (note.isDeleted) {
                true -> Either.Left(CollaboratorsActionResult.NoteInTrash)
                false -> Either.Right(note)
            }
        } catch (e: BucketObjectMissingException) {
            Either.Left(CollaboratorsActionResult.NoteDeleted)
        }
    }

```

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepository.kt
```diff
@@ -14,4 +31,83 @@ class SimperiumCollaboratorsRepository @Inject constructor() : CollaboratorsRepo
...
+
+    override suspend fun collaboratorsChanged(noteId: String): Flow<Boolean> = callbackFlow {
```
ParaskP7, line 70: Warning (⚠️): Consider adding a test for this function, just like it is done for `SimperiumTagsRepository.tagsChanged()`.

Simplenote/src/main/java/com/automattic/simplenote/utils/CollaboratorsAdapter.kt
```diff
@@ -0,0 +1,48 @@
...
+) : ListAdapter<String, CollaboratorsAdapter.CollaboratorViewHolder>(DIFF_CALLBACK) {
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): CollaboratorViewHolder {
+        val binding = CollaboratorRowBinding.inflate(LayoutInflater.from(parent.context), parent, false)
+        return CollaboratorViewHolder(binding, onDeleteClick)
+    }
+
+    override fun onBindViewHolder(holder: CollaboratorViewHolder, position: Int) {
+        val currentCollaborator = getItem(position)
+        holder.bind(currentCollaborator)
```
ParaskP7: Minor (🔍): Consider making this a one-liner: `holder.bind(getItem(position))`

Simplenote/src/main/java/com/automattic/simplenote/utils/CollaboratorsAdapter.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    class CollaboratorViewHolder(
+        private val binding: CollaboratorRowBinding,
+        private val onDeleteClick: (collaborator: String) -> Unit
+    ): RecyclerView.ViewHolder(binding.root) {
+
+        fun bind(collaborator: String) {
+            binding.collaboratorText.text = collaborator
+            binding.collaboratorRemoveButton.setOnClickListener {
```
ParaskP7: Minor (🔍): Consider making this a one-liner: `binding.collaboratorRemoveButton.setOnClickListener { onDeleteClick(collaborator) }`

Simplenote/src/main/java/com/automattic/simplenote/utils/Either.kt
```diff
@@ -0,0 +1,22 @@
+package com.automattic.simplenote.utils
+
+sealed class Either<out L, out R> {
```
ParaskP7, line 3: 1. Praise (❤️): Firstly, kudos for thinking of adding this to the project!
2. Suggestion (💡): Secondly, let's discuss `Either` vs. `Result`. I usually go for `Result` because it is more intuitive than `Either` and as such more readable, especially given function programming is not something mainstream for this project:

```
sealed class Result<out T : Any> {
  data class Success<out T : Any>(val data: T) : Result<T>()
  data class Error(val exception: Exception) : Result<Nothing>()
}
```

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+class CollaboratorsViewModel @Inject constructor(
+    private val collaboratorsRepository: CollaboratorsRepository
+) : ViewModel() {
+    private val _uiState = MutableLiveData<UiState>()
+    val uiState: LiveData<UiState> = _uiState
+
+    private val _event = SingleLiveEvent<Event>()
+    val event: LiveData<Event> = _event
+
+    private lateinit var _noteId: String
```
ParaskP7: Minor (🔍): Consider removing the underscore (`_`) from `noteId`. This underscore is usually preserved for backing fields, like the `_uiState` paired with a `uiState`. Why did you utilize that on `noteId` as well, am I missing something?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+    fun loadCollaborators(noteId: String) {
+        _noteId = noteId
+        viewModelScope.launch {
+            updateUiState(noteId)
+        }
+    }
+
+    private suspend fun updateUiState(noteId: String) {
+        when (val result = collaboratorsRepository.getCollaborators(noteId)) {
+            is CollaboratorsActionResult.CollaboratorsList ->
```
ParaskP7, line 39: 1. Minor (🔍): Consider adding the `is` on all cases.
2. Minor (🔍): Consider adding brackets when splitting `if/else` into multiple lines. In this case I would actually go with another `when` clause, like:
```
                _uiState.value = when (result.collaborators.isEmpty()) {
                    true -> UiState.EmptyCollaborators
                    false -> UiState.CollaboratorsList(result.collaborators)
                }
```

PS: Same applies to the `removeCollaborator(...)` function further down.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+                CollaboratorsActionResult.NoteDeleted -> _uiState.value = UiState.NoteDeleted
+                CollaboratorsActionResult.NoteInTrash -> _uiState.value = UiState.NoteInTrash
+            }
+        }
+    }
+
+    sealed class UiState {
+        object NoteInTrash : UiState()
+        object NoteDeleted : UiState()
+        object EmptyCollaborators: UiState()
```
ParaskP7: Suggestion (💡): Consider running reformat on this class or any other class before committing it. For example, in this case the colon (`:`) would automatically correct itself.

PS: For example, same applies to the `AddCollaboratorFragment` class and its `binding` field definition. This commit dialog automation saves you from these pain points.

Simplenote/src/main/java/com/automattic/simplenote/AddCollaboratorFragment.kt
```diff
@@ -0,0 +1,87 @@
...
+                binding.collaboratorInput.error = null
+            }
+        }
+
+        return dialogEditTag
+    }
+
+    override fun onShow(dialog: DialogInterface?) {
+        setObservers()
+        setupViews()
```
ParaskP7, line 62: Minor (🔍): Consider flipping these two lines to match with the way those two functions are being layered afterwards, mainly for consistency purposes.

Simplenote/src/main/java/com/automattic/simplenote/AddCollaboratorFragment.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private fun setupViews() {
+        val positiveButton = dialogEditTag.getButton(DialogInterface.BUTTON_POSITIVE)
```
ParaskP7, line 67: Suggestion (💡): If a variable is not needed, consider not exposing it to make it explicit that this is not required. Here for example, both `positiveButton` and `collaborator` could be omitted and inlined instead:
```
    private fun setupViews() {
        dialogEditTag.getButton(DialogInterface.BUTTON_POSITIVE).setOnClickListener {
            viewModel.addCollaborator(noteId, binding.collaboratorText.text.toString())
        }
    }
```

Simplenote/src/main/java/com/automattic/simplenote/AddCollaboratorFragment.kt
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    private fun setObservers() {
+        viewModel.event.observe(this, { event ->
+            when (event) {
+                AddCollaboratorViewModel.Event.Close,
+                // In case the note is deleted or trashed, we let the CollaboratorsActivity to handle it
+                AddCollaboratorViewModel.Event.NoteDeleted,
+                AddCollaboratorViewModel.Event.NoteInTrash,
```
ParaskP7: Minor (🔍): Reformatting this class updates the indentation of this comment. Instead of putting the comment there, which applies to only two of those cases, consider adding a separate comment for each, to make it explicit. Also, note the `we` that got removed in the below alternative:
```
                AddCollaboratorViewModel.Event.NoteDeleted, // In case the note is deleted, the activity handles it.
                AddCollaboratorViewModel.Event.NoteInTrash, // In case the note is trashed, the activity handles it.
```

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+        val noteId = intent.getStringExtra(NOTE_ID_ARG)
+        if (noteId == null) {
+            finish()
+            return
+        }
+
+        binding.setupViews()
+        binding.setObservers()
+
+        viewModel.loadCollaborators(noteId)
```
ParaskP7: Suggestion (💡): The `binding` variable doesn't need to be exposed. Consider the `with(...) { ... }` pattern instead:
```
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)

        val noteId = intent.getStringExtra(NOTE_ID_ARG)
        if (noteId == null) {
            finish()
            return
        }

        with(ActivityCollaboratorsBinding.inflate(layoutInflater)) {
            setContentView(root)

            setupViews()
            setObservers()

            viewModel.loadCollaborators(noteId)
        }
    }
```

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private fun ActivityCollaboratorsBinding.setupViews() {
+        val toolbar = findViewById<Toolbar>(R.id.toolbar)
+        setSupportActionBar(toolbar)
+
+        supportActionBar?.apply {
+            title = SpannableString(getString(R.string.collaborators))
+            setDisplayHomeAsUpEnabled(true)
+        }
```
ParaskP7, line 87: Suggestion (💡): Consider extracting these toolbar specific setup to its own function: `setupToolbar()`

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+
+        // Hide all views while loading
+        sharedMessage.visibility = View.GONE
+        collaboratorsList.visibility = View.GONE
+        dividerLine.visibility = View.GONE
+        buttonAddCollaborator.visibility = View.GONE
+        emptyMessage.visibility = View.GONE
```
ParaskP7: Question (❓): Why not have the views start on a `GONE` state so that to avoid setting them to `GONE` during `onCreate(...)`?

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+        collaboratorsList.layoutManager = LinearLayoutManager(this@CollaboratorsActivity)
+
+        buttonAddCollaborator.setOnClickListener { viewModel.clickAddCollaborator() }
+    }
+
+    private fun ActivityCollaboratorsBinding.setObservers() {
+        viewModel.uiState.observe(this@CollaboratorsActivity, { uiState ->
+            when (uiState) {
+                EmptyCollaborators -> handleEmptyCollaborators()
+                is CollaboratorsList -> handleCollaboratorsList(uiState)
```
ParaskP7: 1. Suggestion (💡): Consider adding `is` to all when cases, for consistency purposes.
2. Suggestion (💡): Consider passing only the exact value need to a function instead of passing the whole object, which might not be needed at all. For example in this case, passing only the `uiState.collaborators` to the `handleCollaboratorsList(...)` function will make it more explicit that this is the only value needed.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    private fun ActivityCollaboratorsBinding.handleCollaboratorsList(uiState: CollaboratorsList) {
+        sharedMessage.visibility = View.VISIBLE
+        collaboratorsList.visibility = View.VISIBLE
+        dividerLine.visibility = View.VISIBLE
+        buttonAddCollaborator.visibility = View.VISIBLE
+        emptyMessage.visibility = View.GONE
+
+        val adapter = collaboratorsList.adapter as CollaboratorsAdapter
+        adapter.submitList(uiState.collaborators)
```
ParaskP7: Suggestion (💡): Consider making this a one-liner: `(collaboratorsList.adapter as CollaboratorsAdapter).submitList(collaborators)`

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private fun ActivityCollaboratorsBinding.handleEmptyCollaborators() {
+        // Hide all views while loading
```
ParaskP7: Minor (🔍): This might be a copy-paste mistake, consider removing this comment.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+        val dialog = AddCollaboratorFragment(event.noteId)
+        dialog.show(supportFragmentManager.beginTransaction(), DIALOG_TAG)
+    }
+
+    private fun showRemoveCollaboratorDialog(event: Event.RemoveCollaboratorEvent) {
+        val alert = AlertDialog.Builder(ContextThemeWrapper(this, R.style.Dialog))
+        alert.setTitle(R.string.remove_collaborator)
+        alert.setMessage(R.string.remove_collaborator_message)
+        alert.setNegativeButton(R.string.cancel, null)
+        alert.setPositiveButton(R.string.remove) { _: DialogInterface?, _: Int ->
```
ParaskP7: Minor (🔍): Consider removing the types from the unused lambda parameters, as they are not needed:
```
        alert.setPositiveButton(R.string.remove) { _, _ ->
            viewModel.removeCollaborator(event.collaborator)
        }
```

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
```diff
@@ -801,6 +804,22 @@ public boolean onOptionsItemSelected(@NonNull MenuItem item) {
...
+            return;
+        }
+
+        Intent intent = new Intent(requireActivity(), CollaboratorsActivity.class);
+        intent.putExtra(CollaboratorsActivity.NOTE_ID_ARG, mNote.getSimperiumKey());
+        startActivity(intent);
+
+        AnalyticsTracker.track(
+                AnalyticsTracker.Stat.EDITOR_COLLABORATORS_ACCESSED,
+                CATEGORY_NOTE,
```
ParaskP7, line 818: Minor (🔍): Consider removing the `import static com.automattic.simplenote.analytics.AnalyticsTracker.CATEGORY_NOTE;` import and making this and all other such usages a `AnalyticsTracker.CATEGORY_NOTE` instead.

Simplenote/src/main/res/layout/activity_collaborators.xml
```diff
@@ -0,0 +1,78 @@
+<LinearLayout
```
ParaskP7: 1. Minor (🔍): Consider adding the missing `<?xml version="1.0" encoding="utf-8"?>` on top.
2. Minor (🔍): Consider reformatting this file
3. Minor (🔍): Consider collapsing any empty tags (like the `include`).

PS: Same apply to `add_collaborator.xml` and `collaborator_raw`.
PSS: Consider renaming `add_collaborator.xml` to include the `fragment` word in it, like `add_collaborator_fragment.xml`.

Simplenote/src/main/AndroidManifest.xml
```diff
@@ -162,6 +162,11 @@
             android:name="net.openid.appauth.RedirectUriReceiverActivity"
             tools:node="replace" />
 
+        <activity
+            android:name=".CollaboratorsActivity"
+            android:configChanges="orientation|keyboardHidden|screenSize"
+            android:exported="false" />
+
```
ParaskP7, line 169: Question (❓): I am not sure what happened, but as part of the duplicate commits problem I was seeing, this `AndroidManifest.xml` file was reformatted along with other unrelated changes, see `Add initial (non-functional) version for CollaboratorsActivity` commit. How did that happen?

Simplenote/src/test/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepositoryTest.kt
```diff
@@ -34,4 +64,93 @@ class SimperiumCollaboratorsRepositoryTest {
         assertFalse(collaboratorsRepository.isValidCollaborator("あいうえお@example.com"))
         assertFalse(collaboratorsRepository.isValidCollaborator("just”not”right@example.com"))
     }
+
+    @Test
+    fun getCollaboratorsShouldReturnJustEmails() = runBlockingTest {
```
ParaskP7, line 69: Suggestion (💡): Consider moving and splitting the test according to the `arrange/act/assert` structure, with up to 2 empty lines to split the `arrange` from the `act` and the `act` from the `assert`. Also, make sure to have the `expected` as part of the `assert` because this belongs there. For example in this case, the test would be written as:
```
    @Test
    fun getCollaboratorsShouldReturnJustEmails() = runBlockingTest {
        val result = collaboratorsRepository.getCollaborators(noteId)

        val expected = CollaboratorsActionResult.CollaboratorsList(listOf("test@emil.com", "name@example.co.jp"))
        assertEquals(expected, result)
    }
```
I know that this is a very delicate detail, but imagine inlining the `expected` variable, if so, where would that be? I ask this kind of question to help me understand where to put such code.

Another example would be the below `getCollaboratorsWhenNoteInTrashShouldReturnError()` test, which I would have written as:
```
    @Test
    fun getCollaboratorsWhenNoteInTrashShouldReturnError() = runBlockingTest {
        note.isDeleted = true

        val result = collaboratorsRepository.getCollaborators(noteId)

        val expected = CollaboratorsActionResult.NoteInTrash
        assertEquals(expected, result)
    }
```

Another, more obvious case, with lots of such problems would have been the `addCollaboratorShouldAddATagToNote()` test, which I would have written as:
```
    @Test
    fun addCollaboratorShouldAddATagToNote() = runBlockingTest {
        val collaborator = "test1@email.com"

        val result = collaboratorsRepository.addCollaborator(noteId, collaborator)

        val newCollaborators = listOf("test@emil.com", "name@example.co.jp", collaborator)
        val expected = CollaboratorsActionResult.CollaboratorsList(newCollaborators)
        assertEquals(expected, result)
    }
```

PS: The same applies to all the test written in this PR. Also, try to do the same for the `NoteTest` test suite, which in addition has another problem, the `expected` variable is added as the `actual` argument on the `assertEquals(...)` function, which might create confusion when a test fails.

Simplenote/src/test/java/com/automattic/simplenote/usecases/ValidateTagUseCaseTest.kt
```diff
@@ -1,17 +1,24 @@
...
+import kotlinx.coroutines.test.TestCoroutineDispatcher
 import org.junit.Assert.assertEquals
 import org.junit.Test
 import org.mockito.Mockito
 import org.mockito.kotlin.any
 import org.mockito.kotlin.whenever
 
+@ExperimentalCoroutinesApi
 class ValidateTagUseCaseTest {
+    private val notesBucket = Mockito.mock(Bucket::class.java) as Bucket<Note>
```
ParaskP7, line 18: Suggestion (💡): Consider suppressing this unchecked cast warning with a `@Suppress("UNCHECKED_CAST")` annotation.

ParaskP7 (changes requested):
👋 @danilo04 !

I have reviewed and tested this PR as per the instructions, everything works as expected, awesome job! 🌟 🌟 🌟 

I have left a warning (⚠️), some questions (❓), few minor (🔍) comments and a couple of suggestions (💡) for you to consider. I am **NOT** going to approve this PR for now, just because of the warning (⚠️) I mentioned. However, this is a test related warning (⚠️) so don't be alarmed too much.

PS: I wanted to ask you about the commits. I was seeing a lot of duplicate commits in the history log and that kind of alerted me. Why did that happen?
PSS: This PR was indeed a big one and I am sure that I would have been more thorough in my code review and testing if it split into separate PRs. I was wondering why it couldn't have been split into multiple PRs. Would you like us to have a discussion on that, maybe I can help by providing some advises for next time.

Simplenote/src/main/java/com/automattic/simplenote/di/ThreadModule.kt
danilo04: This is because `runBlockingTest` still doesn't manage multithreading well. There is an [open issue](https://github.com/Kotlin/kotlinx.coroutines/issues/1222) that discusses something about it. They tried a fix but was not successful and found that the change required is more complex. 

> This is totally unrelated to this PR, but SimperiumTagsRepository is untested, maybe we should create an issue to make sure that to cover this repository with adequate tests as well. Wdty?

Yes, definitely. I will add tests to this repo in this sprint.

Simplenote/src/main/java/com/automattic/simplenote/repositories/CollaboratorsRepository.kt
danilo04: Fixed in 027ae7e01107c70acbe7ac10dd20f2b191919a55.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepository.kt
danilo04: Yes, this looks more precise. Fixed in fa33aa22c13c2f925374c4a61317bbfea6ce648b.

Simplenote/src/main/java/com/automattic/simplenote/utils/CollaboratorsAdapter.kt
danilo04: Fixed in cb3d1ad7a4d27d7c44c73ecb1bb4eecdc8c0d467.

Simplenote/src/main/java/com/automattic/simplenote/utils/CollaboratorsAdapter.kt
danilo04: Fixed in 9d054fd8d4e6217180a1f813bed2adc52b4b8c5c.

Simplenote/src/main/java/com/automattic/simplenote/utils/Either.kt
danilo04: The problem with `Result` is that there is a `Result` in Kotlin's standard library and other places in the Android framework which makes the use confusing. So I am hesitant here to change it to `Result`. Do you think this is not a problem at all?

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModel.kt
danilo04: Yes, I agree. Fixed in bb411c547b8e3df6296ee23f11a24c28079c14b7.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModel.kt
danilo04: Fixed in d0f75315e35457fe7f3589112a5338ffda6185f8.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModel.kt
danilo04: Yes, I need to configure this in the IDE. Fixed in 79168d29bb35f45a3d0b0fbb9521f26b1c003980.

Simplenote/src/main/java/com/automattic/simplenote/AddCollaboratorFragment.kt
danilo04: Fixed in c2484033c0816f8efef7c06ac645778a0a9c3010.

Simplenote/src/main/AndroidManifest.xml
danilo04: When I added `CollaboratorsActivity`, Android Studio modified the `AndroidManifest.xml` file but restructure it as well. In later commits I took what it was like before the commit and just added the `CollaboratorsActivity` tag.

Simplenote/src/main/java/com/automattic/simplenote/AddCollaboratorFragment.kt
danilo04: I prefer to have the variable there. Sometimes if the expression is too long, I read it as a long sentence without punctuation. It is just my preference though. Also, the compiler takes care of this if it needs to be optimized or something.

Simplenote/src/main/java/com/automattic/simplenote/AddCollaboratorFragment.kt
danilo04: Yes, it is a better format. Fixed in f44e932d7d752a22240a55840965727ef65d1ec4.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
danilo04: Yes, this sounds more idiomatic Kotlin. Fixed in 9cff3e148f9b1dd41932938eb488b3808ad8ae3f and 833acdf94ded320a11594894e760df903f85c136.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
danilo04: Makes sense. Fixed in 198ee8bf95b75dc7bd24c6a93c7bd3d07237d02a.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
danilo04: Fixed in f731ffcdd94c2a06737470731ecb4ae6dd35e2d9.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
danilo04: Fixed in 78df3d6a74277596e9ce54a4228c240d27d2ba16.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
danilo04: My bad. Fixed in d321b8424b02eabcb5119c713a575bc8c5a73065.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
danilo04: Fixed in a2229c9301906d17d54b9f9a1a151e08d9c292b1.

Simplenote/src/main/res/layout/activity_collaborators.xml
danilo04: Fixed in af1df21887dc4821ba91908e49162460cdc11e3b.

Simplenote/src/test/java/com/automattic/simplenote/usecases/ValidateTagUseCaseTest.kt
danilo04: I am gonna fix this in a next PR where I will include tests for `SimperiumTagsRepository` and other improvements. I am planning to use the `@Mock` annotation and run the tests with the `Mockito` runner. This change will remove all warnings.

Simplenote/src/main/java/com/automattic/simplenote/CollaboratorsActivity.kt
danilo04: Yes, that makes sense. Fixed in 091845455f75746399cf64b549fa7066e335c025.

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
danilo04: I will not make a change about this because the file is too inconsistent with the format. In some cases, the enum type is imported completely, in others not. I will change this when I prefer to rewrite this when we change the file to Kotlin.

Simplenote/src/test/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepositoryTest.kt
danilo04: That is a good point. The structure `arrange/act/assert` defines clear tests even if not enforced by a framework. Fixed in 8cd381cff22c86fc71cb95f17c05b323e8fdc5ef and 9c9395455e4f8b194f7f10eeb79fcba71a3cf8cc.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepository.kt
danilo04: Fixed in 56de650bed91f02c939abdcf2ce662415da8341d. I had to mock `CollaboratorsRepository` in order to be able to test this function. Other approaches such as `spy` the class or partial mocks were not straightforward and did not work well with `coroutines` and `flows`.

danilo04:
Thank you for the thorough review as always @ParaskP7. I have made most of the changes requested and commented in some requests that I did not apply any change. Let me know WDYT?

We have commented over other channels about the long PR and the duplicated commits.

Simplenote/src/main/java/com/automattic/simplenote/di/ThreadModule.kt
ParaskP7: > This is because runBlockingTest still doesn't manage multithreading well. There is an open issue that discusses something about it. They tried a fix but was not successful and found that the change required is more complex.

Hmmm... this is very interesting. So, the problem is purely because of testing, right? If so, what kind of problems did you find while doing testing without the `ThreadModule` alternative?

Questions (❓): If we are to stick with the `ThreadModule`, should the `SimperiumTagsRepository` be updated as well?

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepository.kt
ParaskP7: 👋 @danilo04 !

Thanks for adding those two test. However, I am a bit confused by the value added here, can we discuss it a bit before merging? 🤔 

Question (❓): In addition the test you added on the `ViewModel` layer, is it possible to add `Repository` layer test for the `collaboratorsChanged()`? PS: Same question applies to `SimperiumTagsRepository.tagsChanged()`.

Simplenote/src/main/java/com/automattic/simplenote/utils/Either.kt
ParaskP7: I have been using `Result` in the past and it wasn't a problem, you just need to import the correct `Result` and that's it. I understand that Kotlin's standard library `Result` can get in the way, but in my experience this didn't happen to me.

My main problem with `Either` is the fact that one would need to know what is `Left` what is `Right` what is `l` and what is `r` correspondingly. I am always confused about that and end up reading the class to understand which one is the success and which one is the error. Also, IMHO the `Either` generic definition is unnecessary complicated, at least for our use-cases here:
- `Either<L, Nothing>()` vs. `Result<T>()`
- `Either<Nothing, R>()` vs. `Result<Nothing>()`

Having said the above, please feel free to go with whichever you feel it more appropriate for Simplenote, I was just sharing my thoughts and experiences, nothing more, nothing less. I would be comfortable working with an `Either` either way... see what I did here! 😄

ParaskP7:
👋 @danilo04 !

> Thank you for the thorough review as always @ParaskP7. I have made most of the changes requested and commented in some requests that I did not apply any change. Let me know WDYT?

Thank YOU for applying the suggestions in such a graceful way. I have just left a couple of comments unresolved for us to have one final discussion. However, I don't want to block this PR anymore and will approve it no matter. Feel free to merge it when you feel you are ready.

> We have commented over other channels about the long PR and the duplicated commits.

💯

ParaskP7 (approved):

Simplenote/src/main/java/com/automattic/simplenote/di/ThreadModule.kt
danilo04: >  If so, what kind of problems did you find while doing testing without the ThreadModule alternative?

Mostly `IllegalStateException: This job has not completed yet` exceptions. Even though I was running the suspend functions inside blocks of `runBlockingTest`, the coroutines did not finish executing before I assert the results.

danilo04: > If we are to stick with the ThreadModule, should the SimperiumTagsRepository be updated as well?

Yes. Fixed in 32eca5b829a1a552d1bbb82ad0144a4393e91e2d.

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepository.kt
danilo04: Hey @ParaskP7 👋 !

I need to find a way to inject the calls to notify the callbacks. Since we are using mocks for the `Bucket` objects, I need to figure out whether a partial mock or some other technique helps me to be able to trigger the different callbacks. Do you have any idea how we could test these flows at the repository level?

Simplenote/src/main/java/com/automattic/simplenote/di/ThreadModule.kt
ParaskP7: 👋 @danilo04 !

> Mostly IllegalStateException: This job has not completed yet exceptions. Even though I was running the suspend functions inside blocks of runBlockingTest, the coroutines did not finish executing before I assert the results.

I see... thanks once more for the elaborate explanation and for updating `SimperiumTagsRepository` as well! 🚀

Simplenote/src/main/java/com/automattic/simplenote/repositories/SimperiumCollaboratorsRepository.kt
ParaskP7: 👋 @danilo04 !

> Do you have any idea how we could test these flows at the repository level?

Yeap, I am not sure, I would need to think and play a bit with it before me suggesting a way to add these tests. Please feel free to progress without that change, but can you please add an small issue on that so that we can revisit it later. 🙏

Simplenote/src/main/java/com/automattic/simplenote/utils/Either.kt
danilo04: Yes, it is definitely simpler. But for this case, the simplification doesn't work. The final operation is converging into a `CollaboratorsActionResult` type since I didn't want to use exceptions. This operation can be short-circuit If the note is in the trash or deleted. It would be more simpler if I added something like `flatMap` to handle left or right results (something like [railway programming](https://www.greenbird.com/news/railway-oriented-programming-in-kotlin)). 

I will leave like this for this PR, but I will consider changing it if I see some problems in future PRs. Also, I will keep an eye to external PRs to see if the use of `Either` was not considered because of the name of lack of knowledge on functional programming.

SylvesterWilmott:
Haven't had an opportunity to fully test but some observations so far:

- "Add Collaborator" button should not be a flat button but should instead be a tappable row.
- The list of collaborators has additional padding on the right, eg. the touch feedback of the "x" is cut off on the right-hand side. This is in contrast to similar layouts like tags edit view.
- Once a collaborator is added, the email appears in the tags list.
- The "Email Address" input blue colour does not adapt to dark mode. Blue30 should be used here.
- The "x" icons in the collaborator list don't adapt to dark mode. Gray30 should be used here.

danilo04:
Hi @SylvesterWilmott 👋 . Thanks for the review.  I have made the following changes:

> "Add Collaborator" button should not be a flat button but should instead be a tappable row.

Do you mean the whole row where Add Collaborator is, is tappable? 
I have made the whole row tappable in 74171edecbba4b78e9303bdda4807a106617684d but I am not sure if that is what you mean. 

> The list of collaborators has additional padding on the right, eg. the touch feedback of the "x" is cut off on the right-hand side. This is in contrast to similar layouts like tags edit view.

I have removed the padding but part of the touch feedback is cut off similar to the delete button in edit tags (fixed in 74171edecbba4b78e9303bdda4807a106617684d).

> Once a collaborator is added, the email appears in the tags list.

~I am not sure about this. Could you expand a little bit?~
This will be handled in a different PR.

> The "Email Address" input blue colour does not adapt to dark mode. Blue30 should be used here.

Fixed in 84ee1e42fc2a81996442bebbcc697e04a5b92c32.

> The "x" icons in the collaborator list don't adapt to dark mode. Gray30 should be used here.

Fixed in 74730be846755fb68be55b5a74d6da98383dffa6.

Simplenote/src/main/java/com/automattic/simplenote/utils/Either.kt
ParaskP7: 👋 @danilo04 !

Once more, thanks a lot for all the clarification! 🙏 

Yes, let's then keep the `Either` for now and see how it goes...

SylvesterWilmott:
Thanks for the changes @danilo04 

All looks good. Just one more request, let's add the touch feedback to the "Add Collaborator" row.

danilo04:
Hi @SylvesterWilmott.

> Just one more request, let's add the touch feedback to the "Add Collaborator" row.

Fixed in 208f3435bcbd4edd252328e0d4183276e3823405. Let me know if it works OK on your part.

SylvesterWilmott:
It looks like this PR was merged before the `needs design review` label was removed but there are a couple more issues we can tackle in the next PR.

1. The "Add Collaborator" row should be 48dp tall with 8dp top and bottom margins. This means the area that is tappable should only be 48dp. This is illustrated in Zeplin by clicking on the bounding boxes in the design:

Screenshot 2021-09-24 at 09 29 40

2. It looks like it's possible to add yourself as a collaborator. While we validate the email, let's also make sure that the email doesn't match the email of the currently logged in account. In this event we can show the error `You cannot add yourself as a collaborator` in the red error text below the input.

PR #1504, opened 2021-09-22
Avoid hiding/showing toolbar when note reference is null

### LINKED ISSUE #1490: App crashes when focus changed and note reference is null
We observed that the note reference in `NoteEditorFragment` can be `null` when the content editor loses or gains focus while editing in landscape mode. In these cases, the app crashes. 

Sentry Issue: 2645499426

### Expected

If the `mNote` reference is `null`, do not do anything. 

### Observed
The app crashes

### Reproduced

1. Go to a Note
2. Rotate phone to landscape mode
3. Try to gain and lose focus on the content editor

Device Make: Xiaomi
Device Model: Mi A2
Android Version: 10.0.0
Simplenote Version: 2.21-rc-1

danilo04:
Fixes #1490.

### Fix

There are some cases in which the note reference `mNote` in `NoteEditorFragment` can be `null` when the content editor loses or gains focus while editing in landscape mode. In these cases, the app crashes. This PR just dismisses the hiding/showing of the toolbar if the note reference is `null`, avoiding NPEs. We assume that if the note reference is `null` is because there is not a note selected, therefore the behavior to hide/show the toolbar is not needed.

### Test

1. Go to settings and enable dark mode
2. Tap on a new note or a note from the notes list
3. Start typing on the note
4. Rotate the phone to landscape mode
5. The tabs and toolbar should hide and the background of the text editor should remain dark
6. Rotate the phone again
7. The tabs and toolbar should be shown again
8. Rotate the phone to landscape mode
9. The tabs and toolbar should hide and the background of the text editor should remain dark
10. Hide the keyboard
11. The tabs and toolbar should be shown again
12. Disable markdown
13. Rotate the phone to landscape mode
14. Start typing on the note
15. The toolbar should hide and the background of the text editor should remain dark
16. Hide the keyboard
17. The toolbar (not the tabs) should be shown again

### Review

### Release

`RELEASE-NOTES.txt` was updated in  156843278c07e86e7394b9f3028073578b65fc61 with:
> Fixed crash when note reference is null in editor while hiding toolbar in landscape mode


hichamboushaba (approved):
LGTM, the behavior seems correct for all cases.

There is one conflict to fix before merging though.

danilo04:
Hi @hichamboushaba 👋 . Thanks for the review. I have fixed the conflict (it was the `RELEASE_NOTES.txt` file)

PR #1505, opened 2021-09-22
Add simperium strings to handle compromised passwords to Simplenote

### LINKED ISSUE #1506: Update copy of compromised passwords alert
Update copy of compromised passwords alert based on changes in [this PR](https://github.com/Automattic/simplenote-ios/pull/1441).

Title: `Compromised Password`

Message: `This password has appeared in a data breach, which puts your account at high risk of compromise. To protect your data, you'll need to update your password before being able to log in again.`

Buttons: `Change Password`, `Cancel`

danilo04:
Fixes #1506

### Fix
Login in Simplenote is handled by the simperium library. In order to make the string translatable at the Simplenote level, we need to include those strings in the project. This PR adds the string to handle compromised passwords during login. 

It also updates the strings with respect to the request in #1506. 

### Test

1. Create an account and use a compromised password such as `password` or `123456`
2. Try login into the account
3. A dialog with the title `Compromised Password` and the message `This password has appeared in a data breach, which puts your account at high risk of compromise. To protect your data, you'll need to update your password before being able to log in again.` should appear. 

### Review

Only one developer or designer is required to review these changes, but anyone can perform the review.

### Release

### DIFF
--- a/Simplenote/src/main/res/values/strings.xml
+++ b/Simplenote/src/main/res/values/strings.xml
@@ -404,8 +404,8 @@
     <string name="simperium_account_verification_message">You must verify your email before logging in to your account.</string>
     <string name="simperium_account_verification_url">https://app.simplenote.com/account/verify-email/%1$s</string>
     <string name="simperium_compromised_password">Compromised Password</string>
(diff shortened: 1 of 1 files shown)


AmandaRiu:
@danilo04 Do you have an account with a compromised password that I can use? I cannot create one:

sandymcfadden (approved):
I've tested this out and is working as expected with the new strings!

PR #1507, opened 2021-09-24
Automationuitests

milestonework:
### Fix
1. Added 3  UI tests for Login activity - Login with correct credentials, Login with incorrect credentials and Logout
2. Created a base startup framework for Page Object Model so as to isolate Tests from logic 
3. Created a singleton class 'TestDriver' and added some helper methods for universal use

### Test
There is pending work to be done to perform the necessary checks before starting a Test i.e a setup() before each test is yet to be implemented. Hence, at present each of the UI tests need to be run individually to see the desired behavior. 
Note - The Logout test performs what is expected(does log out successfully) but throws a PerformException due to which the Test fails. 

### Review

### Release

### DIFF
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/uipages/LoginPage.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/uipages/LoginPage.kt
@@ -1,0 +1,12 @@
+package com.automattic.simplenote.uipages
+
+import androidx.test.uiautomator.UiObject
+
+const val LOGINPAGE_LOGINBUTTON: String = "com.automattic.simplenote.debug:id/button"
+const val FORGOT_PASSWORD_LINK: String = "com.automattic.simplenote.debug:id/text_footer"
+const val PASSWORD_HIDE_IMAGE: String = "com.automattic.simplenote.debug:id/text_input_end_icon"
+
+class LoginPage: BasePage() {
+    val loginPageLoginButton: UiObject = getUIDeviceInstance().findObject(getUISelectorInstance().resourceId(LOGINPAGE_LOGINBUTTON))
+
+}
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
@@ -1,0 +1,95 @@
+package com.automattic.simplenote.uitests.login
+
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers.*
+import androidx.test.filters.MediumTest
+import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner
+import com.automattic.simplenote.BaseUITest
+import com.automattic.simplenote.R
+import com.automattic.simplenote.utils.Constants
+import org.junit.Test
+import org.junit.runner.RunWith
+import androidx.test.espresso.matcher.ViewMatchers.withClassName
+import androidx.test.espresso.matcher.ViewMatchers.isDescendantOfA
+import com.automattic.simplenote.utils.TestDriver
+import org.hamcrest.Matchers.*
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions
+import androidx.test.espresso.action.ViewActions.*
+import androidx.test.espresso.matcher.ViewMatchers
+import androidx.test.espresso.matcher.ViewMatchers.isClickable
+import androidx.test.espresso.matcher.ViewMatchers.withId
+import org.junit.Assert
+
+@RunWith(AndroidJUnit4ClassRunner::class)
+@MediumTest
+class LoginTests : BaseUITest() {
+
+    @Test
+    fun testLoginWithWrongPasswordFails() {
+        onView(withId(R.id.button_login))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(withId(R.id.button_email))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(allOf(isDescendantOfA(withId(R.id.input_email)),
+            withParent(withClassName(endsWith("android.widget.FrameLayout"))),
+            withHint("Email")))
+            .check(matches(isDisplayed()))
+            .perform(replaceText(Constants.ACCOUNT_EMAIL), closeSoftKeyboard())
+
+        onView(allOf(isDescendantOfA(withId(R.id.input_password)),
+            withParent(withClassName(endsWith("android.widget.FrameLayout"))),
+            withHint("Password")))
+            .check(matches(isDisplayed()))
+            .perform(replaceText(Constants.INCORRECT_ACCOUNT_PASSWORD), closeSoftKeyboard())
+
+        onView(withId(R.id.button))
+            .check(matches(isDisplayed()))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(allOf((withId(R.id.alertTitle)),
+            withText("Error")))
+            .check(matches(isDisplayed()))
+    }
+
+    @Test
+    fun testLoginWithCorrectPasswordSucceeds() {
+
+        Assert.assertTrue(
+            "User is not on Login page", TestDriver.isUserOnLoginPage())
+
+        Assert.assertTrue(
+            "Login was not successful",
+            TestDriver.logInWithCredentials(Constants.ACCOUNT_EMAIL, Constants.ACCOUNT_PASSWORD))
+    }
+
+    @Test
+    fun testLogOut(){
+
+        onView(allOf(withParent(withId(R.id.toolbar)), withContentDescription("Open drawer")))
+            .check(matches(isDisplayed()))
+            .perform(click())
+
+        onView(allOf(withId(R.id.design_menu_item_text), withText("Settings")))
+                .check(matches(isDisplayed()))
+                .perform(click())
+
+        onView(allOf(withParent(withId(R.id.toolbar)), withText("Settings")))
+            .check(matches(isDisplayed()))
+
+        onView(withId(R.id.preferences_container)).perform(swipeUp())
+
+        onView(allOf(isDescendantOfA(withId(R.id.recycler_view)),
+            withText("Log out")
+        ))
+            .check(matches(ViewMatchers.isDisplayed()))
+            .perform(click())
+
+        Assert.assertTrue("Log Out failed",TestDriver.launchPage.signupButton.exists())
+    }
+}
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/setup/SignUpTest.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/setup/SignUpTest.kt
@@ -1,0 +1,37 @@
+package com.automattic.simplenote.uitests.setup
+
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions
+import androidx.test.espresso.action.ViewActions.click
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers.*
+import androidx.test.filters.MediumTest
+import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner
+import com.automattic.simplenote.BaseUITest
+import com.automattic.simplenote.R
+import com.automattic.simplenote.utils.Constants
+import com.automattic.simplenote.utils.TestDriver
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4ClassRunner::class)
+@MediumTest
+class SignUpTest : BaseUITest(){
+
+    @Test
+    fun newUserSignUpWithMixedCapitalization() {
+
+        onView(withId(R.id.button_signup))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(withId(R.id.email_edit_text))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(withId(R.id.email_edit_text))
+         .perform(ViewActions.replaceText(Constants.SIGNUP_ACCOUNT_EMAIL), ViewActions.closeSoftKeyboard())
+
+        TestDriver.tapButton(Constants.SIGNUP_BUTTON_STR)
+    }
+}
--- a/Simplenote/src/androidTest/java/com/automattic/simplenote/utils/Constants.kt
+++ b/Simplenote/src/androidTest/java/com/automattic/simplenote/utils/Constants.kt
@@ -1,0 +1,15 @@
+package com.automattic.simplenote.utils
+
+class Constants {
+    companion object {
+
+        const val SIGNUP_BUTTON_STR = "SIGN UP"
+        val CLASS_IMAGEBUTTON: String = "android.widget.ImageButton"
+        val MAIN_TOOLBAR: String = "com.automattic.simplenote.debug:id/toolbar"
+
+        val SIGNUP_ACCOUNT_EMAIL: String = "tests.automation@trial.project"
+        val ACCOUNT_EMAIL: String = "tests.automation@trial.project"
+        val ACCOUNT_PASSWORD: String = "tests.automation.pw"
+        val INCORRECT_ACCOUNT_PASSWORD: String = "invalidpassword"
+    }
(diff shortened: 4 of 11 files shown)


pachlava:
Thanks for the PR and for providing the extra details! 👍  I'll start reviewing today and will leave some notes, but it might be stretched to the next day.

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
```diff
@@ -0,0 +1,95 @@
...
+        onView(withId(R.id.button_login))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(withId(R.id.button_email))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(allOf(isDescendantOfA(withId(R.id.input_email)),
+            withParent(withClassName(endsWith("android.widget.FrameLayout"))),
```
pachlava, line 39: Just curious about the reason to include the `withParent` matcher here and below. Did you notice any stability issues when it's not used?

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
```diff
@@ -0,0 +1,95 @@
...
+            .check(matches(isDisplayed()))
+            .perform(replaceText(Constants.INCORRECT_ACCOUNT_PASSWORD), closeSoftKeyboard())
+
+        onView(withId(R.id.button))
+            .check(matches(isDisplayed()))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(allOf((withId(R.id.alertTitle)),
+            withText("Error")))
```
pachlava, line 56: What do you think of making this test more robust, and checking for the actual error message (`Could not log in with the provided...`)? While checking for `Error` text presence is good, it does not give us enough confidence about the reason.

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
```diff
@@ -0,0 +1,95 @@
...
+import androidx.test.espresso.matcher.ViewMatchers.isClickable
+import androidx.test.espresso.matcher.ViewMatchers.withId
+import org.junit.Assert
+
+@RunWith(AndroidJUnit4ClassRunner::class)
+@MediumTest
+class LoginTests : BaseUITest() {
+
+    @Test
+    fun testLoginWithWrongPasswordFails() {
```
pachlava, line 29: I like how you keep the test names consistent with the ones in the testing checklist. It's much easier to track them this way 👍

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
```diff
@@ -0,0 +1,95 @@
...
+            withHint("Password")))
+            .check(matches(isDisplayed()))
+            .perform(replaceText(Constants.INCORRECT_ACCOUNT_PASSWORD), closeSoftKeyboard())
+
+        onView(withId(R.id.button))
+            .check(matches(isDisplayed()))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(allOf((withId(R.id.alertTitle)),
```
pachlava, line 55: I've noticed quite a high flakiness of this test on my setup, which might be related to performance. It looks like the check for `alertTitle` object presence is initiated before it actually appears on the screen, in my case. While I could solve this by adding a 1-second wait before this line, I think it would be better to have a wait with timeout (something similar to [waitForExistence](https://developer.apple.com/documentation/xctest/xcuielement/2879412-waitforexistence) from XCTest for Xcode) .

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/setup/SignUpTest.kt
```diff
@@ -0,0 +1,37 @@
...
+        onView(withId(R.id.email_edit_text))
+            .check(matches(isClickable()))
+            .perform(click())
+
+        onView(withId(R.id.email_edit_text))
+         .perform(ViewActions.replaceText(Constants.SIGNUP_ACCOUNT_EMAIL), ViewActions.closeSoftKeyboard())
+
+        TestDriver.tapButton(Constants.SIGNUP_BUTTON_STR)
+    }
+}
```
pachlava, line 37: Since this test/file is not finished (for a very good reason), I can see value in not having it included into the PR, so that it does not create an impression that the signup area is covered.

Simplenote/src/androidTest/java/com/automattic/simplenote/uipages/LoginPage.kt
```diff
@@ -0,0 +1,12 @@
+package com.automattic.simplenote.uipages
+
+import androidx.test.uiautomator.UiObject
+
+const val LOGINPAGE_LOGINBUTTON: String = "com.automattic.simplenote.debug:id/button"
+const val FORGOT_PASSWORD_LINK: String = "com.automattic.simplenote.debug:id/text_footer"
+const val PASSWORD_HIDE_IMAGE: String = "com.automattic.simplenote.debug:id/text_input_end_icon"
+
+class LoginPage: BasePage() {
+    val loginPageLoginButton: UiObject = getUIDeviceInstance().findObject(getUISelectorInstance().resourceId(LOGINPAGE_LOGINBUTTON))
```
pachlava, line 10: Since we are in the context of `LoginPage`, we already know that we are on the login page, so something like `LoginPage().loginPageLoginButton` looks a bit redundant. Because of this, it could be just something like:

```suggestion
    val loginButton: UiObject = getUIDeviceInstance().findObject(getUISelectorInstance().resourceId(LOGIN_BUTTON))
```

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/Constants.kt
```diff
@@ -0,0 +1,15 @@
...
+
+class Constants {
+    companion object {
+
+        const val SIGNUP_BUTTON_STR = "SIGN UP"
+        val CLASS_IMAGEBUTTON: String = "android.widget.ImageButton"
+        val MAIN_TOOLBAR: String = "com.automattic.simplenote.debug:id/toolbar"
+
+        val SIGNUP_ACCOUNT_EMAIL: String = "tests.automation@trial.project"
+        val ACCOUNT_EMAIL: String = "tests.automation@trial.project"
```
pachlava, line 11: Is there a reason to keep both?

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
```diff
@@ -0,0 +1,95 @@
...
+            .check(ViewAssertions.matches(ViewMatchers.isDisplayed()))
+            .perform(ViewActions.replaceText(Constants.ACCOUNT_PASSWORD), ViewActions.closeSoftKeyboard())
+
+        loginPage.loginPageLoginButton.click()
+
+        // if True (Open Drawer menu is visible) then login was successful
+        return isOpenDrawerMenuVisible()
+    }
+
+    fun isOpenDrawerMenuVisible(): Boolean {
```
pachlava, line 67: Possibly the intention has changed in the process, but knowing the method name, I would expect it to check for the menu drawer presence, however it checks for the search icon.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
```diff
@@ -0,0 +1,95 @@
...
+
+        if(allNotesHomePage.searchIcon.exists())
+            return true
+        else
+            return false
```
pachlava, line 72: Since `exists()` returns a boolean, this part can be shortened to:

```suggestion
        return allNotesHomePage.searchIcon.exists()
```

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    fun logOutUser(): Boolean {
+        return launchPage.signupButton.exists()
+    }
```
pachlava, line 77: Judging by the name, this method is unfinished. I consider that overall code legibility and maintainability benefits from the absence of not used / not finished parts.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
```diff
@@ -0,0 +1,95 @@
...
+
+        Espresso.onView(
+            Matchers.allOf(
+                ViewMatchers.isDescendantOfA(ViewMatchers.withId(R.id.input_email)),
+                ViewMatchers.withParent(ViewMatchers.withClassName(Matchers.endsWith("android.widget.FrameLayout"))),
+                ViewMatchers.withHint("Email")
+            )
+        )
+            .check(ViewAssertions.matches(ViewMatchers.isDisplayed()))
+            .perform(ViewActions.replaceText(Constants.ACCOUNT_EMAIL), ViewActions.closeSoftKeyboard())
```
pachlava, line 49: The method still uses the hardcoded values for both email and password, however, the presence of method parameters implies a more flexible usage.

```suggestion
            .perform(ViewActions.replaceText(email), ViewActions.closeSoftKeyboard())
```

What do you think of having the credentials form `Constants` be used as default/fallback values?

Same for password below.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Tap a button using text
+     */
+    fun tapButton(buttonName: String) {
+        try {
+            val uiDevice = UiDevice.getInstance(androidx.test.platform.app.InstrumentationRegistry.getInstrumentation())
+            val button = uiDevice.findObject(UiSelector().text(buttonName))
+            Assert.assertTrue("Tap button $buttonName...", button.exists() && button.isEnabled)
```
pachlava, line 86: This assertion checks for the object to exist and be enabled, however, the assertion message is about tapping.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
```diff
@@ -0,0 +1,95 @@
...
+            launchPage.loginEmailButton.click()
+        }
+        return loginPage.loginPageLoginButton.exists()
+    }
+
+    /**
+     * Insert the username and password in the LogIn page and does the login
+     */
+    @Throws(UiObjectNotFoundException::class)
+    fun logInWithCredentials(username: String, password: String): Boolean {
```
pachlava, line 39: - Since it's easy to see that credentials are required for the method, having `WithCredentials` included does not look necessary.
- [just a nitpick] In the context of Simplenote, there is no concept of username. Accounts have an associated email and password only.

```suggestion
    fun logIn(email: String, password: String): Boolean {
```

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    fun logOutUser(): Boolean {
+        return launchPage.signupButton.exists()
+    }
+
+    /**
+     * Tap a button using text
+     */
+    fun tapButton(buttonName: String) {
```
pachlava, line 82: I think having the method or parameter named a bit different would remove the necessity of comment for the method. 

Also, since the property by which the button is located is called `text`, I think using this in name would add more clarity to the code.

The third point here it that we're not providing any additional input to `findObject` that will filter only the buttons. In other words, it will work for every object with certain text. Using something like this for the name will reflect what the method does more precisely (just an example):

```suggestion
    fun tapObjectByText(objectText: String) {
```

pachlava (changes requested):
Left some review notes @milestonework 👋 They are mainly language-agnostic ones. There are a few related to naming, and several where I'd suggest some changes.

I left comments only to the first test and to the helpers and surrounding files, keeping in mind that other two tests except for the preconditions that are not currently ensured (like being logged in or logged out).

Thank you for your work 🙂

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
milestonework: Actually I was unable to identify the text field otherwise. Initially I tried using _withClassName(endsWith("android.widget.EditText"))_ but it did not work. Using the _withParent_ matcher however helped.

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
milestonework: For sure, that was definitely one of the items on my list to improve the current tests.

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
milestonework: Yes, an action item for most of the tests is to include the waiting for element to appear. Will work on that.

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/setup/SignUpTest.kt
milestonework: Agree. Since this is my Trial period, as part of giving updates, I was under the impression that I had to include everything I had worked upon thus far to show progress, what I am working on, updates etc. and in general showcase my workings and thought process. In a real working scenario though, definitely not something to include in the PR, its not ready just yet.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
milestonework: Sorry, about that. Actually same reasons as above, I assumed I had to include all work in the PR since I was in Trial. These was one of the methods I was going to use in my tests in order to work towards a POM model.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
milestonework: sure, I will fix the naming here.

Simplenote/src/androidTest/java/com/automattic/simplenote/uipages/LoginPage.kt
milestonework: agree, makes sense.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/Constants.kt
milestonework: Ah, I was going to use a different email for the Sign up, missed updating it.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
milestonework: Absolute blunder on my part :( 
I will fix it.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
milestonework: agree, concise and neat.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
milestonework: That is perfect, makes sense.

Simplenote/src/androidTest/java/com/automattic/simplenote/utils/TestDriver.kt
milestonework: This actually originated from the method name but from the discussion around the method name and intent of it, it would be meaningful to change the assertion message as well. I will work on it.

Simplenote/src/androidTest/java/com/automattic/simplenote/uitests/login/LoginTests.kt
pachlava: Hm, that's interesting 🤔 In my case, removing them had no effect - objects were still located.

PR #1509, opened 2021-09-24
Refactor tag screens to filter out collaborators

### LINKED ISSUE #1480: Refactor tags views to filter out collaborators. This includes: `TagsActivity` and the list of tags shown on the left panel menu.

danilo04:
Fixes #1480

### Fix

This PR makes a series of fixes to make the tag screens consistent with the dedicated UI for collaboration added in #1476. The changes are the following:

- Left panel (navigation drawer) filters out tags that are collaborators (emails)

| Before      | After |
| ----------- | ----------- |
|       |        |

- The list of tags (in the Edit Tags screen) filters out tags that are collaborators (emails)  

| Before      | After |
| ----------- | ----------- |
|       |        |

- If you try to add an email as a tag from the list of tags (in the Edit Tags screen), it shows an error

| Before      | After |
| ----------- | ----------- |
|       |        |

- If you try to edit a tag as an email from the list of tags (in the Edit Tags screen), it shows an error

| Before      | After |
| ----------- | ----------- |
|       |        |

### Test

1. Go to a note
2. Tap on the tags field at the bottom and add a couple of tags
3. Tap on (...) button at the top right corner and tap on Collaborators
4. Add a couple of collaborators
5. Go back to edit the note
6. Go back to the list of notes
7. Open the left panel. Just the tags and not the emails should be shown
8. Tap on one of the tags and the list of notes should be filtered by the notes that have that tag.
9. Tap on All Note and the list of notes should be filtered to all notes
8. Tap on `EDIT` button on the left panel
9. Just the tags and not the emails should be shown
10. Try to add a tag
11. Try to add an email on the `Add Tag` form. It should show an error
12. Go back and dismiss the dialog
13. Tap on one of the tags on the list
14. Try to edit the tag and write it as an email on the `Edit Tag` form. It should show an error
12. Go back and dismiss the dialog

### Review

### Release

`RELEASE-NOTES.txt` was updated in  aa5c545f2c1bd35deebdb2c56cb8154c2cb5bf93 with:
> [Internal] Updated list of tags to filter out collaborators (emails)

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -680,7 +680,6 @@ public class NotesActivity extends ThemedAppCompatActivity implements NoteListFr
         }
 
         mTagsAdapter.submitList(tags);
-        //mTagsAdapter.changeCursor(tagCursor);
         mNavigationMenu.removeGroup(GROUP_SECONDARY);
         mNavigationMenu.removeGroup(GROUP_TERTIARY);
 
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -654,10 +654,6 @@ public class NotesActivity extends ThemedAppCompatActivity implements NoteListFr
 
     private List<Tag> getTagsFromCursor(Bucket.ObjectCursor<Tag> tagCursor) {
         List<Tag> tags = new ArrayList<>();
-        if (tagCursor.getCount() == 0) {
-            return tags;
-        }
-
 
         for (int i = 0; i < tagCursor.getCount(); i++) {
             tagCursor.moveToNext();
--- a/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
@@ -1,5 +1,34 @@
 package com.automattic.simplenote;
 
+import static com.automattic.simplenote.NoteListFragment.TAG_PREFIX;
+import static com.automattic.simplenote.analytics.AnalyticsTracker.CATEGORY_NOTE;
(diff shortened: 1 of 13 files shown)


Simplenote/src/main/java/com/automattic/simplenote/NotesActivity.java
```diff
@@ -648,16 +652,30 @@ private void checkForSharedContent(Intent intent) {
         }
     }
 
+    private List<Tag> getTagsFromCursor(Bucket.ObjectCursor<Tag> tagCursor) {
+        List<Tag> tags = new ArrayList<>();
+
+        for (int i = 0; i < tagCursor.getCount(); i++) {
+            tagCursor.moveToNext();
+            Tag tag = tagCursor.getObject();
+            if (!collaboratorsRepository.isValidCollaborator(tag.getName())) {
```
nbradbury, line 661: I'm curious about this. Since this is really just a check for a valid email address, then if a user previously used an email address as a tag it won't show up any more. Is that anything to be concerned about?

danilo04: If the tag (email) is assigned to a note, it can be accessed from the collaborators' UI. If it is not assigned to a note, then the user would not be able to access it. The concern here would be that there are some tags (emails) that should be deleted but there won't be a way to delete them. Emails will be accessible just from the collaborators' UI for each note.

nbradbury (approved):
Changes look good, we can :shipit: after design review.

SylvesterWilmott:
Thanks @danilo04 

This looks good. Regarding the tags validation error. Let's just update the copy to `Tags cannot be emails`.

danilo04:
Hi @SylvesterWilmott, thanks for the review

>  Let's just update the copy to Tags cannot be emails.

Fixed in 4eeec8e31bb0ba1d51bfa3503ec2ce8c1fea6efd:

image

SylvesterWilmott:
:shipit:

SylvesterWilmott (approved):

PR #1569, opened 2022-11-14
Add cross platform IAP status sync and tracking.

khaykov:
This PR adds some tracking and a subscription status sync using Symperium. We are using existing Preferences bucket, to which we added subscription related fields.

Simperium has a totaly different mindset when it comes to networking, so I'm not sure I utilized it in the best possible way, but it works so far.

Before testing:

- Make sure you are logged in to the latest version of Play Store.
- Make sure you are using release version of the app (or the debug version without .debug preffix).
- Let me know what account you will be using for purchase testing so I can add it in Play Store. This will allow you to use test cards.
- Subscriptions renew every five minutes. You can cancel them from Subscription screen of Play Store.

Since we are using Simperium bucket to determine visibility of IAP banner the testing steps are mostly the same:
- Test purchase flow from sidebar with working test card. Confirm that after purchase is successful, "Thank you!" snackbar appears, Sustainer banner is not visible in sidebar, and black "Thank you" banner is visible in settings.
- Test purchase flow from Settings screen with working test card. Confirm that after purchase is successful, "Thank you!" snackbar appears, Sustainer banner is replaced with "Thank you" banner in Settings screen and it is not visible in Sidebar.
- Confirm that the subscription and banner visibility status remains the same after app is reinstalled.
- Cancel Subscription in Play store and confirm that the IAP banner is visible in the sidebar and Settings screen.

I completed the cross-platform testing part with Jorge and confirmed that it works as expected.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/viewmodels/IapViewModel.kt
+++ b/Simplenote/src/main/java/com/automattic/simplenote/viewmodels/IapViewModel.kt
@@ -128,7 +128,7 @@ class IapViewModel(application: Application) :
                     val preferences = preferencesBucket.get(Preferences.PREFERENCES_OBJECT_KEY)
                     preferences.removeActiveSubscription()
                 }
-
+                updateIapBannerVisibility()
             } else {
                 Log.e(TAG, billingResult.debugMessage)
             }
--- a/Simplenote/src/main/java/com/automattic/simplenote/viewmodels/IapViewModel.kt
+++ b/Simplenote/src/main/java/com/automattic/simplenote/viewmodels/IapViewModel.kt
@@ -289,10 +289,10 @@ class IapViewModel(application: Application) :
     }
 
     private fun doesNotHaveSubscriptionOnOtherPlatforms(): Boolean {
-        val preference = preferencesBucket.get(Preferences.PREFERENCES_OBJECT_KEY)
+        val preferences = preferencesBucket.get(Preferences.PREFERENCES_OBJECT_KEY)
 
-        preference?.let {
-            val currentSubscriptionPlatform = preference.currentSubscriptionPlatform
+        preferences?.let {
+            val currentSubscriptionPlatform = preferences.currentSubscriptionPlatform
 
             return currentSubscriptionPlatform == null
                     || currentSubscriptionPlatform == Preferences.SubscriptionPlatform.ANDROID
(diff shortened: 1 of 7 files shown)


Simplenote/src/main/java/com/automattic/simplenote/viewmodels/IapViewModel.kt
```diff
@@ -33,13 +39,18 @@ class IapViewModel(application: Application) :
     private val _snackbarMessage = SingleLiveEvent<IapSnackbarMessage>()
     val snackbarMessage: LiveData<IapSnackbarMessage> = _snackbarMessage
 
+    private val preferencesBucket = getApplication<Simplenote>().preferencesBucket
```
danilo04, line 42: (🔧 ) I think you can use DI to get the reference for the preference bucket. I normally put buckets behind a repository but I think it is OK if we use it here in the ViewModel. Just modify the constructor with the preferences bucket and add the `@Provides` method. in `DataModule.kt`:

```kotlin
abstract class DataModule {
    companion object {
        @Provides
        fun providesTagsBucket(simplenote: Simplenote): Bucket<Tag> = simplenote.tagsBucket

        @Provides
        fun providesNotesBucket(simplenote: Simplenote): Bucket<Note> = simplenote.notesBucket

        @Provides
        fun providesPreferencesBucket(simplenote: Simplenote): Bucket<Preferences> = simplenote.preferencesBucket

        @Provides
        fun providesSimperium(simplenote: Simplenote): Simperium = simplenote.simperium
    }
```

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/IapViewModel.kt
```diff
@@ -249,4 +287,43 @@ class IapViewModel(application: Application) :
...
+        val preferences = preferencesBucket.get(Preferences.PREFERENCES_OBJECT_KEY)
+
+        preferences?.let {
+            val currentSubscriptionPlatform = preferences.currentSubscriptionPlatform
+
+            return currentSubscriptionPlatform == null
+                    || currentSubscriptionPlatform == Preferences.SubscriptionPlatform.ANDROID
+        }
+        return false
+    }
```
danilo04, line 301: (♻️ ) This is a little confusing. If the preference object doesn't exist, it should return `true` since it does not have a subscription on another platform. However, I know that case is not possible since `preferences.get(...)` would throw an exception. So it doesn't matter, just pointing out this.

danilo04 (approved):
Hi @khaykov. I tested it on different devices with different google accounts and it worked correctly. I commented on things that can be improved but it is not a big deal.

danilo04:
I am going to merge this PR since it is working correctly and has just a few things to improve, but they are not blocking.

PR #1571, opened 2022-11-18
Upgrade  tagetSdkVersion to Android 31

danilo04:
Fixes ##1570.

### Fix
Upgrade to `targetSdkVersion = 31` and `compileSdkVersion = 31`. Following the [recommendations](https://developer.android.com/google/play/requirements/target-sdk) for the upgrade, we were affected by the following:

- Intent filters: If your app contains [activities](https://developer.android.com/guide/components/activities/intro-activities), [services](https://developer.android.com/guide/components/services), or [broadcast receivers](https://developer.android.com/guide/components/broadcasts) that use [intent filters](https://developer.android.com/guide/components/intents-filters#Receiving), you must explicitly declare the [android:exported](https://developer.android.com/guide/topics/manifest/activity-element#exported) attribute for these components.
- [Pending intent mutability](https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability): You must specify the mutability of each PendingIntent object that your app creates.

We apply the following changes to support the new `targetSdkVersion`:

- Upgrade `androidx.work:work-runtime` to `2.7.1`. The previous version was crashing the app because immutability of pending intents. 
- Ad immutable flag to pending intents. In one case (widgets) we need mutable pending intents. 
- Added `exported:true` to components that use intent filters.

### Test

Smoke test the whole app. 

- Test notes, adding, editing, opening links, etc.
- Add collaborators
- Add tags
- Setup lock screen
- Import/Export
- Widgets

### Review

### Release
`RELEASE-NOTES.txt` was updated in a7465aeee3a84990e006b6038b633b28acf1728e with:
> Upgrade to targetSdkVersion 31 (Android 31)

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 2.24.2
 -----
 * Added ability to purchase a Simplenote Sustainer subscription [#1567](https://github.com/Automattic/simplenote-android/pull/1567)
+* [Internal] Upgrade to targetSdkVersion 31 (Android 31) [#1571](https://github.com/Automattic/simplenote-android/pull/1571)
 
 2.24.1
 -----
--- a/Simplenote/build.gradle
+++ b/Simplenote/build.gradle
@@ -37,7 +37,7 @@ android {
         }
         versionCode 150
         minSdkVersion 23
-        targetSdkVersion 30
+        targetSdkVersion 31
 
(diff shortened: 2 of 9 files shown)


khaykov (approved):
Thanks for taking care of this, @danilo04 ! Smoke tested all kind of things with release build, and everything works as expected 👍

matthewejohnson711:
sounds good

PR #1603, opened 2023-08-08
Improve IAP handling

khaykov:
This PR fixes some ANR's that result from billing service. Despite following Google's guidelines, we are getting ANR's when initializing billing service, so I moved some calls to background thread, and added a backoff logic.

Backoff logic:
- Comment out [this](https://github.com/Automattic/simplenote-android/compare/issue/fix-iap-anr?expand=1#diff-02da68460a4936efcb23323f8b23eefb45c739341570425c672e05ae67af3362R104) line.
- Add   debuggable true to the release config. Should look like this:
```
release {
            debuggable true
            minifyEnabled true
            shrinkResources false
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
        }
```
- Launch app in release configuration.
- In the logcat, confirm that you see `IapViewModel I  Billing response OK`
- Open device settings, navigate to Apps, and find Play Store.
- Clear Play Store storage.
- Check logcat, and confirm that you see `IapViewModel I  Retrying billing service connection after 1000 MS delay`
- Do this again, and confirm that same message with 2000 MS delay.
- etc. 

Afte this:

- Confirm that billing flow works by purchasing sustainer plan.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,7 @@
 2.28
 -----
+* Fixed occasional crash on Samsung devices [#1601](https://github.com/Automattic/simplenote-android/pull/1601)
+* Fixed recurring ANR issues [#1603](https://github.com/Automattic/simplenote-android/pull/1603)
 
 
 2.27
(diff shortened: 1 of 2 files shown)


danilo04 (approved):
Hi @khaykov, this is working as expected. Great job here!!

PR #1605, opened 2023-08-12
Fix crash when visiting preference activity

khaykov:
This PR fixes the crash that happens when you open Preferences.

In recent IAP changes, I modified IAP view model, and dod not add a required `@AndroidEntryPoint` annotation to `PreferencesActivity`, which caused it to crash.

This PR addresses this.

To test, navigate to "Settings" from the nav drawer and confirm that you can see Settings activity without a crash.


danilo04 (approved):
I think this is required since some people is having issues.

PR #1607, opened 2023-08-25
Upgraded targetSdkVersion from 31 to 33

shivank-dvlpr:
### Problem
According to new Google Play target API level requirement new apps must target Android 13 API level 33 or above except for wear OS 

https://developer.android.com/google/play/requirements/target-sdk

### Fix
Changed compileSdkVersion and targetSdkVersion from 31 to 33.


danilo04:
The changes to use SDK 33 were merged in #1612. Thank you for the changes.

PR #1613, opened 2023-10-11
Update Simperium to version `v1.3.0`

spencertransier:
This PR updates Simperium from a hash to the `v1.3.0` release. There is no functional change here. The  `45b683bc8445340222df4c80136f7b9fa05d1a89` hash was the latest commit before the `v1.3.0` release was created.


danilo04 (approved):
LGTM :shipit: !

PR #1616, opened 2023-10-26
Update release toolkit to 9.1.0

crazytonyli:
The main motivation of this change is migrating to Ruby 3 which isn't compatible with older versions of release toolkit.

There aren't many android related breaking changes in [the migration guide](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md). The only one I can find is calling `push_to_git_remote` before triggering a new build. However, I didn't find any place that triggers new builds in the Fastfile. So, I did make any changes to it.

From [the 7.0.0 to 8.0.0 migration guide](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md#from-700-to-800)
- [ ] Call `push_to_git_remote` in order to trigger new builds with latest changes.
**Update:** ⚠️ I didn't make ⬆️ change because I don't see any action in Fastfile which triggers new CI builds.

From [the 8.0.0 to 9.0.0 migration guide](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/MIGRATION.md#from-800-to-900)
- [x] `ios_get_app_version` is updated in this PR to pass the new config file parameter.
- [x] `ios_localize_project` is not used
- [x] `ios_update_metadata` is not used
- [x] This project does not use `Deliverfile`.
- [x] This project does not call `get_prs_list`.

### Test

I'm not familiar enough with the release-toolkit nor the app's release process to know which fastlane lanes are safe to run locally and repeatedly for debugging or testing purpose.


Gemfile
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-source "https://rubygems.org"
+source 'https://rubygems.org'
 gem 'fastlane', '~> 2'
-gem 'fastlane-plugin-wpmreleasetoolkit', '~> 7.0'
+gem 'fastlane-plugin-wpmreleasetoolkit', '~> 9.0'
```
mokagio: What do you think of `~> 9.1` give that the version resolved now?

mokagio (commented):
>  Call push_to_git_remote in order to trigger new builds with latest changes. ⚠️

Is the warning sign emoji in the description a note to self or is there something we should be aware of?

crazytonyli:
@mokagio Sorry, I missed that! I think I've updated the PR description.

Gemfile
crazytonyli: Addressed in https://github.com/Automattic/simplenote-android/pull/1616/commits/305de22d93d56b6a72e36ebd5ace724c238ab37a

mokagio (approved):

PR #1627, opened 2024-01-19
Update to use SDK 34, and updated gradle build tools

roundhill:
### Fix
This PR updates the build and target SDK to 34 and updates the gradle plugin to the latest version. 

I was hoping to also update more of the dependencies, but as soon as I touched them things went crazy, especially with the `appcompat` libs. We can update those in future PRs piece by piece.

### Test
1. Clean, build and run the app. Hopefully it works!
2. Try building the Watch app as well, it should compile. Note that it won't launch directly to a watch emulator because it only listens for google voice intents.


danilo04 (approved):
Hi @roundhill, thank you for upgrading the app to SDK 34. Everything is working as expected. I tested the sustainer flow and is working as well.

PR #1632, opened 2024-02-02
Merge changes from finalized 2.31 build

mokagio:
- [x] `versionName` dropping the `rc` suffix and `versionCode` bump in `build.gradle`
- [x] `strings.xml` updates (none were expected given there have been no changes in the source, but we got one that I can only assume was a late arrival or a community tweak)
- [x] `release_notes.xml` and `changelogs/*.txt` updates

@spencertransier @oguzkocer thank you in advance for reviewing this, you were suggested by GitHub 😅


fastlane/metadata/android/release_notes.xml
```diff
@@ -1,4 +1,85 @@
...
+*修復了應用程式啟動時因筆記量過大而沒有反應的問題。
+
+</zh-TW>
 <en-US>
-2.30:
-• Fixed a crash when syncing large notes
+2.31:
+* Fixed the app from becoming unresponsive at launch with a large amount of notes.
+
 </en-US>
```
mokagio, line 85: Out of scope given the state of this project, but I wonder whether we should add this to the `.gitignore` given it's build from the individual changelogs 🤔

spencertransier: We can probably remove this file and the automation that creates it once we use `upload_to_play_store` to automatically upload the release notes to the Play Store tracks. 

See the "Changelogs (What's new)" section in https://docs.fastlane.tools/actions/upload_to_play_store

spencertransier (approved):
Looks great!

fastlane/metadata/android/release_notes.xml
mokagio: Great idea. Now... how do we remember to do this on the next release given this project is in maintenance mode 🤔

mokagio: Maybe we should leave a note in the release scenario to implement this change before starting the next code freeze.

mokagio: I opened a PR for that in one of our internal repos.

PR #1645, opened 2024-04-04
Add custom icon preference for sustainers

roundhill:
### Fix
This PR adds a new preference in the settings for sustainers to set a green app icon:

Changing the icon is kind of a pain on Android involving `activity-alias` to use a different drawable for the (gulp) main activity. From my testing it is working pretty well but does crash the app when you swap out the activities. 

I also couldn't get the "New Note" shortcut to work, so just left that out when switching to the activity-alias.

### Test
* On a non-sustainer account, launch the app. You should see no sustainer icon setting in the preferences view and the app should work normally.
* Switch to a sustainer account, you can run this code to make yourself a sustainer:

```
long now = System.currentTimeMillis();
try {
    Preferences preferences = mPreferencesBucket.get(Preferences.PREFERENCES_OBJECT_KEY);
    preferences.setActiveSubscription(now);
    throw new RuntimeException(e);
}
```

* The sustainer icon preference should be in the settings.
* Flip the switch! The app should close and you'll see the green icon.
* Verify tapping the icon works and that you can still share content from other apps to Simplenote.

### Release


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please add tests for class <code>IntentUtils</code> (or add <code>unit-tests-exemption</code> label to ignore this).</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR is not assigned to a milestone.</td>
    </tr>
  </tbody>
</table>

</p>

khaykov:
Thanks, @roundhill ! It mostly works well, but I found a couple of issues :)
- After you toggle the switch, app does not restart automatically on devices with API 26 (and maybe other).
- After switching to sustainer icon, if you open a note from widget and press Up Arrow, the app closes instead of navigating to main activity.

Simplenote/src/main/java/com/automattic/simplenote/utils/IntentUtils.java
```diff
@@ -0,0 +1,27 @@
...
+        return "com.simplenote.automattic.NotesActivity";
+    }
+
+    public static Intent maybeAliasedIntent(Context context) {
+        String packageName = context.getPackageName();
+
+        Intent intent = new Intent();
+        intent.setComponent(new ComponentName(packageName, getMainActivityClassName(context)));
+
+        return intent;
```
roundhill, line 25: I added this function to get the proper main activity, if it is the alias or not. This fixes the back button crash and probably some other scenarios where we were referring directly to the `NotesActivity` class which may or may not exist depending on if the activity alias for the sustainer icon is active.

roundhill:
> * After you toggle the switch, app does not restart automatically on devices with API 26 (and maybe other).

I think it does do it but it takes a while (at least on the emulator it does). I've updated the requirement to show the switch to API 28 because I was getting an exception that I couldn't track down: `Unexpected null component name or activity info: ComponentInfo{com.automattic.simplenote.debug/com.automattic.simplenote.NotesActivity}, null`. The exception went away on API 28 and above.

> * After switching to sustainer icon, if you open a note from widget and press Up Arrow, the app closes instead of navigating to main activity.

This should be fixed now.

khaykov (approved):
Thanks for the fixes, @roundhill ! Works great now 👍

PR #1660, opened 2024-06-26
[Authentication] Add logic to login via magic link

notandyvee:
### Fix

This PR adds the logic necessary to login via magic link. Instead of entering a password for login, you request an email. You go to the your email and click the login link in order to auto-magically login without password.

### Test

TBD

### Review

Nothing special was added. I avoided a major overhaul. For example, instead of json parsing with a library I used `JSONObject` and used `OkHttp`. See `SimpleHttp`. If we do more with network requests and json parsing it would be worth using retrofit and a better json parsing library, like Moshi. 

### Release

This change requires release notes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>OkHttpMagicLinkRepository</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>SimpleHttp</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.kt
```diff
@@ -0,0 +1,58 @@
...
+        super.onCreate(savedInstanceState)
+        val uri = intent.data
+        when (uri?.host) {
+            AUTHENTICATION_SCHEME -> {
+                val intent = Intent(this, RedirectUriReceiverActivity::class.java)
+                intent.setData(uri)
+                startActivity(intent)
+            }
+            LOGIN_SCHEME -> {
+                if (queryParamContainsData(uri.query, "auth_key") && queryParamContainsData(uri.query, "auth_code")) {
```
danilo04: (⛏️ ) Move `"auth_key"` and `"auth_code"` as constants.

Simplenote/src/main/java/com/automattic/simplenote/authentication/MagicLinkableFragment.kt
```diff
@@ -0,0 +1,105 @@
...
+import androidx.fragment.app.DialogFragment
+import androidx.fragment.app.Fragment
+import com.automattic.simplenote.R
+import com.google.android.material.textfield.TextInputLayout
+import com.simperium.android.ProgressDialogFragment
+
+/**
+ * Base class used to share logic between sign up and login, specifically related to magic links.
+ */
+abstract class MagicLinkableFragment : Fragment() {
```
danilo04, line 22: Are you thinking on implementing more fragments related to magic links?

notandyvee: Sorta. This class was based off of `SignupFragment`, which is similar. Both Sign Up and Sign In with magic link are similar. So it made sense to use that logic and share it between them. Additionally, we can likely remove the okhttp use from `SignupFragment` and move it over the business logic to a viewmodel. Hence why I created the `SimpleHttp` helper object. But I didn't want to inflate the work here further.

Simplenote/src/main/java/com/automattic/simplenote/networking/SimpleHttp.kt
```diff
@@ -0,0 +1,57 @@
...
+import okhttp3.MediaType
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody
+import okhttp3.Response
+import org.json.JSONException
+import org.json.JSONObject
+import java.util.concurrent.TimeUnit
+
+object SimpleHttp {
```
notandyvee: Just a heads up. While trying to add the unit tests I began to run into problems. Firstly, mocking static methods is annoying due to how behind we are with library updates. So I will have to make it a regular class. I will move `OkHttp` into a singleton instance using hilt so it can be injected into an `SimpleHttp` instance. Then we'll also inject `SimpleHttp`. This will make unit testing easier. I'll do it monday morning.

There is the other issue of passing `Simplenote`, which I cannot mock with the current version of mockito. I will try and also inject that in somehow through an interface. I have some thoughts. Either way, with the exception of those gotchas feel free to review the rest of it, which I will likely not change.

notandyvee:
Ok @danilo04 , ready for review. Sorry it took so long. I had trouble with unit tests. I attempted to migrate to flows as a result, but I don't feel comfortable using outdated libraries. I tried updating a few libs and it caused a cascading effect of updating way too many things. We can save it for another time. Just wanted to be clear why I didn't use flows.

Additionally, in order to test this end-to-end, you must build it in release mode with the right creds. DM me if you need help. Won't work in debug at the moment due to the creds and debug package name.

Also, do I need to do anything special with strings? It's giving me a warning. But I'm pretty sure I only edited the base `strings.xml`.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/RequestMagicLinkViewModel.kt
```diff
@@ -0,0 +1,49 @@
...
+@HiltViewModel
+class RequestMagicLinkViewModel @Inject constructor(
+    private val simpleHttp: SimpleHttp,
+    @Named(IO_THREAD) private val ioDispatcher: CoroutineDispatcher,
+) : ViewModel() {
+    private val _magicLinkRequestUiState = MutableLiveData<MagicLinkRequestUiState>()
+    val magicLinkRequestUiState: LiveData<MagicLinkRequestUiState> get() = _magicLinkRequestUiState
+
+    fun requestLogin(username: String) = viewModelScope.launch(ioDispatcher) {
+        _magicLinkRequestUiState.postValue(MagicLinkRequestUiState.Loading(message = "Requesting Magic Link"))
```
roundhill: Can you make this localizable, and maybe say `Requesting log in link` instead?

notandyvee: Yes! I guess I will do the same for the others. There are a few of those 😅 . Thanks for pointing it out.

notandyvee: Done

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+        val authKey = uri?.getQueryParameter(AUTH_KEY_QUERY)
+        val authCode = uri?.getQueryParameter(AUTH_CODE_QUERY)
+        if (!authKey.isNullOrBlank() && !authCode.isNullOrBlank()) {
+            val intent = Intent(this, SimplenoteAuthenticationActivity::class.java)
+            intent.putExtra(SimplenoteAuthenticationActivity.KEY_IS_MAGIC_LINK, true);
+            intent.putExtra(SimplenoteAuthenticationActivity.KEY_MAGIC_LINK_AUTH_KEY, authKey)
+            intent.putExtra(SimplenoteAuthenticationActivity.KEY_MAGIC_LINK_AUTH_CODE, authCode)
+            intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK or Intent.FLAG_ACTIVITY_NEW_TASK)
+            startActivity(intent)
+        }
```
danilo04, line 73: I think we can add a Toast message here for the else case, saying that the link was invalid. Just use the app context to launch the Toast.

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.automattic.simplenote.R
+import com.automattic.simplenote.utils.NetworkUtils
+import com.automattic.simplenote.viewmodels.MagicLinkRequestUiState
+import com.automattic.simplenote.viewmodels.RequestMagicLinkViewModel
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
+class SignInFragment: MagicLinkableFragment() {
+
+    var viewModel: RequestMagicLinkViewModel? = null
```
danilo04: You can use the property delegate to build the view model:

```suggestion
    val viewModel: RequestMagicLinkViewModel by viewModels()
```

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
```diff
@@ -0,0 +1,65 @@
...
+                intent.putExtra("EXTRA_IS_LOGIN", true)
+                this.startActivity(intent)
+                act.finish()
+            }
+        }
+        return view
+    }
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        viewModel = ViewModelProvider(this).get(RequestMagicLinkViewModel::class.java)
```
danilo04: This can be deleted if you use the property delegate `by viewModels()` (comment above).

Simplenote/src/main/java/com/automattic/simplenote/networking/SimpleHttp.kt
```diff
@@ -0,0 +1,50 @@
...
+import okhttp3.Request
+import okhttp3.RequestBody
+import okhttp3.Response
+import org.json.JSONException
+import org.json.JSONObject
+import javax.inject.Inject
+
+private const val HTTP_SCHEME = "https"
+private const val BASE_URL = "app.simplenote.com"
+open class SimpleHttp @Inject constructor(val client: OkHttpClient) {
```
danilo04, line 15: Consider refactoring `AccountNetworkUtils` by maybe using this class instead of the OkHtttp service it creates. It can be handled in a future PR.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/CompleteMagicLinkViewModel.kt
```diff
@@ -0,0 +1,62 @@
...
+            ).use { response ->
+                if (response.isSuccessful) {
+                    _magicLinkUiState.postValue(MagicLinkUiState.Success)
+                    val body = response.body()?.string()
+                    val json = JSONObject(body ?: "")
+                    val syncToken = json.getString("sync_token")
+                    val username = json.getString("username")
+                    simplenote.loginWithToken(username, syncToken)
+                } else if (response.code() == 400) {
+                    _magicLinkUiState.postValue(MagicLinkUiState.Error(messageRes = R.string.magic_link_complete_login_error_message))
```
danilo04: (❤️ ) Cool

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/CompleteMagicLinkViewModel.kt
```diff
@@ -0,0 +1,62 @@
...
+    private val _magicLinkUiState = MutableLiveData<MagicLinkUiState>()
+    val magicLinkUiState: LiveData<MagicLinkUiState> get() = _magicLinkUiState
+
+    fun completeLogin(authKey: String, authCode: String) = viewModelScope.launch(ioDispatcher) {
+        _magicLinkUiState.postValue(MagicLinkUiState.Loading(messageRes = R.string.magic_link_complete_login_loading_message))
+        try {
+            simpleHttp.firePostRequest(
+                "account/complete-login",
+                mapOf(Pair("auth_key", authKey), Pair("auth_code", authCode))
+            ).use { response ->
```
danilo04: Consider using a repository for this kind of call. Maybe you can create a `MagicLinkRepository` to handle the calls and then inject it into the view models that need it.

Simplenote/src/main/java/com/automattic/simplenote/viewmodels/RequestMagicLinkViewModel.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val _magicLinkRequestUiState = MutableLiveData<MagicLinkRequestUiState>()
+    val magicLinkRequestUiState: LiveData<MagicLinkRequestUiState> get() = _magicLinkRequestUiState
+
+    fun requestLogin(username: String) = viewModelScope.launch(ioDispatcher) {
+        _magicLinkRequestUiState.postValue(MagicLinkRequestUiState.Loading(
+            messageRes = R.string.magic_link_request_login_loading_message)
+        )
+        try {
+            // TODO: Uncomment the request_source once finalized.
+            simpleHttp.firePostRequest(
```
danilo04: Same comment as above, consider moving this call into a repository.

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/CompleteMagicLinkViewModelTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Rule
+
+const val AUTH_KEY_TEST = "auth_key_test"
+const val AUTH_CODE_TEST = "auth_code_test"
+
+@ExperimentalCoroutinesApi
+class CompleteMagicLinkViewModelTest {
```
danilo04, line 21: (❤️ ) Nice!

danilo04 (changes requested):
Hi @notandyvee, great job on the changes. The magic link is working as expected. However, it looks like it messed up the login with password. Check the following video:


https://github.com/Automattic/simplenote-android/assets/195721/4cfd8f52-60d5-485b-a362-dcad84d5906e

Also, if I try to open a magic link and the user is already logged in, the user is logged out. We need to handle this edge case. 

I left a couple of code comments for your consideration as well.

Simplenote/src/main/java/com/automattic/simplenote/networking/SimpleHttp.kt
notandyvee: Oh hey. I guess you had a similar idea 😅 . Ya. Definitely future PR.

notandyvee:
@danilo04 Ready for re-review. I will not add unit tests for the 2 classes the build was complaining about since it would essentially be testing library code. So added the unit test exemption. Appreciate the thorough review.

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.kt
danilo04: And take the user to the notes activity.

notandyvee: Assuming no other big issues, I can make an update to this on a subsequent PR if that's ok.

danilo04: Yeah, this is a pretty edge case, but we cannot be sure if the email client will mess up the link or something like that. So if the link is wrong, then we need to send the user to the login screen if it is not logged in.

danilo04 (approved):
Hi @notandyvee, thank you for the changes. The PR is looking good and I tested the scenarios that were failing and now they are working. Good job on this :shipit: 

We can handle the edge case discussed in the code comments in a different PR.

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.kt
notandyvee: > So if the link is wrong, then we need to send the user to the login screen if it is not logged in.

Oh. I see what you mean 🤔 . Fair enough. Created the issue.

PR #1663, opened 2024-07-03
[Analytics] Add trackers for magic link sign in.

notandyvee:
### Fix

This PR adds two analytics events for magic link: `user_requested_login_link` and `user_confirmed_login_link`. I added a new `Stat` for this to make it clearer.

### Test

- [ ] 🚬 test magic link and successfully log in.
- [ ] I assume tracks shows events in logs?

### Review

@roundhill since I believe you added analytics in the past. Let me know if the new `Stat` is necessary. I did it for clarity, but none of the other ones seemed to fit.


roundhill (changes requested):
I think we need to create a stat _per_ event since it is the unique identifier for the tracks event.

So instead of `AUTH_MAGIC_LINK` in `Stat`, we need to have `USER_REQUESTED_LOGIN_LINK` and `USER_CONFIRMED_LOGIN_LINK` and pass those in as the first argument to the track() call. I believe the label is a bit arbitrary but can identify which part of the UI triggered the stat.

roundhill (approved):
Changes look good, but what's up with the CI build?

PR #1665, opened 2024-07-11
[Auth] Ability to add Passkeys to your account.

notandyvee:
### Fix

This PR is the 2nd step in this 3-4 part series in adding passkeys to Simplenote.

This step allows you to add a passkey for your account in order to be able to authenticate with something.

### Test

- [ ] Build with release as it's required for security.
- [ ] Go to settings
- [ ] Scroll down to the "Account" section and click "Add Passkey Authentication".
- [ ] Enter the password associated with your account.
- [ ] Ensure there is some loading.
- [ ] Once we receive the challenge to generate a passkey, the `CredentialManager` will pop up a bottomsheet to associate a passkey with your account. Select the device you are using. Using other methods has not been tested.
- [ ] Assuming it worked, you should see one last loading dialog and then a toast with the text: "Successfully added passkey".
- [ ] DM me to confirm the account you used was actually saved on the backend.

### Review

Only thing I'd like to note on this PR is `PasskeyManager` was created in order extract the logic out of the Fragment and into kotlin land. I didn't put it in the viewmodel because the credential manager needs a context, which could expose a memory leak. Your thoughts on a better place is welcome.

### Release

This PR requires updates to release notes.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>OkHttpPasskeyRepository</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/src/main/res/layout/passkey_add_dialog.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<EditText
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/add_passkey_textfield"
+    android:layout_width="match_parent"
```
danilo04, line 5: Add password input type

```xml
android:inputType="textPassword"
```

Simplenote/src/main/java/com/automattic/simplenote/authentication/passkey/OkHttpPasskeyRepository.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.automattic.simplenote.repositories.PasskeyRepository
+import com.automattic.simplenote.repositories.PasskeyResponseResult
+import org.json.JSONObject
+import java.io.IOException
+import javax.inject.Inject
+
+const val EMAIL_KEY = "email"
+const val PASSWORD_KEY = "password"
+const val WEB_AUTHN_KEY = "webauthn"
+
```
danilo04, line 16: (⛏️ ) Maybe make this `@Singleton`?

Simplenote/src/main/java/com/automattic/simplenote/authentication/passkey/PasskeyManager.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.automattic.simplenote.viewmodels.PasskeyViewModel
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+const val TAG = "PasskeyManager"
+
+// Helper class to extract the credential manager work. Also, caller is in Java, and we need kotlin.
+object PasskeyManager {
```
danilo04, line 18: (⛏️ ) You can make this an object and inject the dispatchers to make the class more testable. It is a nitpick, so you can disregard this comment.

danilo04 (commented):
Hi @notandyvee, good job on implementing this feature. As we discussed in chat, there is an issue with Android 14. I left a couple of nitpick comments for your consideration.

Simplenote/src/main/java/com/automattic/simplenote/authentication/passkey/PasskeyManager.kt
notandyvee: I am thinking about this. For now I will not do that. Part of the reason is because I wouldn't unit test this class, hence why I have static methods. It heavily uses 3rd party SDK, which has very little value to unit test. I'd be mocking everything anyways. That being said, I will pass in the dispatcher instead of directly doing `Dispatcher.IO`.

notandyvee:
Hey @khaykov if you can fill in for Danilo 🙏🏻

khaykov:
@notandyvee It works! I was able to get to "Successfully added passkey". 

I have a couple of usability comments:

- I have no idea what the "Passkey" does and why should I use it. Can we have some sort of explanation on what it is? Maybe combined with password dialog?

- If you do not have a screen lock enabled, you will have to setup it up first. After you are done, I would assume that passkey will be enabled, but it is not - I have to go back to the app and restart the process. Wonder is it possible to resume process automatically? Or let user know that passkey is not set up?

- After I add a passkey, the menu item still says "Add Passkey Authentication". 

- Error message when entering wrong password is not very clear for user. I believe it should be something simple like "Wrong credentials", etc. Part about "challenge" will not make much sense to regular person.
f9513ba5c5b6f654536130f37634a66a

- I see an error when dismissing passkey bottom sheet, the message is very generic and does not represent what actually happens.
580c7f3cffbffe558f9fe808946eba44

notandyvee:
@khaykov I think all your points are valid. But I think we should address all of them in a separate PR. Most of them require discussions. For example, there is no way to know when a passkey was added on the backend. We could check the local passkeys. But what if the user added the passkey on web? Should mobile not show that? For now I think it's out of scope for the PR.

The exception is

> I see an error when dismissing passkey bottom sheet, the message is very generic and does not represent what actually happens.

That's sort of a legitimate issue. For some reason the library throws an exception when that happens 😅 . I can update the event to say user cancelled so there is no error.

notandyvee:
Done @khaykov . Any other improvements we can add in a future PR. You shouldn't see an error anymore. For the record, CredentialManager is not properly documented. Apparently they use [these errors](https://webidl.spec.whatwg.org/#idl-DOMException-error-names). Not very descriptive for our specific issue. But for now this helps.

khaykov (approved):
Sure thing, feel free to address rest of the suggestions in separate PR's , if required :) This one looks good 👍

PR #1666, opened 2024-07-11
[Bug] Entering email in magic link screen shows it in login with pass.

notandyvee:
### Fix

This PR is a small enhancement. While testing I realized that if a user were to enter their email during the magic link process, but then decide to enter their password, clicking the button clears the email field.

This PR fixes that so it carries over the info to make the users life a bit easier.

### Test

- [ ] On the screen where it says "Instantly Log In With Email", enter your email.
- [ ] Click login with password.
- [ ] The text or email you had in the email field will appear in the email field on the login with password screen.

### Review

No need to login while testing this.

### Release

No relese update required.


Simplenote/src/main/java/com/automattic/simplenote/authentication/SimplenoteCredentialsActivity.java
```diff
@@ -83,8 +84,12 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
             EditText inputPassword = ((TextInputLayout) findViewById(R.id.input_password)).getEditText();
 
             if (inputEmail != null && inputPassword != null) {
-                inputEmail.setText(BuildConfig.LOGIN_EMAIL);
-                inputPassword.setText(BuildConfig.LOGIN_PASSWORD);
+                if (!TextUtils.isEmpty(BuildConfig.LOGIN_EMAIL)) {
```
notandyvee, line 87: I added this because during development it clears things out leading to confusion. Presumably we meant this value to only set if it had something.

roundhill (approved):
Works great and LGTM

PR #1668, opened 2024-07-17
[Auth] Update magic links for the new complete login update.

notandyvee:
### Fix

This PR makes some updates related to the magic link flow.

* Clicking login now doesn't show a bottom sheet, instead takes you directly to a login screen to enter your email.
* Here you have the option to enter your email and request a magic link. Or you can login with wordpress.
* If you click "Instantly Log In With Email", you will be taken to a screen to enter a code. The code is a backup to the magic link. In case the magic link doesn't work, the user can enter the code.

### Test

**Login with Wordpress**

- [ ] Click Log In.
- [ ] Click Log In with Wordpress.com
- [ ] This will pop up a web browser.
- [ ] Ensure at the end you are logged in.

The main reason to test this is because I update the deprecated `onActivityForResult`. It was just easier to handle it in the fragment.

**Login With Magic Link**

- [ ] Click Log In
- [ ] Enter your email and click "Instantly Log In With Email"
- [ ] Once you are on the "Enter Code" screen, go check your email.
- [ ] In your email, click the "Log In" button to deeplink into the app so you are auto logged in.
- [ ] You should be logged in.

**Login With Magic Link Code**

- [ ] Click Log In
- [ ] Enter your email and click "Instantly Log In With Email"
- [ ] Once you are on the "Enter Code" screen, go check your email.
- [ ] Copy the code from the email and go back to the app.
- [ ] Enter the code in the textfield and submit the code.
- [ ] This should login you in.

### Release

No Additional release update needed.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
+++ b/Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
@@ -127,6 +127,7 @@ class SignInFragment: MagicLinkableFragment() {
                     Toast.makeText(context, getString(state.messageRes), Toast.LENGTH_LONG).show()
                 }
                 is MagicLinkRequestUiState.Success -> {
+                    viewModel.resetState()
                     hideDialogProgress()
                     showConfirmationScreen(state.username, false)
                     AnalyticsTracker.track(
@@ -134,6 +135,10 @@ class SignInFragment: MagicLinkableFragment() {
                         AnalyticsTracker.CATEGORY_USER,
                         "user_requested_login_link"
                     )
+                }
+
+                else -> {
+                    // no-ops
                 }
             }
         }
--- a/Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
+++ b/Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
@@ -36,10 +36,7 @@ class SignInFragment: MagicLinkableFragment() {
 
     val viewModel: RequestMagicLinkViewModel by viewModels()
 
-
-    private var loginWithWordpress: Button? = null
-
-    private var mAuthState: String? = null
+    private var authState: String? = null
 
     private var signUpCallback: SignUpCallback? = null
 
@@ -68,7 +65,7 @@ class SignInFragment: MagicLinkableFragment() {
             val authSuccess = WordPressUtils.processAuthResponse(
                 activity?.application as Simplenote?,
                 authorizationResponse,
-                mAuthState,
+                authState,
                 true
             )
             if (!authSuccess) {
@@ -87,19 +84,16 @@ class SignInFragment: MagicLinkableFragment() {
     override fun inflateLayout(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
         val view = inflater.inflate(R.layout.fragment_login, container, false)
 
-        loginWithWordpress = view.findViewById(R.id.button_login_with_wordpress)
+        val loginWithWordpress: Button? = view.findViewById(R.id.button_login_with_wordpress)
         loginWithWordpress?.setOnClickListener {
-            val ctx = context ?: return@setOnClickListener
             val authRequestBuilder = WordPressUtils.getWordPressAuthorizationRequestBuilder()
 
             // Set unique state value.
-
-            // Set unique state value.
-            mAuthState = "app-" + UUID.randomUUID()
-            authRequestBuilder.setState(mAuthState)
+            authState = "app-" + UUID.randomUUID()
+            authRequestBuilder.setState(authState)
 
             val request = authRequestBuilder.build()
-            authService = AuthorizationService(ctx)
+            authService = AuthorizationService(loginWithWordpress.context)
             authService?.getAuthorizationRequestIntent(request)?.let {
                 resultLauncher.launch(it)
             }
@@ -150,7 +144,7 @@ class SignInFragment: MagicLinkableFragment() {
 
     override fun onActionButtonClicked(view: View, emailEditText: EditText) {
         if (NetworkUtils.isNetworkAvailable(requireContext())) {
-            viewModel?.requestLogin(emailEditText.text.toString())
+            viewModel.requestLogin(emailEditText.text.toString())
         } else {
             showDialogError(getString(R.string.simperium_dialog_message_network))
         }
--- a/Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
+++ b/Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
@@ -1,7 +1,6 @@
 package com.automattic.simplenote.authentication
 
 import android.app.Activity
-import android.content.Context
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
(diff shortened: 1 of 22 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
```diff
@@ -1,49 +1,137 @@
...
 @AndroidEntryPoint
 class SignInFragment: MagicLinkableFragment() {
 
     val viewModel: RequestMagicLinkViewModel by viewModels()
 
-    private var loginWithPassword: TextView? = null
+
+    private var loginWithWordpress: Button? = null
+
+    private var mAuthState: String? = null
```
khaykov: I would avoid mixing hungarian notation into kotlin :)

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
```diff
@@ -1,49 +1,137 @@
...
-                intent.putExtra("EXTRA_IS_LOGIN", true)
-                this.startActivity(intent)
-                act.finish()
+
+        loginWithWordpress = view.findViewById(R.id.button_login_with_wordpress)
+        loginWithWordpress?.setOnClickListener {
+            val ctx = context ?: return@setOnClickListener
+            val authRequestBuilder = WordPressUtils.getWordPressAuthorizationRequestBuilder()
+
+            // Set unique state value.
```
khaykov, line 88: Duplicate comment.

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
```diff
@@ -1,49 +1,137 @@
...
+const val TAG = "SignInFragment"
 
 @AndroidEntryPoint
 class SignInFragment: MagicLinkableFragment() {
 
     val viewModel: RequestMagicLinkViewModel by viewModels()
 
-    private var loginWithPassword: TextView? = null
+
+    private var loginWithWordpress: Button? = null
```
khaykov: Looks like we can remove loginWithWordpress from here, since it's only used to set a listener in inflateLayout.

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
```diff
@@ -1,49 +1,137 @@
...
-        loginWithPassword?.setOnClickListener { _ ->
-            activity?.let { act ->
-                val intent = Intent(act, SimplenoteCredentialsActivity::class.java)
-                intent.putExtra("EXTRA_IS_LOGIN", true)
-                this.startActivity(intent)
-                act.finish()
+
+        loginWithWordpress = view.findViewById(R.id.button_login_with_wordpress)
+        loginWithWordpress?.setOnClickListener {
+            val ctx = context ?: return@setOnClickListener
```
khaykov: Can the context be null here? If so, should we use a context from the view instead?

Simplenote/src/main/java/com/automattic/simplenote/authentication/magiclink/MagicLinkConfirmationFragment.kt
```diff
@@ -0,0 +1,122 @@
...
+import com.automattic.simplenote.viewmodels.MagicLinkUiState
+import com.simperium.android.ProgressDialogFragment
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
+class MagicLinkConfirmationFragment : Fragment() {
+
+    private var progressDialogFragment: ProgressDialogFragment? = null
+
+    private var codeEditText: EditText? = null
```
khaykov, line 34: Wonder if those View variables could be local?

khaykov (commented):
I'm waiting for the rate limit to get better, but in meantime a couple of comments:

- When I enter the wrong code, the message says "Link no longer valid" which is misleading.

- After entering code I'm brought to the initial login screen:
30e1c239e72723c346a3cb9141966aab

- Email delivery is pretty finicky, but is probably no the app's issue :)

- After rotating a screen when "Link no longer valid" dialog is visible, I get a "We're having problems" dialog appearing.
b5e1f7674efeaee0a6c726c60ab2d290

- There is no connection check when pressing Login button after entering code. With no internet I'm getting "we're having problems" toast.

khaykov:
Just a heads up - I tested with a regular, non cursed account, and both magic links and code work as expected. So I think we can ignore the "going to back to title screen after entering code" issue.

Simplenote/src/main/java/com/automattic/simplenote/authentication/magiclink/MagicLinkConfirmationFragment.kt
notandyvee: Yes to `codeEditText`. But no to `progressDialogFragment`. The reason being is we show and hide it from multiple places.

khaykov: Yes, I only meant the views - codeEditText, actionCodeButton and loginWithPassword in this case:)

roundhill (changes requested):
Looking good! I have a few UX requests to match up with iOS:

* Change `Instantly log in with email` button label to `Log in with email`

* Can you add the same `Or` and nice `Enter password` button on the code entry view?
IMG_1930

Also one more nice thing iOS does is not require you to enter your email again if you select `Enter password`:

IMG_5EF1CE57AC75-1

notandyvee:
@roundhill I did everything except:

> Also one more nice thing iOS does is not require you to enter your email again if you select Enter password:

There may be changes that need to be made with simperium. I'll confirm. So either way best to wait until we discuss this.

@khaykov ready for re-review. After your thorough review I found some small general fragment issues and viewmodel issues. I also updated some of the error messages. Again, keep in mind we have to balance security vs transparency here.

khaykov:
@notandyvee Thanks for the changes! I found one issue - when opening magic link, the app "closes" after logging you in:
982fa3ba0a64a84049203b259cc63e24

notandyvee:
Thanks @khaykov . I'll investigate. The reason this is happening seems to be related to a bug I was fixing. Originally I would call `Simplenote#loginWithToken` and start `NotesActivity`. But doing so caused two `NoteActivity`s to pop up. During my testing it worked. But maybe I missed something.

notandyvee:
Hey @khaykov , so I think I fixed it. If you want to re-test. Idky since simperium is a bit of a black box for me. But I was able to reproduce what you encountered. The change I made should fix it. Let me know if you run into any problems.

roundhill (changes requested):
Working great, I have a few more UI requests if that's cool ;)

1. Can the line in the `Or` not extend out into the padding area? Try to make it flush with where the buttons end:
Screenshot 2024-07-23 at 4 06 25 PM

2. Update the top of the code view to have the `We sent a code to...` text like on iOS:

IMG_1951

3. Update the label `Login with password` here to say `Enter password`
Screenshot 2024-07-23 at 4 10 01 PM

khaykov (approved):
Confirmed that the fix for opening link works 👍

roundhill (approved):
Working well, thanks for the UI updates :shipit:

notandyvee:
@roundhill changes requested made and and also fixed the issue with the code confirmation. Not sure what happened.

lol. sorry. Missed your message.

PR #1672, opened 2024-07-31
Go back one screen when confirming a code response says it expired.

notandyvee:
### Fix

During the magic link code confirmation and the server returns a 400 with `invalid-code`, we want to go back one screen. Basically so the user can request another magic link.

### Test

- [ ] Request a magic link.
- [ ] Wait 10 -15 minutes so it expires.
- [ ] Or request another magic link.
- [ ] In the code confirmation screen enter the invalid code.
- [ ] You should see a toast and be taken back a screen so you can submit another magic link.

### Review

Note: The invalid code has to be a previously valid one I think.

### Release

n/a

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/authentication/magiclink/MagicLinkAuthError.kt
+++ b/Simplenote/src/main/java/com/automattic/simplenote/authentication/magiclink/MagicLinkAuthError.kt
@@ -1,0 +1,7 @@
+package com.automattic.simplenote.authentication.magiclink
+
+enum class MagicLinkAuthError(val str: String) {
+    INVALID_CODE("invalid-code"),
+    REQUEST_NOT_FOUND("request-not-found"),
+    UNKNOWN_ERROR("")
+}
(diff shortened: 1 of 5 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>MagicLinkAuthError</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

roundhill (approved):
Tested fine! A user may type the code in wrong while it is still valid, which in this case kicks them back out to request another one which doesn't seem ideal. But I doubt users would enter it wrong very often as I think most would copy-paste. Maybe we could send different error codes based on the token being expired or not? What do you think @jleandroperez?

jleandroperez:
Most of the apps I've seen around, at most, have a button somewhere that says something like... `Send new code in x seconds`, and it only becomes clickable once the code actually expired.

We could try something like that perhaps? (as v2?)

PR #1676, opened 2024-08-08
Upgrade `release-toolkit`

mokagio:


fastlane/Fastfile
```diff
@@ -9,14 +9,26 @@ USER_ENV_FILE_PATH = File.join(Dir.home, '.simplenoteandroid-env.default')
...
-# Constants
 PROJECT_ROOT_FOLDER = File.dirname(File.expand_path(__dir__))
-GITHUB_REPO = 'automattic/simplenote-android'
+
+ORGANIZATION = 'automattic'
+REPO_NAME = 'simplenote-android'
+GITHUB_REPO = "#{ORGANIZATION}/#{REPO_NAME}".freeze
 DEFAULT_BRANCH = 'trunk'
+BUILDKITE_ORG = ORGANIZATION
+BUILDKITE_PIPELINE = REPO_NAME
```
mokagio, line 24: The idea with these over-engineered constants is that they should be easy to copy paste into other `Fastfile`s.

The more homogeneous we can keep the `Fastfile`s the easier it should be to maintain them.

fastlane/Fastfile
```diff
@@ -37,66 +49,79 @@ SUPPORTED_LOCALES = [
...
-  # This lane executes the steps planned on code freeze
-  # -----------------------------------------------------------------------------------
-  # Usage:
-  # bundle exec fastlane code_freeze [update_release_branch_version:<update flag>] [skip_confirm:<skip confirm>]
-  #
-  # Example:
-  # bundle exec fastlane code_freeze
-  # bundle exec fastlane code_freeze update_release_branch_version:false
-  # bundle exec fastlane code_freeze skip_confirm:true
-  #####################################################################################
```
mokagio, line 67: I decided to remove all these comments. Some were out of date, all felt to me like noise in the source.

fastlane/Fastfile
```diff
@@ -37,66 +49,79 @@ SUPPORTED_LOCALES = [
...
   desc 'Creates a new release branch from the current trunk'
```
mokagio:
```suggestion
  desc 'Creates a new release branch from the current default branch'
```

fastlane/Fastfile
```diff
@@ -708,3 +684,61 @@ platform :android do
...
+def release_branch_name(release_version: release_version_current)
+  "#{RELEASE_BRANCH_ROOT}#{release_version}"
+end
+
+def ensure_git_branch_is_release_branch
+  # Verify that the current branch is a release branch. Notice that `ensure_git_branch` expects a RegEx parameter
+  ensure_git_branch(branch: "^#{RELEASE_BRANCH_ROOT}")
+end
+
+RELEASE_BRANCH_ROOT = 'release/'
```
mokagio, line 747: This const _could_ go at the top with all the others, but since it's used only by the two release branch methods, it felt clearer to put it at the bottom here, after its consumers.

It's a bit inconsistent but I think it works.

Happy to move it up with the rest if you think it's not worth breaking the pattern.

AliSoftware:
FYI I've just opened https://github.com/Automattic/simplenote-android/pull/1677 to fix CVE on `rexml`… which might conflict with your draft PR here

AliSoftware:
Note that while #1677 fixed the `rexml`-related CVEs, there's currently [one remaining CVE about `nokogiri` still open](https://github.com/Automattic/simplenote-android/security/dependabot/73).

Fixing it will require to update `release-toolkit` to a newer version, as `nokogiri` is a dependency of it and that's where the version constraint on it comes from. Since the update of `release-toolkit` is already planned to be handled by this very PR, I'm leaving it to @mokagio and this PR to fix that last CVE as a side-effect of updating `release-toolkit`.

mokagio:
Closed in favor of #1680

PR #1688, opened 2024-09-09
Add automation to run code freeze in CI

mokagio:
This includes various automation updates. I kept running into things to uplift as I was working my way through the code freeze process...

- New automation and pipeline to build app in CI for distribution
- Adds lanes to trigger code freeze start and completion in CI
- Adds pipelines and scripts to run code freeze start and completion in CI
- CI pipeline uses quotes only when necessary and `:emoji:` syntax
- `release-toolkit` 12.0.0 upgrade
- Adopt `shared-pipeline-vars`
- Extract lanes in dedicated files under `fastlane/lanes/`
- Use `fastlane/metadata/android/en-US/` as the source of truth for Play Store metadata
- Automate milestones moving (see internal ref paaHJt-6Zx-p2)
- Create code freeze integration PR from Fastlane

Similar to https://github.com/Automattic/simplenote-ios/pull/1646, I will use my admin privileges to merge on `trunk` in order to test the process. The code is all an adaptation from other projects that have already used it in production for a while.

I'd be thankful for a posthumous review and will address any feedback coming that way, @Automattic/apps-infrastructure . Thanks!

As a side note, just a remark that this process of admin-merging PRs is not something I plan to keep resorting to. It is only a convenient way to test the code freeze in CI process. On top of that, most of the code here is copy-pasted or slightly adapted from the implementation in previous projects, which have been code reviewed and battle tested already.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/simplenote-android/milestone/45">2.34</a>. This milestone is due in less than 4 days.<br />
Please make sure to get it merged by then or assign it to a milestone with a later deadline.</td>
    </tr>
  </tbody>
</table>

</p>

mokagio:
The start code freeze [worked on the first go](https://buildkite.com/automattic/simplenote-android/builds/343) 😄 Glad I learned something from the various additions I had to do to the SImplenote implementation 😅 

image

One thing to notice on the milestone error: Because it could not find the 2.35 milestone, it logged the error and moved one _without adding the frozen marker._

We can either fix this by splitting the error handling so that adding the frozen marker occurs regardless, or we could file this as an edge case, given we'll rarely be in the state this repo is, where there are no future milestones planned.

Regardless, I will update the 2.34 milestone with the frozen marker manually this time around.

.buildkite/commands/checkout-release-branch.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -eu
+
+# We expect BUILDKITE_RELEASE_VERSION to be as an environment variable, e.g. by the automation that triggers the build on Buildkite.
+# It must use the `BUILDKITE_` prefix to be passed to the agent due to how `hostmgr` works, in case this runs on a Mac agents.
+if [[ -z "${BUILDKITE_RELEASE_VERSION}" ]]; then
+    echo "BUILDKITE_RELEASE_VERSION is not set."
+    exit 1
+fi
```
AliSoftware, line 8: 🚨  As I've mentioned in the past on other PRs, we should **really not use env vars prefixed with `BUILDKITE_`**.
 - Those are reserved for Buildkite and using such env var names ourselves could break Buildkite's behavior
 - But more importantly, this trick is not needed in the first place, as there are other ways to achieve the same outcome without needing to rely on that workaround

Basically the idea to fix this is to resolve the `RELEASE_VERSION` env var in the `yml`, and pass it as a `$1` parameter to this shell script (`command: .buildkite/commands/checkout-release-branch.sh "$RELEASE_VERSION"`); that way, the env var is resolved at `pipeline upload` time on the `upload` agent, and the whole issue (about `hostmgr` on Mac agents only forwarding env vars that are prefixed with `BUILDKITE_` to the VMs where the bash script is run) becomes irrelevant.

See https://github.com/Automattic/apps-infra-plans/issues/69 and https://github.com/bloom/DayOne-Apple/pull/24228#discussion_r1719712518 for more details.

.buildkite/commands/checkout-release-branch.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -eu
+
+# We expect BUILDKITE_RELEASE_VERSION to be as an environment variable, e.g. by the automation that triggers the build on Buildkite.
+# It must use the `BUILDKITE_` prefix to be passed to the agent due to how `hostmgr` works, in case this runs on a Mac agents.
+if [[ -z "${BUILDKITE_RELEASE_VERSION}" ]]; then
```
AliSoftware, line 5: Such a test won't work as expected given `set -u`. `bash` will exit with "undefined variable" error due to `set -u` before even getting to testing if its (undefined) value is `-z`. You need to use the `if [[ -z "${FOO:-}" ]];` trick instead in those cases.

.buildkite/commands/checkout-release-branch.sh
```diff
@@ -0,0 +1,14 @@
...
+    echo "BUILDKITE_RELEASE_VERSION is not set."
+    exit 1
+fi
+
+# Buildkite, by default, checks out a specific commit.
+# For many release actions, we need to be on a release branch instead.
+BRANCH_NAME="release/${BUILDKITE_RELEASE_VERSION}"
```
AliSoftware, line 12: 1️⃣  Step 1 of my suggestion to avoid the need of the ugly workaround for `BUILDKITE_` prefix in the env var name:
```suggestion
RELEASE_VERSION="${1?:RELEASE_VERSION is not set}"

# Buildkite, by default, checks out a specific commit.
# For many release actions, we need to be on a release branch instead.
BRANCH_NAME="release/${RELEASE_VERSION}"
```

.buildkite/commands/release-build.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "--- :computer: Use deterministic APK name for next steps"
+
+OUTPUT_DIR="build"
+if [ "$(find "$OUTPUT_DIR/*.apk" | wc -l)" -gt 1 ]; then
+  echo "Found more than one APK in $OUTPUT_DIR."
+  exit 1
+fi
+
+ORIGINAL_APK_PATH=$(ls -1 "$OUTPUT_DIR/*.apk")
```
AliSoftware, line 20: Nitpick but wondering if it wouldn't be nicer to use bash glob patterns for this instead?

```bash
shopt -s nullglob
FOUND_APKS=($OUTPUT_DIR/*.apk)
if [[ ${#FOUND_APKS[@]} -gt 1 ]]; then
  …
fi
ORIGINAL_API_PATH="${FOUND_APKS[0]}"
```

.buildkite/pipeline.yml
```diff
@@ -1,44 +1,50 @@
...
   - wait
 
-  - label: "🛠 Build and Test"
+  - label: ":hammer_and_pick: Build and Test"
     command: ./gradlew --stacktrace testRelease
-    plugins: *common_plugins
+    plugins: [$CI_TOOLKIT_PLUGIN]
+    notify:
+      - github_commit_status:
+          context: Build and Test
```
AliSoftware, line 22: Love ❤️ the custom GitHub Commit Status name to not only keep a nice name but also avoid the odd name that would be generated by Buildkite due to snake-casing the `hammer-and-pick` emoji name otherwise… 🙂

.buildkite/release-pipelines/complete-code-freeze.yml
```diff
@@ -0,0 +1,32 @@
...
+steps:
+  - label: Complete Code Freeze
+    key: complete_code_freeze
+    plugins: [$CI_TOOLKIT_PLUGIN]
+    command: |
+      echo '--- :robot_face: Use bot for Git operations'
+      source use-bot-for-git wpmobilebot
+
+      echo '--- :git: Checkout release branch'
+      .buildkite/commands/checkout-release-branch.sh
```
AliSoftware, line 18: 2️⃣   Step 2 of my suggestion to avoid the need of the ugly workaround for `BUILDKITE_` prefix in the env var name:
```suggestion
      .buildkite/commands/checkout-release-branch.sh "$RELEASE_VERSION"
```

In this case the `$RELEASE_VERSION` env var will be resolved at `pipeline upload` time, which runs on the `upload` agent. Then when the actual job for "Complete Code Freeze" starts, the command it runs would thus already look like `.buildkite/commands/checkout-release-branch.sh 12.3`, so no more issue with `hostmgr` filtering out env vars when forwarding the command script to the VM.

.buildkite/shared-pipeline-vars
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
+# to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
+
+export CI_TOOLKIT_PLUGIN_VERSION="3.5.1"
```
AliSoftware, line 6: Not sure we really need to `export` this as it doesn't really _need_ to be visible outside of the script? 🤷

fastlane/lanes/release.rb
```diff
@@ -0,0 +1,269 @@
...
+    repository: GITHUB_REPO,
+    numbers: [pr_number],
+    to_milestone: release_version
+  )
+
+  # Return the PR URL
+  pr_url
+end
+
+def trigger_buildkite_release_build(branch:, beta:)
```
AliSoftware, line 238: Worth pondering if it'd make sense to move that one with the other similar ones in `lanes/ci_releases.rb`?

AliSoftware (commented):
I have to admit I didn't really review the changes in the `Fastfile` because with most of the changes being code just being code moves from `Fastfile` to `lanes/*.rb`, this made it not easy to browse the `diff` and see what was just moved vs what might have also had code changes too.

But I still skimmed the rest of the diff and left some other feedbacks to fix on your next iteration.

Issue #1580, opened 2022-12-28
Long lists jump to the bottom when making changes


phantomlord2oo:
[Type] Bug

### Expected

Long lists jump to the bottom when making changes

### Observed
When editing a long list (like my grocery list) whenever I make a change the app automatically jumps to the end of the list.

### Reproduced
List with around 77 lines will jump to bottom of the list after making a change to anywhere else on the list (even changing just 1 character).

 If you reproduced the bug on a specific device, please include the details of the device and app.  For example:
Device Make: Samsung
Device Model: Galaxy S21
Android Version: 13
Simplenote Version: 2.25 (153)

joashrajin:
+1 at 5857035-zd-woothemes

reginabally:
> Really good usually but there's a problem using it now on my Samsung s21 phone, cursor jumps to the end of the note when I'm using it. Really really annoying. I might have to stop using this. Really hope this can be fixed...

Device: Samsung Galaxy S21 5G
App version name: 2.25
Android version: Android 13 (SDK 33)

[4-star app review](https://play.google.com/console/u/0/developers/7957760354032996428/app/4975233711145652703/user-feedback/review-details?reviewId=d35eff96-88d5-4380-b6a6-f80e0640e778&corpus=PUBLIC_REVIEWS)

reginabally:
> Used to work smoothly, but now view shifts unbidden to the end of text if you're editing at a point higher up. This seems to happen if you pause briefly. The cursor stays where you were, but the viewing pane moves to the bottom of the text. This makes editing a longer note all but impossible.

Device: Samsung Galaxy S20
App version name: 2.25
Android version: Android 13 (SDK 33)

[1-star app review](https://play.google.com/console/u/0/developers/7957760354032996428/app/4975233711145652703/user-feedback/review-details?reviewId=4e9c1741-3209-4354-9ae1-d4efdb1b66d6&corpus=PUBLIC_REVIEWS)

CameronMilliken:
I have used your app for years (8 maybe) and loved it that whole time! Unfortunately recently it has been spontaneously scrolling to the bottom of the note intermittently while the app is open; it's becoming unusable since I can't reliably type, or read through my note/s without having to scroll through multiple times. The cursor stays in place.

thabotswana:
[...]

PR #1690, opened 2024-09-10
[Bug] Fixed long list jumping in Android 13 samsung devices.

notandyvee:
### Fix

This PR attempted to fix a long standing issues on Android 13 samsung devices: #1580 . While editing a note and moving around the cursor, it can be observed that the view takes you all the way to the bottom of the screen, just above the tags. The tags part is very important as we will briefly go into.

Based on my extensive testing I figured out **why** this bug was happening. If you pay close attention when the bug occurs, you'll notice that the visible screen stops right above the tags component. This is very specific as it is not technically the very bottom of the screen. This led me to investigate `NestedScrollView` and not the `SimplenoteEditText`, which I initially investigated.

I was able to drill down on a specific method in `NestedScrollView`, `computeScrollDeltaToGetChildRectOnScreen`. The key is `scrollYDelta`. I observed that the parameter, `Rect` has a weird offset between what the `scrollY` position is and what the `Rect.top` position is after the `EditText` gets focus. Unfortunately I could not figure out the exact cause of the problem.   Debugging the app didn't show me a useful debug stack, which really hindered understanding the root cause.

With that said, I came up with a hack to make this work. I created a custom view, `CustomNestedScrollView`. You just override `requestChildRectangleOnScreen`, which is the starting point for these calculations. You then just check if the device manufacturer and the android version numbers matches, and force set the `Rect.top` to the `scrollY`, which is the scroll position of the `NestedScrollView`. Meaning it should be an accurate representation of the top y position of the screen.

**Note**: In order to avoid bigger issues, I opted to gate this fix to samsung devices running Android 13. Should we attempt to gate it further by checking the keyboard? Is it accurate?

### Test

- [ ] Test using a samsung device running Android 13 using the stock samsung keyboard.
- [ ] Test creating a note and editing it.
- [ ] Test using a really long note, as described in the issue. Trying and reproduce the issue.

### Review

N/A

### Release

N/A


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Message">
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR is still a Draft: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

mokagio:
@notandyvee @danilo04 FYI, I merged `trunk` in here because of a change in the GitHub check requirement which would otherwise have prevented this PR to land in either `trunk` or `release/2.34`.

image

PR #1693, opened 2024-09-11
Add message to allow user to login manually in the enter email screen.

### LINKED ISSUE #1689: Add log in with email label with option to sign in with password
Beneath the Log in with email button, let's add this text:

`We'll email you a code to log in, or you can log in manually.`

Link `log in manually` with a button to the username/password form.

notandyvee:
### Fix #1689 

This PR adds a new label on the screen to enter your email for a magic link. That lable will have a link to login manually, meaning with password instead.

### Test

- [ ] On app launch click login.
- [ ] Under the "Log In With Email" button, there is a new label, "We'll email you a code to log in, or you can log in manually".
- [ ] The "log in manually" is hyperlinked.
- [ ] Clicking the hyperlink should take you to login with password screen.

### Release
 N/A

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
+++ b/Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
@@ -2,14 +2,17 @@ package com.automattic.simplenote.authentication
 
 import android.app.Activity
 import android.content.Intent
+import android.graphics.Color
 import android.net.Uri
 import android.os.Bundle
+import android.text.method.LinkMovementMethod
 import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.Button
 import android.widget.EditText
+import android.widget.TextView
 import android.widget.Toast
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.fragment.app.viewModels
@@ -101,6 +104,17 @@ class SignInFragment: MagicLinkableFragment() {
                 "wpcc_button_press_signin_activity"
             )
         }
+        val manualLoginTextView = view.findViewById<TextView>(R.id.sign_in_login_manually)
+        val message = getString(R.string.signin_login_with_email_manually);
+        val span = StrUtils.generateClickableSpannableString(LOGIN_MANUALLY_SUBSTRING, message
+        ) {
+            val email = getEmailEditText()
+            showLoginWithPassword(activity, email?.text?.toString())
+        }
+
+        manualLoginTextView.text = span
+        manualLoginTextView.movementMethod = LinkMovementMethod.getInstance()
+        manualLoginTextView.highlightColor = Color.TRANSPARENT
         return view
     }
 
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/simplenote-android/milestone/45">2.34 ❄️</a>. The due date for this milestone has already passed.<br />
Please assign it to a milestone with a later deadline or check whether the release for this milestone has already been finished.</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/src/main/java/com/automattic/simplenote/authentication/SignInFragment.kt
```diff
@@ -101,6 +104,17 @@ class SignInFragment: MagicLinkableFragment() {
                 "wpcc_button_press_signin_activity"
             )
         }
+        val manualLoginTextView = view.findViewById<TextView>(R.id.sign_in_login_manually)
+        val message = getString(R.string.signin_login_with_email_manually);
+        val span = StrUtils.generateClickableSpannableString(LOGIN_MANUALLY_SUBSTRING, message
```
roundhill: I don't think this will work for localized strings... can it instead work like this string does?

`<string name="simperium_footer_signup">By creating an account, you agree to our %1$s%2$s%3$sTerms and Conditions%4$s</string>`

And the code that formats it:

```
String.format(
                if (this.isLogin) this.resources.getString(R.string.simperium_footer_login) else this.resources.getString(
                    R.string.simperium_footer_signup
                ), "<span style=\"color:#", colorLink, "\">", "</span>"
            )
```

Something like that?

roundhill (changes requested):
Worked pretty well, requesting a change to support localizing the linked text. 

Also, when you press the back arrow from the password field, you end up at the welcome screen. Is that expected? I think it works ok but I expected to be back to the email login form.

notandyvee:
> Also, when you press the back arrow from the password field, you end up at the welcome screen. Is that expected?

Unfortunately it is. This is a remnant of the old way of doing things. I can update it, but would require me to replace the user and password screen with one that isn't made by simperium. Or we can ticket it to fix it. Basically a bigger fix.

> I don't think this will work for localized strings... can it instead work like this string does?

And great catch! This was an oversight. You are correct it wouldn't work for localized strings. The fix could be to localize the substring as well. Not sure if there is risk with the people translating and needing to translate the last part.

As to why I opted out of doing what you are suggesting is because the entire `TextView` is clickable. It felt off to highlight part of the text but make the whole thing clickable. But if you don't mind, I can do it that way instead. You let me know @roundhill .

roundhill:
> As to why I opted out of doing what you are suggesting is because the entire `TextView` is clickable. It felt off to highlight part of the text but make the whole thing clickable. But if you don't mind, I can do it that way instead. You let me know @roundhill .

Ah, I see. I think it's ok to make it all clickable so that the localization can work properly.

notandyvee:
@roundhill I made the changes. Using the old method for now. Will create an issue to explore using a `ClickableSpan` instead of a click action on the text view.

roundhill (approved):
Looks good to me, and tested well!

PR #1695, opened 2024-09-19
Merge release 2.34 into `trunk`

mokagio:
```mermaid
%%{init: { 'gitGraph': { 'mainBranchName': 'trunk', 'mainBranchOrder': 1000, 'showCommitLabel': false }}}%%
gitGraph
   branch release/2.34
   checkout release/2.34
   commit
   commit
   commit
   branch merge/release-2.34-into-trunk
   checkout merge/release-2.34-into-trunk
   commit
   checkout trunk
   commit
   commit
   merge merge/release-2.34-into-trunk
```

cc @Automattic/apps-infrastructure to have a look at the automation changes. Context paaHJt-6Zx-p2


fastlane/Fastfile
```diff
@@ -169,64 +169,6 @@ platform :android do
     trigger_release_build(branch_to_build: release_branch_name)
   end
 
-  desc 'Updates store metadata and runs the release checks'
-  lane :finalize_release do |skip_confirm: false|
```
mokagio, line 173: Moved to `release.rb`

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Message">
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR has the <code>Releases</code> label: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lanes/ci_releases.rb
```diff
@@ -1,26 +0,0 @@
...
-    buildkite_trigger_build(
-      buildkite_organization: BUILDKITE_ORGANIZATION,
-      buildkite_pipeline: BUILDKITE_PIPELINE,
-      branch: release_branch_name(release_version: release_version),
-      pipeline_file: File.join(PIPELINES_ROOT, 'complete-code-freeze.yml'),
-      message: "Complete code freeze for #{release_version}",
-      environment: { RELEASE_VERSION: release_version }
-    )
-  end
-end
```
mokagio, line 26: Deleted as discussed in https://github.com/Automattic/simplenote-ios/pull/1655#discussion_r1753900531. See also https://github.com/wordpress-mobile/WordPress-iOS/pull/23589

fastlane/lanes/release.rb
```diff
@@ -203,36 +312,43 @@ def report_milestone_error(error_title:)
...
-
-  pr_url = create_pull_request(
-    api_token: token,
-    repo: GITHUB_REPO,
-    title: title,
-    head: Fastlane::Helper::GitHelper.current_git_branch,
-    base: base_branch,
-    labels: 'Releases'
-  )
+def create_backmerge_pr!
```
mokagio, line 315: What do you think of what I did here?

- `create_backmerge_pr!` does exactly what is says, returns the URL of the PR or crashes if something went wrong. This allows callers to declare they expect a single backmerge PR, like in this repo for a code freeze or a release finalization.
- `create_backmerge_prs!` below allows for more backmerge PRs, like when we finalize a hotfix. It also has the `!` suffix because if something goes wrong it will call `UI.user_error!`

.buildkite/release-pipelines/finalize-release.yml
```diff
@@ -0,0 +1,24 @@
...
+
+steps:
+  - label: Finalize Release
+    plugins: [$CI_TOOLKIT_PLUGIN]
+    command: |
+      echo '--- :robot_face: Use bot for git operations'
+      source use-bot-for-git wpmobilebot
+
+      echo '--- :git: Checkout Release Branch'
+      .buildkite/commands/checkout-release-branch.sh
```
mokagio, line 12: I plan to followup with a dedicated PR that surfaces `RELEASE_VERSION` here and remove reading it from the environment in the script.

I think by doing that we might also make the script more portable and possibly move it to the CI plugin. Being such a simple script, I wonder what's the value, but then again, it would make the implementation, the error handling, and the messages it print consistent across project, which I think is desirable for us.

It would also make the implementation consistent with the recent changes accessing `RELEASE_VERSION` in the pipeline `command` that we made in some iOS projects.

fastlane/lanes/release.rb
iangmaia: I wrote a [short comment about this in another PR](https://github.com/Automattic/simplenote-ios/pull/1655#discussion_r1755228424) but I guess it got lost amidst all PRs and discussions currently happening.

My main question is whether we can be sure the validation on `create_backmerge_pr!` is guaranteed to be semantically correct, meaning: I think there may be cases when a previous release might take a bit longer or there is a hotfix in progress, then you "unexpectedly" will have a backmerge.

mokagio: I see... 🤔 

So, for example if `finalize_code_freeze` for `1.2` runs when `1.1` hasn't landed yet? Or if we are code freezing `1.2` while `1.1.1` is still in progress?

I'm tempted to say that 1) that should not happen because one should not run a code freeze before having finalized the previous release and 2) [the action should only create backmerges in branches that are from later releases](https://github.com/wordpress-mobile/release-toolkit/blob/ee03eb9232597fd812c18d3fb0fbf91130facf97/spec/create_release_backmerge_pull_request_spec.rb#L97-L116) (e.g. 1.2.1 or 1.3). That to me suggests that if we have more than one backmerge then something went wrong and we should abort.

Having said that... Maybe it's best to be flexible (and account for me not being able to predict every future possible case in which this action should run) and let release manager deal with multiple backmerge PRs when only one should be there.

In fact, now that I think about it, the release manager would have to deal with them anyway because the failure implemented here is _after_ the PRs have been created.

Here's what I might do:

- Remove the `create_backmerge_pr!`
- Move the logic to print to STDOUT annotate CI into `create_backmerge_prs!` (I am keen on keeping the plural form to make it clear at call site that more than one might be opened)
- Add a `warn_on_multiple_prs_opened: false` parameter that lanes such as `finalize_code_freeze` can use to change the STDOUT and CI annotation to include a warning if more than one backmerge was create

What do you think @iangmaia ?

mokagio: Actually @iangmaia sorry I'm going to merge this before you have a chance to reply to this because I had to make a few more changes to the release branch and will need to submit a new build. As such, it's convenient to have this PR on `trunk` before the `publish_release` task will open a new integration PR (because of the new changes).

mokagio:
Admin merging as discussed above, https://github.com/Automattic/simplenote-android/pull/1695#discussion_r1767796358

fastlane/lanes/release.rb
iangmaia: > So, for example if finalize_code_freeze for 1.2 runs when 1.1 hasn't landed yet? Or if we are code freezing 1.2 while 1.1.1 is still in progress?

Yeah, in that case the PR would be created from 1.1.1 to 1.2 on `finalize_hotfix`, but I guess it's just a matter of using the plural method in this case.

> In fact, now that I think about it, the release manager would have to deal with them anyway because the failure implemented here is after the PRs have been created.

Makes sense to me 👍  I'm still a bit on the fence on what's best to do but annotating a warning sounds like a good compromise.

> sorry I'm going to merge this before you have a chance to reply to this because I had to make a few more changes to the release branch and will need to submit a new build

No worries!

mokagio: A minor followup: 

> Move the logic to print to STDOUT annotate CI into create_backmerge_prs! (I am keen on keeping the plural form to make it clear at call site that more than one might be opened)

I run into the case of using `create_backmerge_pr` expecting it to return a single URL in [this Simplenote iOS build](https://buildkite.com/automattic/simplenote-ios/builds/1160#01921de0-8921-4b6c-93cf-53dd88a882fd) (all my own doing). There result was a buggy UX where the link to the PR in the build annotation didn't work. The reason? The markup had an array in the URL with got rendered as empty:

image

mokagio: Following up via https://github.com/Automattic/simplenote-android/pull/1698

PR #1718, opened 2024-12-12
[Tooling] Update GitHub Actions

iangmaia:
Reference: p7H4VZ-59z-p2

Simplenote Android doesn't use one of the GitHub Actions about to have its older versions deprecated as `actions/upload-artifact`, `actions/download-artifact` or `actions/cache`, but I used this opportunity to upgrade the Action `gradle/actions/wrapper-validation` and to move the Buildkite Gradle Wrapper Validation to the Linter queue.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,7 +7,8 @@ agents:
 steps:
   - label: ":gradle: Gradle Wrapper Validation"
     command: validate_gradle_wrapper
-    plugins: [$CI_TOOLKIT_PLUGIN]
+    agents:
+      queue: linter
     notify:
       - github_commit_status:
           context: Gradle Wrapper Validation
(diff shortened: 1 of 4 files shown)


spencertransier (approved):

spencertransier:
@iangmaia Approved to unblock, but is there a reason this change would affect the linting errors that Danger picked up?

iangmaia:
> @iangmaia Approved to unblock, but is there a reason this change would affect the linting errors that Danger picked up?

It's not related to these changes but they might have been dormant for some reason -- my guess is that the Linter didn't pick them up in a previous PR due to a bug (now fixed) that it would use a different Rubocop version than the one configured in the project Bundle.
I confirmed locally that Rubocop reports the same issues, so I fixed them and pushed on a3a00bd6b1abb68237a6164e6474def159ec7bce.

PR #1733, opened 2025-08-01
Update libraries and fix crashes

mzorz:
Note: this PR sits on top of #1731 

### Fix
This app was not being consistently maintained so I had to bring a lot of libraries up to date and do some changes to make it all work as expected (also fixes a couple crashes, one of them listed [here](https://github.com/Automattic/simplenote-android/pull/1731#issuecomment-3137536096) from the previous PR)

### Test
Smoke test the app, create entires, verify they sync on another platform, check against the production app.

### Review
@ParaskP7 

### Release
Maintenance update - support for Android 15


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">View files have been modified, but no screenshot or video is included in the pull request. Consider adding some for clarity.</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

PasscodeLock/src/main/java/org/wordpress/passcodelock/AbstractPasscodeKeyboardActivity.java
```diff
@@ -24,9 +25,11 @@ public abstract class AbstractPasscodeKeyboardActivity extends Activity {
     protected InputFilter[] filters = null;
     protected TextView topMessage = null;
 
+	@SuppressLint("RestrictedApi")
```
mzorz, line 28: Added this to avoid linter complaining for now, but tracking separately here https://github.com/Automattic/simplenote-android/issues/1734

PasscodeLock/build.gradle
```diff
@@ -26,16 +26,21 @@ android {
             shrinkResources false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
+        releaseTest {
+            initWith release
+            minifyEnabled false
+            shrinkResources false
```
mzorz: had to create a different task to run tests on Release with minify disabled, since otherwise there was a ClassDefNotFound error on the AbstractAppLock class, and R8 was stripping the PasscodeLock classes.  Kudos to Claude Code for helping me out on this one, see https://github.com/Automattic/simplenote-android/pull/1733/commits/8c792b453529ad8997a80dada224f30c3da999c9

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
```diff
@@ -234,7 +234,8 @@ public void onPageScrollStateChanged(int state) {
 
     @Override
     public void onBackPressed() {
-        handleBackPressed();
+        super.onBackPressed();
```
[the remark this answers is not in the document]
mzorz: this suggestion collides with the linter rule to call super in this deprecated method - I'm looking to make the minimum changes, a revamp to use getOnBackPressedDispatcher() consistently can be revisited later.

Simplenote/src/main/java/com/automattic/simplenote/authentication/NewCredentialsActivity.kt
```diff
@@ -90,6 +90,7 @@ open class NewCredentialsActivity : AppCompatActivity() {
     }
 
     override fun onBackPressed() {
+        super.onBackPressed()
```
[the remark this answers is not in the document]
mzorz: as said in https://github.com/Automattic/simplenote-android/pull/1733#discussion_r2250755187 I'm looking to do the minimum changes at this moment - calling super satisfies the linter error and the code after will be executed (it finishes the activity and launches a new one).

Simplenote/src/main/AndroidManifest.xml
```diff
@@ -85,7 +86,7 @@
 
                 <data
                     android:host="app.simplenote.com"
-                    android:pathPattern="/account/.*/reset?redirect=simplenote://launch"
+                    android:pathPattern="/account/.*/reset"
```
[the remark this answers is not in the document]
mzorz: there is additional validation in DeepLinkActivity - and again, looking to use the least changes at this moment.

gradlew
```diff
@@ -114,7 +114,7 @@ case "$( uname )" in                #(
   NONSTOP* )        nonstop=true ;;
 esac
 
-CLASSPATH=$APP_HOME/gradle/wrapper/gradle-wrapper.jar
+CLASSPATH="\\\"\\\""
```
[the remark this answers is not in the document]
mzorz: leaving this one up to @ParaskP7 🙏

gradlew.bat
```diff
@@ -56,32 +59,34 @@ set JAVA_EXE=%JAVA_HOME%/bin/java.exe
...
+echo Please set the JAVA_HOME variable in your environment to match the 1>&2
+echo location of your Java installation. 1>&2
 
 goto fail
 
 :execute
 @rem Setup the command line
 
-set CLASSPATH=%APP_HOME%\gradle\wrapper\gradle-wrapper.jar
+set CLASSPATH=
```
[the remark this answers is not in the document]
mzorz: leaving this one up to @ParaskP7 🙏

gradlew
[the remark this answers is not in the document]
ParaskP7: Nah, this is not a manual change and we trust Gradle when updating these files automatically, can be ignored, but thanks bot! 😊

gradlew.bat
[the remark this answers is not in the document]
ParaskP7: Nah, this is not a manual change and we trust Gradle when updating these files automatically, can be ignored, but thanks bot! 😊

ParaskP7:
👋 @mzorz and just an FYI on my part that I started with reviewing this change today, but I didn't manage to finish with it. I'll continue tomorrow. ⏳

PS: In the meanwhile I am not sure if my review should be enough here, lots of changes and lots of library updates. On the other hand, it might be okay if we move faster (than usual) and only do a final check right at the end, when we're done with all the changes. What we want is to at least know that next time we'll need to release an update, we should be set. Wdyt? 🤔

mzorz:
> PS: In the meanwhile I am not sure if my review should be enough here, lots of changes and lots of library updates. On the other hand, it might be okay if we move faster (than usual) and only do a final check right at the end, when we're done with all the changes. What we want is to at least know that next time we'll need to release an update, we should be set. Wdyt? 🤔

I'm fine either way, the main reason I split the PRs and asked for your review in this one is because this PR updates libraries and fixes crashes on the base PR (the one you wrote). The actual functional changes are on the other PRS (targeting SDK 35 plus changing the cwac:anddown library which is not 16kb-page-size compatible).

If you don't spot anything weird on this PR, this one should be good to go (since the PR we merged actually crashes the app anyway, this would be already putting things in a better place). 😅 

I'll have someone else review and smoke test the rest of the changes too 👍

ParaskP7:
👋 @mzorz and I am on it! 🏃 💨 💨 💨

> I'm fine either way, the main reason I split the PRs and asked for your review in this one is because this PR updates libraries and fixes crashes on the base PR (the one you wrote). The actual functional changes are on the other PRS (targeting SDK 35 plus changing the cwac:anddown library which is not 16kb-page-size compatible).

Got it! 👍 

> If you don't spot anything weird on this PR, this one should be good to go (since the PR we merged actually crashes the app anyway, this would be already putting things in a better place). 😅

True... 😅 

> I'll have someone else review and smoke test the rest of the changes too 👍

Awesome, thanks! 🙇

.buildkite/pipeline.yml
```diff
@@ -16,7 +16,7 @@ steps:
   - wait
 
   - label: ":hammer_and_pick: Build and Test"
-    command: ./gradlew --stacktrace testRelease
+    command: ./gradlew --stacktrace testReleaseTest
```
ParaskP7: Warning (⚠️): I am not sure why this `testRelease` task was chosen for CI builds (part of this initial 9d79c7296ce2e3c555045da4f8d5c29d209bb11b commit), and any `release` task for that matter, but I would suggest that instead us adding this new `releaseTest` build type, and all those new ProGuard configurations, that we instead change that to make CI run debug builds:

For the `Build and Test` specific job, to change to to:
```yml
  - label: ":hammer_and_pick: Unit Test"
    command: ./gradlew --stacktrace testDebugUnitTest
    plugins: [$CI_TOOLKIT_PLUGIN]
    notify:
      - github_commit_status:
          context: Unit Test
```

For the `Lint` specific job, to change that to:
```yml
  - label: ":microscope: Lint"
    command: ./gradlew --stacktrace lintDebug
    plugins: [$CI_TOOLKIT_PLUGIN]
    notify:
      - github_commit_status:
          context: Lint
```

FYI: The `Prototype Build`, which is using [build_and_upload_prototype_build](https://github.com/Automattic/simplenote-android/blob/trunk/fastlane/Fastfile#L136-L179), already builds on the [debug](https://github.com/Automattic/simplenote-android/blob/trunk/fastlane/Fastfile#L139) build type, not the `release`, so with this change everything will match `debug` for such development related builds.

PS: [release-build.yml](https://github.com/Automattic/simplenote-android/blob/trunk/.buildkite/release-build.yml) is dedicate to building releases, which point to build_for_distribution this [build_for_distribution](https://github.com/Automattic/simplenote-android/blob/trunk/fastlane/lanes/build.rb#L4-L31) that targets the [release](https://github.com/Automattic/simplenote-android/blob/trunk/fastlane/lanes/build.rb#L12C19-L12C26) build type and as such handles this case.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-bin.zip
```
ParaskP7: 1. Suggestion (💡): Consider using the `all` distribution, which includes sources and docs, just like we do (almost) everywhere else, for consistency purposes.
2. Suggestion (💡): Consider adding the `distributionSha256Sum` for that specific distribution (see [release-checksums](https://gradle.org/release-checksums/)).

FYI: For your convenience feel free to use this command: `./gradlew wrapper --gradle-version=8.14.3 --distribution-type=all --gradle-distribution-sha256-sum=ed1a8d686605fd7c23bdf62c7fc7add1c5b23b2bbc3721e661934ef4a4911d7c && ./gradlew wrapper --distribution-type=all`

PasscodeLock/build.gradle
```diff
@@ -26,16 +26,21 @@ android {
...
+        }
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_11
+        targetCompatibility JavaVersion.VERSION_11
```
ParaskP7, line 32: Suggestion (💡): Just because I saw this, I remembered about the [.java-version](https://github.com/Automattic/simplenote-android/blob/trunk/.java-version) file, which it too is still pointing to Java `17.0`, and maybe we can update that to `21.0`, just like we did on DOAndroid quite recently.

PasscodeLock/build.gradle
```diff
@@ -12,7 +12,7 @@ plugins {
 
 android {
```
ParaskP7, line 13: Suggestion (💡): Unrelated to the changes in this PR, but since we are trying to update it all, maybe we should consider updating `jcenter()` with `mavenCentral()` everywhere, see warning:

`JCenter Maven repository is no longer receiving updates: newer library versions may be available elsewhere`

PasscodeLock/build.gradle
```diff
@@ -12,7 +12,7 @@ plugins {
...
     namespace 'org.wordpress.passcodelock'
-    compileSdk 34
+    compileSdk 36
 
     defaultConfig {
         minSdk 23
```
ParaskP7, line 18: Suggestion (💡): Again, somehow unrelated, but since we are here, let's remove this unnecessary `testInstrumentationRunner` from this module, it not having any `androidTest` and all.

PasscodeLock/build.gradle
```diff
@@ -26,16 +26,21 @@ android {
...
 }
 
 dependencies {
-    implementation 'androidx.appcompat:appcompat:1.0.2'
-    implementation 'androidx.preference:preference:1.0.0'
+    implementation 'androidx.appcompat:appcompat:1.7.1'
+    implementation 'androidx.preference:preference:1.2.1'
 }
 
 android.buildTypes.all { buildType ->
```
ParaskP7, line 41: Suggestion (💡): Again, somehow unrelated, but since we are cleaning-up things, let's remove these unnecessary (or duplicate) [gradle.properties](https://github.com/Automattic/simplenote-android/blob/trunk/PasscodeLock/gradle.properties) configuration from this module, in the end, it seems that it could look like this:

`gradle.properties`
```properties
passcodelock.password_preference_key=passcode_lock_prefs_password_key
passcodelock.password_enc_secret=5-maggio-2002-Karel-Poborsky
passcodelock.fingerprint_enabled_key=passcode_lock_prefs_fingerprint_enabled_key

```

PasscodeLock/build.gradle
```diff
@@ -26,16 +26,21 @@ android {
...
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_11
+        targetCompatibility JavaVersion.VERSION_11
     }
 }
 
 dependencies {
-    implementation 'androidx.appcompat:appcompat:1.0.2'
-    implementation 'androidx.preference:preference:1.0.0'
+    implementation 'androidx.appcompat:appcompat:1.7.1'
```
ParaskP7, line 37: Suggestion (💡): Just like done [here](https://github.com/Automattic/simplenote-android/blob/trunk/build.gradle#L67-L74), I would recommend that now we are updating all those versions, we also extract all them to make sure we avoid updating one version and not the other, while at the same time make the first step towards using version catalogs (`libs.versions.toml`).

FYI: I can then take it from there and quickly, hopefully painlessly (famous last words), migrate this project to version catalogs. 

PS: Same for all those `min/compile/targetSdkVersion` values everywhere.

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
```diff
@@ -73,6 +73,13 @@ public boolean enoughToFilter() {
             return false;
         }
 
+		// solves a crash after updating dependencies in which this method
+	    // gets called in super() instantiation before the mTokenizer variable
+	    // is instantiated
+		if (mTokenizer == null) {
+			return false;
+		}
```
ParaskP7, line 81: Minor (🔍): AS complains that the indentation is not correct here, because it uses tabs instead of empty spaces.

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.kt
```diff
@@ -27,8 +27,13 @@ class DeepLinkActivity : AppCompatActivity() {
                 startActivity(intent)
             }
             VERIFIED_WEB_SCHEME -> {
-                // New MagicLink
-                startMagicLinkConfirmation(uri)
+                // Check if this is a password reset URL
+                if (uri.path?.contains("/account/") == true && uri.path?.contains("/reset") == true) {
```
ParaskP7, line 31: ℹ️ Info: Acknowledging commit 28c7b258c8a6f27417808f6cf276f52c95591228 and change! 👍

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.kt
```diff
@@ -88,6 +93,21 @@ class DeepLinkActivity : AppCompatActivity() {
         }
     }
 
+    private fun handlePasswordReset(uri: Uri) {
+        val redirectParam = uri.getQueryParameter("redirect")
+        if (redirectParam == "simplenote://launch") {
```
ParaskP7: Question (❓): The logic within both `if/else` seems the same to me, what that done on purpose? 🤔

Simplenote/src/main/java/com/automattic/simplenote/NoteEditorActivity.java
```diff
@@ -249,7 +250,7 @@ private void handleBackPressed() {
 
             finish();
         } else {
-            super.onBackPressed();
+	        getOnBackPressedDispatcher().onBackPressed();
```
ParaskP7, line 253: ℹ️ Info: Acknowledging commit 22fefdef25267033af1b4d2b07a1fd2d1ecaaad5 and change! 👍

Simplenote/src/main/res/layout/empty_view.xml
```diff
@@ -17,7 +18,7 @@
         android:layout_marginTop="@dimen/padding_large"
         android:layout_width="@dimen/icon_empty"
         android:src="@drawable/ic_notes_24dp"
-        android:tint="?attr/emptyImageColor">
+        app:tint="?attr/emptyImageColor">
     </ImageView>
```
ParaskP7, line 22: Minor (🔍): Just an FYI that there are multiple other places where such an empty `</ImageView>` tag can be collapsed into `/>` (3 in here, another 3 within `fragment_note_editor.xml`, 5 within `note_list_row.xml`, etc).

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModelTest.kt
```diff
@@ -144,16 +144,26 @@ class CollaboratorsViewModelTest {
 
     @Test
     fun collaboratorAddedAfterStoppedListeningChangesShouldNotUpdateUiState() = runTest {
```
ParaskP7, line 146: 1. Praise (❤️): Thanks for fixing this unit test! 🥇 
2. Warning (⚠️): In addition to having unit test working, let's consider having instrumented tests working as well, in this or another follow-up PR (if not too much of a trouble). On that side of things I have the following:
    1. First, we need to add this `androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.3.0'` dependency to have those instrumented tests compile.
    2. Then, we need to use `./gradlew Simplenote:connectedDebugAndroidTest` over `./gradlew connectedDebugAndroidTest` to make them run, because otherwise compilation fails on the `PasscodeLock` module with multiple duplicate class failures on Kotlin (we might consider getting to the bottom of this).
    3. Finally, after managing to run instrumented tests successfully, you'll find that at least `13/74` fail.

ParaskP7 (changes requested):
👋 @mzorz !

I have reviewed and tested this PR as per the instructions, great job updating all these dependencies! 🌟 x 🌟 ^ 🌟

-----

I have left a couple of warnings (⚠️), one question (❓), a few suggestions (💡) and some minor (🔍) comments for you to consider. I am **NOT** going to approve this PR just yet, mainly because of this one CI related warning (⚠️), but other than that I think this is good to go, we can keep iterating on modernizing this codebase on subsequent PRs.

Simplenote/src/main/java/com/automattic/simplenote/DeepLinkActivity.kt
mzorz: oof good catch, fixed in 2589709f - it should just launch the app when capturing a password reset link and the app is installed (this to do the same thing it was doing before, no behavior change)

ParaskP7: Awesome, thanks for the fix @mzorz , so in the end, no query parameters no nothing needed here, right? 😊

.buildkite/pipeline.yml
mzorz: Good question - the `testRelease` task was running tests on Release, but R8 is stripping off classes from PasscodeLock in the generated build being tested, so the suggestion I took from Claude was to create a new flavor that didn't have minify enabled so the classes would exist and tests would pass, that's why . See https://github.com/Automattic/simplenote-android/commit/8c792b453529ad8997a80dada224f30c3da999c9 -  I can see that we could continue to play with the proguard arguments and perhaps set minify to true in release just to use the same task that existed before, but I wonder if we can. Thoughts?

ParaskP7:
FYI: I was able to reproduce [this exact problem](https://github.com/Automattic/simplenote-android/pull/1737#issuecomment-3154067905) while testing this PR, but only did it once, then I just couldn't any more... 🤷

PasscodeLock/build.gradle
mzorz: 👍 done in 9229a7a6

ParaskP7: Awesome, thanks @mzorz ! 🙇 ❤️ 🚀

PasscodeLock/build.gradle
mzorz: done in 6f51438f

PasscodeLock/build.gradle
mzorz: done in e9447a50 - any efforts regarding instrumented tests I'd say let's tackle separately

PasscodeLock/build.gradle
mzorz: sounds like a great idea, can we do that on a different PR?

.buildkite/pipeline.yml
ParaskP7: 1. Yea, no, I understood all that @mzorz , Claude is playing smart here... 😅 But, I honestly don't think we need to run unit tests on release, thus just changing this `testRelease` to `testDebug`, then making it explicit, changing `testDebug` to `testDebugUnitTest` might be the best we can do here.
2. Also, I am in favor of the above, because I would try to avoid anything ProGuard related in this PR, and just for testing purposes, let alone creating a new build type, just for that. I really think we should avoid these changes unless totally necessary.

PasscodeLock/build.gradle
ParaskP7: Awesome, thanks @mzorz ! 🙇 ❤️ 🚀

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
mzorz: good call fixed in 7a4fb340

PasscodeLock/build.gradle
ParaskP7: Awesome, thanks @mzorz ! 🙇 ❤️ 🚀

PasscodeLock/build.gradle
ParaskP7: Totally @mzorz , resolving this! 💯

Simplenote/src/main/res/layout/empty_view.xml
mzorz: that's right - I only fixed those where I also visited to fix something else. We can fix them all separately.

Simplenote/src/main/java/com/automattic/simplenote/widgets/SimplenoteEditText.java
ParaskP7: Great, thanks @mzorz ! 🙇 ❤️ 🚀

Simplenote/src/main/res/layout/empty_view.xml
ParaskP7: 💯 this is truly minor (🔍) @mzorz and I only mention it because you already tried to do something similar, resolving this now... 😊

gradle/wrapper/gradle-wrapper.properties
mzorz: thanks much for the pointers here! done in daae06db

ParaskP7: Awesome, thank you so much @mzorz ! 🙇 ❤️ 🚀

mzorz:
Thanks a lot for all this feedback @ParaskP7 ! 🙌  really appreciated.

> FYI: I was able to reproduce https://github.com/Automattic/simplenote-android/pull/1737#issuecomment-3154067905 while testing this PR, but only did it once, then I just couldn't any more... 🤷

Thanks for flagging - all of that is being addressed either on the PR changing to `targetSdk 35` or later on but on top of that PR. 

Just so we're on the same page: for this PR, I concentrated on:
- upgrading the libraries to latest (not the target SDK)
- making sure what exists doesn't crash

This PR is already large enough with that aim (upgrading the libraries versions and updating code to match behavior changes only).

So, I fixed some of the comments that were related to the changes above, but all further suggestions I think are better to be handled in separate PRs.

> Warning (⚠️): In addition to having unit test working, let's consider having instrumented tests working as well, in this or another follow-up PR (if not too much of a trouble).

Thanks for the pointers here! Also for instrumented tests, I agree it'd be nice to bring them back to life, yet it exceeds the aim of updating the libraries only. I also can imagine integrated tests will change after the changes we're making due to the edge-to-edge UI changes by targeting SDK 35, so let's keep that task isolated for the time being.

I think the only really blocking thing in this PR should be [this](https://github.com/Automattic/simplenote-android/pull/1733#discussion_r2253878342). Let's continue the discussion there 👍

ParaskP7:
💯 @mzorz and I agree with all that you said [above](https://github.com/Automattic/simplenote-android/pull/1733#issuecomment-3155054999), thank you so much for applying all and any of my suggestions! 🙇

FYI: I've resolved all comments but did left the ones we chose to differ as unresolved, just to make it more visible going forward, when and if we decide to do something about them. 🤞

> I think the only really blocking thing in this PR should be https://github.com/Automattic/simplenote-android/pull/1733#discussion_r2253878342. Let's continue the discussion there 👍

Yes, let's do that and then merge this! 🚀

.buildkite/pipeline.yml
mzorz: Got it - it took me 5 extra minutes to grasp these concepts better and think I understand your suggestion now 😅. I removed the unneeded flavors and updated the yaml to get buildkite to run unit tests on debug in this commit here 3616f1ac and 0448e630.

ParaskP7: Awesome @mzorz , as soon as CI goes 🟢 this is ready to merge, thank you so much! 🙇 ❤️ 🚀

ParaskP7 (approved):

mzorz:
Thanks once more for you thorough review @ParaskP7 ! We'll take over the things that were left unresolved in subsequent PRs as agreed 🙌

PasscodeLock/build.gradle
mzorz: Following up on this - tried to remove the file but these are used in BuildConfig which is referenced in DefaultAppLock. The keys are actually used as Android prefs keys, so we need to keep them otherwise changing them it means an app upgrade would be reading a different setting. So, keeping them for now 👍

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModelTest.kt
mzorz: Moved this one out to a ticket here https://github.com/Automattic/simplenote-android/issues/1744

PasscodeLock/build.gradle
ParaskP7: Thanks for checking @mzorz but just to make it clear, I suggested to remove everything else but these 3 from this [gradle.properties](https://github.com/Automattic/simplenote-android/blob/trunk/PasscodeLock/gradle.properties) file, is that what you tried? 😊

`before`
```
android.enableJetifier=true
android.useAndroidX=true

passcodelock.password_preference_key=passcode_lock_prefs_password_key
passcodelock.password_enc_secret=5-maggio-2002-Karel-Poborsky
passcodelock.fingerprint_enabled_key=passcode_lock_prefs_fingerprint_enabled_key

ossrhUsername=hello
ossrhPassword=world

signing.keyId=byebye
signing.password=secret
signing.secretKeyRingFile=/home/user/secret.gpg
```

`after`
```
passcodelock.password_preference_key=passcode_lock_prefs_password_key
passcodelock.password_enc_secret=5-maggio-2002-Karel-Poborsky
passcodelock.fingerprint_enabled_key=passcode_lock_prefs_fingerprint_enabled_key
```

Simplenote/src/test/java/com/automattic/simplenote/viewmodels/CollaboratorsViewModelTest.kt
ParaskP7: Awesome, thanks @mzorz , marking this as resolved now. 🙇 ❤️

PasscodeLock/build.gradle
mzorz: OH I misread that. You're right we can keep those ones only. PR https://github.com/Automattic/simplenote-android/pull/1745

ParaskP7: Thanks so much @mzorz , #1745 comment resolved! 🙇 ✅ ❤️ 🚀

PR #1737, opened 2025-08-02
Update part 2: updating for Android 15 (sdk 35) compatibility

mzorz:
This PR sits on top of https://github.com/Automattic/simplenote-android/pull/1733
### Fix
Checked the migration tool (Android SDK upgrade assistant) on Android Studio, the path "Android 14 / API level 34 -> Android 15 / API level 35"

These are the checcks performed.
### Open JDK API:
- [x] 1) changes to formatting strings API

> Validation of argument index, flags, width, and precision are now more strict when using the following String.format() and Formatter.format() APIs:
String.format(String, Object[])
String.format(Locale, String, Object[])
Formatter.format(String, Object[])
Formatter.format(Locale, String, Object[])
For example, the following exception is thrown when an argument index of 0 is used (%0 in the format string):

- [x] 2) Changes to component type of Arrays.asList().toArray()
> When using Arrays.asList(...).toArray(), the component type of the resulting array is now an Object—not the type of the underlying array's elements. So the following code throws a ClassCastException:
> `String[] elements = (String[]) Arrays.asList("one", "two").toArray();`
> For this case, to preserve String as the component type in the resulting array, you could use Collection.toArray(Object[]) instead:
> `String[] elements = Arrays.asList("two", "one").toArray(new String[0]);`

- [x] Changes to language code handling
> When using the Locale API, language codes for Hebrew, Yiddish, and Indonesian are no longer converted to their obsolete forms (Hebrew: iw, Yiddish: ji, and Indonesian: in). When specifying the language code for one of these locales, use the codes from ISO 639-1 instead (Hebrew: he, Yiddish: yi, and Indonesian: id).

- [x] `SequencedCollection` collisions
> The new SequencedCollection API can affect your app's compatibility after you update compileSdk in your app's build configuration to use Android 15 (API level 35):
Collision with MutableList.removeFirst() and MutableList.removeLast() extension functions in kotlin-stdlib
The List type in Java is mapped to the MutableList type in Kotlin. Because the List.removeFirst() and List.removeLast() APIs have been introduced in Android 15 (API level 35), the Kotlin compiler resolves function calls, for example list.removeFirst(), statically to the new List APIs instead of to the extension functions in kotlin-stdlib.

> If an app is re-compiled with compileSdk set to 35 and minSdk set to 34 or lower, and then the app is run on Android 14 and lower, a runtime error is thrown:

We don't use those methods so all good.

- [x] Collision with other methods in Java
[...]

PR #1738, opened 2025-08-03
Replaced markdown library `com.commonsware.cwac:anddown` with jetbrains' markdown for html conversions

mzorz:
### Fix
- replaced markdown library `com.commonsware.cwac:anddown` which was not 16kb-page-size compatible, with `org.jetbrains:markdown:0.7.3` to handle markdown to html conversions
- converted `NoteMarkdownFragment` to Kotlin

### Test
1. Edit an entry, verify the markdown option is checked in the three dots menu
2. add a markdown title with ###, checkboxes, a quote, etc
3. exist to save
4. open the entry on another platform (i.e. web or desktop) and see the titles being rendered in bold

### Review

cc @ParaskP7


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Class <code>SimplenoteMarkdownFlavorDescriptor</code> is missing tests, but <code>unit-tests-exemption</code> label was set to ignore this.</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
```diff
@@ -0,0 +1,379 @@
...
+                    "<link href=\"https://fonts.googleapis.com/css?family=Noto+Serif\" rel=\"stylesheet\">" +
+                    "<meta name=\"viewport\" content=\"width=device-width,minimum-scale=1,initial-scale=1\">\n" +
+                    cssContent + "</head><body>"
+
+            val flavour = CommonMarkFlavourDescriptor()
+            val parsedTree = MarkdownParser(flavour).buildMarkdownTreeFromString(sourceContent)
+            var parsedMarkdown = HtmlGenerator(sourceContent, parsedTree, flavour).generateHtml()
+
+            // Set auto alignment for lists, tables, and quotes based on language of start.
+            parsedMarkdown = parsedMarkdown
```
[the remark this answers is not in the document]
mzorz: looking for minimal changes atm, thanks

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    override fun onDeleteObject(bucket: Bucket<Note>, note: Note) {
+    }
+
+    override fun onNetworkChange(bucket: Bucket<Note>, type: Bucket.ChangeType, key: String) {
+    }
+
+    override fun onSaveObject(bucket: Bucket<Note>, note: Note) {
+        if (note == mNote) {
```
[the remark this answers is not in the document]
ParaskP7: 1. FYI: Actually `if (note == mNote) { ... }` and `if (note.equals(mNote)) { ... }` is equal in Kotlin, both check for value (structural) equality, not reference equality, which is done using `===` instead, thus Kotlin converted the Java `equals` call to `==`.
2. FYI.2: The only diff is that `if (note == mNote) { ... }` is less safe when `note` is null, but this is not the case here... 🤷 
    - The `==` operator in Kotlin checks structural equality by translating to a call to the `equals()` method, but with important null-safety handling. Specifically, `a == b` compiles to `a?.equals(b) ?: (b === null)`, so if `a` is `null`, it only returns `true` if `b` is also `null`.
    - `note.equals(mNote)` directly calls the equals method on note. This can throw a `NullPointerException` if note is `null` (unless you’re using Kotlin’s safe call operator as in `note?.equals(mNote)`).

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
```diff
@@ -0,0 +1,380 @@
...
+import com.automattic.simplenote.utils.ThemeUtils
+import com.google.android.material.snackbar.Snackbar
+import com.simperium.client.Bucket
+import com.simperium.client.BucketObjectMissingException
+import org.intellij.markdown.flavours.commonmark.CommonMarkFlavourDescriptor
+import org.intellij.markdown.html.HtmlGenerator
+import org.intellij.markdown.parser.MarkdownParser
+import java.lang.ref.SoftReference
+
+class NoteMarkdownFragment : Fragment(), Bucket.Listener<Note> {
```
ParaskP7, line 45: 1. Suggestion (💡): I always recommend to have a Java-to-Kotlin conversion commit first, before making any additional manual change to a file. Otherwise, it can become quite tricky to review. For example, in this case I had to make sure to review both the conversation and the `getMarkdownFormattedContent(...)` change, all in one commit, risking missing bits.
2. Suggestion (💡): This is just future suggestion for such Java-to-Kotlin conversions, that using an extra commit for `.java` > `.kt` renames helps with preserving the files history and makes it easier to diff the changes, for example:

Before committing the change, you could ☑️ this `Extra commit for .java > .kt renames` on that IDE commit dialog:

image

And then, instead of having a change like this below, where the Java file is deleted, and another Kotlin file is created:

image

You now get this instead:

image

This above is possible because prior to that an automated `Rename .java to .kt` is commit for you:

image

Really hope that this will become handy next time you'll converted a Java file! 🙏

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
```diff
@@ -0,0 +1,380 @@
...
+        @JvmStatic
+        fun getMarkdownFormattedContent(cssContent: String, sourceContent: String): String {
+            val header = "<html><head>" +
+                    "<link href=\"https://fonts.googleapis.com/css?family=Noto+Serif\" rel=\"stylesheet\">" +
+                    "<meta name=\"viewport\" content=\"width=device-width,minimum-scale=1,initial-scale=1\">\n" +
+                    cssContent + "</head><body>"
+
+            val flavour = CommonMarkFlavourDescriptor()
+            val parsedTree = MarkdownParser(flavour).buildMarkdownTreeFromString(sourceContent)
+            var parsedMarkdown = HtmlGenerator(sourceContent, parsedTree, flavour).generateHtml()
```
ParaskP7, line 59: Warning (⚠️): I asked my friend the AI to compare the previous `com.commonsware.cwac:anddown` Java implementation to this new `org.jetbrains:markdown` Kotlin implementation, it comparing these 2 versions:

`java`
```
If this Java code below, which is using the 'com.commonsware.cwac:anddown' library:

String parsedMarkdown = new AndDown().markdownToHtml(
sourceContent,
AndDown.HOEDOWN_EXT_STRIKETHROUGH | AndDown.HOEDOWN_EXT_FENCED_CODE |
AndDown.HOEDOWN_EXT_QUOTE | AndDown.HOEDOWN_EXT_TABLES,
AndDown.HOEDOWN_HTML_ESCAPE
);
```

`kotlin`
```
val flavour = CommonMarkFlavourDescriptor()
val parsedTree = MarkdownParser(flavour).buildMarkdownTreeFromString(sourceContent)
var parsedMarkdown = HtmlGenerator(sourceContent, parsedTree, flavour).generateHtml()
```

It came back with this table below:

Feature | AndDown (Hoedown) | JetBrains Markdown (CommonMarkFlavourDescriptor)
-- | -- | --
Strikethrough | Yes (with HOEDOWN_EXT_STRIKETHROUGH) | No (CommonMark); Yes with GFMFlavourDescriptor
Fenced code blocks | Yes (with HOEDOWN_EXT_FENCED_CODE) | Yes (supported by default)
Block quotes | Yes (with HOEDOWN_EXT_QUOTE) | Yes (supported by default)
Tables | Yes (with HOEDOWN_EXT_TABLES) | No (CommonMark); Yes with GFMFlavourDescriptor
Task lists | No (unless using custom extensions) | No (CommonMark); Yes with GFMFlavourDescriptor

I then tested `Strikethrough`, `Tables` and `Task lists` and actually it was right, see below:

`EDIT`
EDIT

`PREVIEW` (before) -> `val flavour = CommonMarkFlavourDescriptor()`

PREVIEW (Before)

`PREVIEW` (after) -> `val flavour = GFMFlavourDescriptor()`

PREVIEW (After)

This too me looks like AI is right and we should use the `GFMFlavourDescriptor` in favor of `CommonMarkFlavourDescriptor`. Yes, for some reason the ~~Strikethrough~~ doesn't look too good (I don't see the actual strikethrough), same with the task lists, but it is nevertheless better than before, right? 🤔

ParaskP7 (changes requested):
👋 @mzorz !

I have reviewed and tested this PR as per the instructions, everything works as expected (ish), nicely done replacing `com.commonsware.cwac:anddown`! 🙌 x 🙌 ^ 🙌

-----

I have left one warning (⚠️) and a couple of future suggestions (💡) for you. I am **NOT** going to approve this PR just yet, mainly because of this one warning (⚠️), but other than that this all LGTM.

ParaskP7:
Ah, and btw @mzorz, unrelated to this PR, but as I was testing this PR, I found another "bug" related to the screen scrolling behavior as well as the keyboard being on top of the content, without a means to move the content up, above the keyboard, see below:

image

Notice that I can't scroll the content up to show this 3rd `[ ] Test Three` in that task list of mine, nor the content scrolls automatically when the keyboard is called-up. 🤔

mzorz:
Great testing there @ParaskP7 ! do you have a video for the screen scrolling behavior? I can't reproduce (testing both this version and the prod version)

ParaskP7:
Here's a recording for you @mzorz :

https://github.com/user-attachments/assets/73a368b4-48e6-4ef6-91d7-6794dfed5aa1

Does it help? 🙏

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
mzorz: Great catch, I thought I had `GFMFlavourDescriptor`  in one of my branches but no!

I did observe the prod app was not particularly good in the Preview tab so I didn't check one by one, see for example how an empty checkbox looks like on preview on prod:

image

Will update and perhaps try something about that sneaky strikethrough 👍

mzorz:
Wow that recording helps. I can't reproduce it here though 🤔 - could you paste the markdown here?

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
mzorz: oh I use the command line for git stuff.  I'll remember to do the rename, commit, then convert, commit 👍

mzorz:
@ParaskP7 good news, updated the code here 7e2623dc and now it supports markdown strikethrough correctly alongside all the other good bits the `GFMFlavourDescriptor` comes with.

Turns out the `GFMFlavourDescriptor` converted strikethrough tokens to the following, for example in converting the string `Not sure why we are ~~using~~ this`:
```
Not sure why we are <span class="user-del">using</span> this
``` 

which then was not being parsed correctly by the Android WebView since it would need some css treatment. So I checked the original code which produced this code:
```
Not sure why we are <del>using</del> this
```

The [deleted element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/del) is supported off the shelf since its basic html.

Had to do some digging into how the GFMFlavourDescriptor worked, and was able to override that mapping to produce `<del>` tags instead. It works now! see screenshot below:

image

mzorz:
> Here's a recording for you @mzorz :

- I was able to reproduce by filling in the same contents locally 👍
- I also see this is **not happening in production.**
- I also see this is **not happening on `trunk`**, so likely related to the targetSdk 35 and edge-to-edge changes in https://github.com/Automattic/simplenote-android/pull/1737

Will look into this 👍

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
mzorz: followed the advice and converted the utils file on this other PR in 3543434b730af41369aa63e67681ce6805fd8986 and f42e53e42f0ac076847ac4f3d9998bc394e2c0a6 👍

mzorz:
> > as I was testing this PR, I found another "bug" related to the screen scrolling behavior as well as the keyboard being on top of the content, without a means to move the content up, above the keyboard, see below:
> - I was able to reproduce by filling in the same contents locally 👍
> - I also see this is not happening in production.
> - I also see this is not happening on trunk, so likely related to the targetSdk 35 and edge-to-edge changes in https://github.com/Automattic/simplenote-android/pull/1737

 ✅ this [issue](https://github.com/Automattic/simplenote-android/pull/1738#issuecomment-3155517103) where content couldn't be scrolled and the virtual keyboard was hiding it was fixed in dd575ff54891084a95762e0f1dbf0289fe27a7b3  in https://github.com/Automattic/simplenote-android/pull/1737

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
ParaskP7: > Great catch, I thought I had GFMFlavourDescriptor in one of my branches but no!

Never when you need it, right! 🤣 

> I did observe the prod app was not particularly good in the Preview tab so I didn't check one by one, see for example how an empty checkbox looks like on preview on prod:

Yea, I know... 🤷 

> Will update and perhaps try something about that sneaky strikethrough 👍

Thanks so much @mzorz ! 🙇 ❤️

Simplenote/src/main/java/com/automattic/simplenote/NoteMarkdownFragment.kt
ParaskP7: This is awesome @mzorz , I did saw you doing that trick on this other PR and worked fantastically, much much better, thanks so much! 🙇 ❤️

ParaskP7:
> ✅ this https://github.com/Automattic/simplenote-android/pull/1738#issuecomment-3155517103 where content couldn't be scrolled and the virtual keyboard was hiding it was fixed in https://github.com/Automattic/simplenote-android/commit/dd575ff54891084a95762e0f1dbf0289fe27a7b3 in https://github.com/Automattic/simplenote-android/pull/1737

As per my [other](https://github.com/Automattic/simplenote-android/pull/1737#issuecomment-3157744066) comment, nicely done @mzorz , I just tested there and it works as expected, thank YOU! 🙇 ❤️ 🚀

ParaskP7:
> @ParaskP7 good news, updated the code here https://github.com/Automattic/simplenote-android/commit/7e2623dc4c69e04e2c3870d48ab3da98c5acaa3b and now it supports markdown strikethrough correctly alongside all the other good bits the GFMFlavourDescriptor comes with.

Awesome, I am checking `SimplenoteMarkdownFlavorDescriptor` as we speak and testing it, everything works as expected, nicely done extending `GFMFlavourDescriptor` @mzorz ! 🥇 

> Turns out the GFMFlavourDescriptor converted strikethrough tokens to the following, for example in converting the string
> `Not sure why we are ~~using~~ this:`
> 
> `Not sure why we are <span class="user-del">using</span> this`
> 
> which then was not being parsed correctly by the Android WebView since it would need some css treatment. So I checked the original code which produced this code:
> 
> `Not sure why we are <del>using</del> this`
> 
> The [deleted element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/del) is supported off the shelf since its basic html.

Hmmm, thanks for the explanation on all that, not sure why too... 🤷 

> Had to do some digging into how the GFMFlavourDescriptor worked, and was able to override that mapping to produce <del> tags instead. It works now! see screenshot below:

Indeed it does, tested! 🎉

ParaskP7 (approved):

mzorz:
Thanks a ton @ParaskP7 ! 🙌  I'll wait until the base PR is reviewed before merging, to help ease the review there 👍

PR #1808, opened 2026-05-11
[Tooling] Replace ActiveSupport's String#parameterize with a pure-Ruby helper

iangmaia:
### Fix

`release-toolkit` [v14.3.1](https://github.com/wordpress-mobile/release-toolkit/pull/709) dropped its `activesupport` runtime dependency. `generate_prototype_build_number` in our `Fastfile` still calls `String#parameterize`, which will raise `NoMethodError` as soon as we bump `release-toolkit` to `>= 14.3.1` (currently pinned to `~> 13.8`). Proactive fix so the bump doesn't break Prototype Builds — `wordpress-android` already hit this.

Adds a tiny pure-Ruby `parameterize` helper (downcase, replace runs of non-`[a-z0-9_-]` with `-`, squeeze, strip leading/trailing `-`) and routes the two call sites through it. Output matches ActiveSupport's `String#parameterize` for the branch/commit names involved.

### Test

1. `ruby -c fastlane/Fastfile` passes.
2. Prototype Build CI step still produces the expected build number for typical branch names.

### Review

The helper is a drop-in replacement for the two `.parameterize` call sites; behavior is unchanged for the inputs we pass (git branch / ref names).

### Release

Release notes are not needed (CI/tooling only).

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -261,7 +261,7 @@ platform :android do
   def generate_prototype_build_number
     if ENV['BUILDKITE']
       commit = ENV.fetch('BUILDKITE_COMMIT', nil)[0, 7]
-      branch = parameterize(ENV['BUILDKITE_BRANCH'])
+      branch = parameterize(ENV.fetch('BUILDKITE_BRANCH', nil))
       pr_num = ENV.fetch('BUILDKITE_PULL_REQUEST', nil)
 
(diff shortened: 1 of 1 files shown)


mokagio:
Thanks @iangmaia! I run into this issue in Gravatar's Android SDK https://github.com/Automattic/Gravatar-SDK-Android/pull/683 but didn't think to check other repos. Shame on me.

mokagio (approved):

PR #1815, opened 2026-05-29
[Diagnostic] Disable GMS Wearable integration to isolate Play Protect crash

mzorz:
### Fix

Diagnostic-only branch. **Do not merge.**

Several users report that the Android system dialog *"Google Play Protect Service keeps stopping"* appears every few minutes (and sometimes immediately) while editing notes in Simplenote. See [DODROID-884](https://linear.app/a8c/issue/DODROID-884/android-17-beta-crashstop-messages-while-editing-notes) and the [forum thread](https://forums.simplenote.com/forums/topic/android-app-keeps-crashing-google-play-protect-service/). One user is on a Pixel 6 Pro (likely Android 16 stable) where the dialog appears immediately on app launch.

The dialog is a Google Play Services process crash, not a Simplenote crash. The leading hypothesis is that our `WearableListenerService` causing GMS to bind to our app for Wearable transport / node discovery is what tickles the GMS Chimera crash on these Android builds. Simplenote is unusual in hosting that service alongside a continuous Simperium WebSocket session, which lines up with the user observation that "no other apps do this."

This branch strips every phone-app touchpoint with GMS Wearable:

- Drops `com.google.android.gms:play-services-wearable:19.0.0`
- Drops `wearApp project(':Wear')` bundling
- Removes `SimplenoteWearListenerService` and its `MESSAGE_RECEIVED` intent filter
- Removes the now-orphan `com.google.android.gms.version` meta-data (Wearable was the only GMS client lib)

The `:Wear` module source is untouched.

### Test

1. Build the diagnostic APK: `./gradlew :Simplenote:assembleDebug`
2. Output: `Simplenote/build/outputs/apk/debug/Simplenote-debug.apk` (applicationId `com.automattic.simplenote.debug`, installs side-by-side with production).
3. Hand the APK to an affected user (Pixel 6 Pro / Android 16+, or a Pixel 10 Pro / Android 17 beta tester).
4. Ask them to **uninstall the Play Store production Simplenote first** while testing — otherwise the production app's `WearableListenerService` is still on the device and Play Protect could still trigger from it.
5. Use the app normally for 15+ minutes of note editing.
6. **Pass:** no "Google Play Protect Service keeps stopping" dialogs. → GMS Wearable is confirmed as the trigger; next step is to gate the listener behind "a paired Wear node exists" or push a clean repro to Google.
[...]

PR #1829, opened 2026-06-21
Fix large-note editor crash (TransactionTooLargeException)

### LINKED ISSUE #1809: Google Play Protect service stops
DODROID-884

### Expected

Google Play Protect service runs stably in the background while editing notes, without interruption.

### Observed

Every few minutes while editing a note, a pop-up appears stating "Google Play Protect service has stopped" or "Google Play Protect service keeps stopping." Occasionally, Google Play Services or the Simplenote app itself crashes. The pop-ups appear mid-session (sometimes mid-word), disrupting the writing flow. No data loss or other functional effects have been observed beyond the disruptive notifications. The issue occurs consistently during note editing sessions and has been reported across multiple devices over the course of several days. It does not appear to affect other apps.

### Reproduced

The bug reproduces consistently during active note editing sessions:

1. Open the Simplenote app on Android.
2. Open or create a note and begin editing.
3. Continue editing for 3–4 minutes.
4. Observe a pop-up notification: "Google Play Protect service has stopped" or "Google Play Protect service keeps stopping."
5. Dismiss the pop-up and continue editing.
6. Observe the pop-up reappearing every few minutes.

---

**Device 1**
- Device Make: Google
- Device Model: Pixel 10 Pro
- Android Version: 17 (beta)

**Device 2**
- Device Make: Google
- Device Model: Pixel 6 Pro
- Android Version: Not specified

mzorz:
Closes https://github.com/Automattic/simplenote-android/issues/1809
Closes DODROID-884
Closes SIMPL-50

### Fix

Editing a note larger than ~1 MB and then leaving the editor (Home / Recents / process death) crashed with `TransactionTooLargeException` on `activityStopped`. The `note_content` `EditText` serialized its **entire text** into saved instance state, overflowing the ~1 MB binder transaction limit:

```
TransactionTooLargeException: data parcel size 1109936 bytes
  android:support:fragments
    viewState
      0x7f0901cc [size=1100880]   <- note_content
  at ActivityClient.activityStopped(...)
```

On Pixel devices the same oversized ContentCapture stream from the editor also crashed the on-device anti-phishing service (`com.google.android.odad`), which the OS surfaces as the **"Google Play Protect Service keeps stopping"** dialog the user originally reported. This is DODROID-884 — and it is **not** GMS Wearable (the earlier diagnostic PR #1815 ruled that out).

Changes (all in `NoteEditorFragment`):

- `mContentEditText.setSaveEnabled(false)` — stop persisting the note body in view state. The content is reloaded from Simperium (`mNote.getContent()`) on resume, so the saved-state copy was pure redundancy.
- Persist just the **cursor position + focus** ourselves in `onSaveInstanceState` / `onCreateView` so they survive activity recreation (the one thing we lose by disabling the EditText's own state save).
- `setImportantForContentCapture(IMPORTANT_FOR_CONTENT_CAPTURE_NO)` on API 30+ — stop feeding the body to ContentCapture, removing the odad / "Play Protect" trigger.

### Test

Repro requires a note big enough to push the saved-state parcel over ~1 MB.

1. Create/open a note with **> ~1 MB of text** in the editor.
2. Tap into the body so it has a cursor, type a character.
3. Developer Options -> **"Don't keep activities" -> ON** (makes the stop deterministic).
4. Press **Home**.

- **Before:** `TransactionTooLargeException` -> "Simplenote error" dialog (and on Pixel, the "Google Play Protect Service keeps stopping" dialog).
- **After:** no crash; reopening the note shows the full content intact and the cursor where you left it.

Verified on a Pixel 6 Pro (Android 16) with a 1.05 MiB note. An instrumented `saveHierarchyState` probe measured `note_content` saved-state size dropping from **1,100,872 bytes to 0** with the fix.

### Review

- The note body is reloaded from Simperium on resume, so disabling view-state save loses no content — only the transient cursor/scroll, and we now re-persist the cursor explicitly.
- `setImportantForContentCapture` is correctly guarded for API 30+ (minSdk 23).
- Worth a sanity check that normal small-note editing, autocomplete, checklist, and history flows behave unchanged.

### Release

Fixes a crash dialog ("Google Play Protect Service keeps stopping" / "Simplenote error") that appeared while editing large notes on recent Android versions.

### DIFF
--- a/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
+++ b/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java
@@ -21,6 +21,7 @@ import android.graphics.Typeface;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
 import android.os.AsyncTask;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.text.Editable;
@@ -117,6 +118,10 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
     public static final String ARG_PREVIEW_ENABLED = "preview_enabled";
 
     private static final String STATE_NOTE_ID = "state_note_id";
+    // DODROID-884: the content EditText no longer saves its own instance state (it would
+    // overflow binder for large notes), so we persist just the cursor position ourselves.
+    private static final String STATE_CURSOR_POSITION = "state_cursor_position";
+    private static final String STATE_CONTENT_HAS_FOCUS = "state_content_has_focus";
     private static final int AUTOSAVE_DELAY_MILLIS = 2000;
     private static final int MAX_REVISIONS = 30;
     private static final int PUBLISH_TIMEOUT = 20000;
@@ -136,6 +141,9 @@ public class NoteEditorFragment extends Fragment implements Bucket.Listener<Note
     private CursorAdapter mLinkAutocompleteAdapter;
     private CursorAdapter mTagAutocompleteAdapter;
     private boolean mIsLoadingNote;
+    // DODROID-884: cursor position/focus to restore after activity recreation; -1 = none
+    private int mRestoreCursorPosition = -1;
+    private boolean mRestoreContentFocus;
     private boolean mIsMarkdownEnabled;
     private boolean mIsPreviewEnabled;
     private ActionMode mActionMode;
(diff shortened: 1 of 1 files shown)


mzorz:
I've tested and verified this fixes the issue for SIMPL-50 reflected in  #1809 so will proceed to merge and create a new release

PR #1856, opened 2026-07-30
Realign the release branch on its remote during checkout

AliSoftware:
## What does it do?

Part of [AINFRA-2725](https://linear.app/a8c/issue/AINFRA-2725/fix-release-selection-and-checkout-release-branchsh-in-release-tooling), following the [WooCommerce iOS 25.1 release incident](https://woomobilep2.wordpress.com/2026/07/20/retrospective-dotview-crash-recurrence-woo-ios-25-1-2/). The same change is going out to every mobile product repo.

### 1. Realign the release branch on the remote (the actual fix)

`.buildkite/commands/checkout-release-branch.sh` fetched the release branch and checked it out, but never moved the local branch onto the fetched commit:

```sh
git fetch origin "$BRANCH_NAME"
git checkout "$BRANCH_NAME"
```

Buildkite cleans the working copy between jobs, but it can *reuse* it. A `refs/heads/release/x.y` left behind by an earlier job on the same agent therefore survives, and `git checkout` then simply switches to that stale local ref rather than to what was just fetched. Whatever runs next — the version bump, or the GitHub Release draft that `finalize_release` creates from `HEAD` — would then be based on the wrong commit.

The fix adds the missing realignment. `reset --hard` rather than `git pull`: no extra network round trip, and no merge commit if the refs diverged.

### 2. Reset to `FETCH_HEAD` rather than the remote-tracking ref

`git fetch origin <branch>` always writes `FETCH_HEAD`, but it only updates `refs/remotes/origin/<branch>` when the remote's fetch refspec covers that branch. With Buildkite's default `+refs/heads/*:refs/remotes/origin/*` the two are equivalent — but on a clone whose refspec was narrowed after `origin/<branch>` already existed, the fetch leaves that ref stale and `reset --hard "origin/$BRANCH_NAME"` lands on the old commit: the very failure this script exists to prevent, reintroduced through the back door. Resetting to `FETCH_HEAD` drops the refspec dependency entirely.

### 3. Standardize the script across all repos

Slightly tangent to the issue, but bundled deliberately: the script had drifted into **five different shapes** across the repos — argument required via `${1?…}` or `${1:?…}`, argument plus a `BUILDKITE_BRANCH` fallback, argument with a hand-rolled usage check, the same under a different variable name, and (in one repo) no argument at all, reading `RELEASE_VERSION` from the environment. Having rolled the same one-line fix out thirteen times this week, that divergence is pure friction.

All repos now share one canonical script. The resolution order is a **superset** of what every repo did before — argument, then `RELEASE_VERSION` from the environment, then the `release/*` branch the build runs on — so **no call site needed changing**. The `BUILDKITE_BRANCH` fallback only fires on a branch matching `^release/`, and derives the branch name back from that same value, so it cannot select a branch other than the one the build was already triggered on.

It also closes two latent bugs: under `bash -eu`, both `[[ -z "${RELEASE_VERSION}" ]]` on an unset variable and a bare `RELEASE_VERSION=$1` with no arguments abort with `unbound variable` before their intended usage message can print. And an argument that is *passed but empty* — what happens when a pipeline forwards an unset `$RELEASE_VERSION` — no longer resolves to a bare `release/`: it falls through to the environment variable, then to the `release/*` branch the build runs on, and finally to a clear error if none of those yields a version.

### 4. `release-toolkit` bump deferred to a separate PR

This repo is pinned to `~> 13.8`, so picking up [wordpress-mobile/release-toolkit#763](https://github.com/wordpress-mobile/release-toolkit/pull/763) means crossing a **major version**, not a patch bump. The 13.x → 14.0.0 migration replaces `ios_update_metadata_source` / `an_update_metadata_source` with `gp_update_metadata_source`, drops translator comments from PO files unless they are migrated into the `source_files` hash, and changes PO entry ordering — and this repo does call those actions. Doing it here would break the localization lanes.

So this PR ships the `checkout-release-branch.sh` fix on its own, which stands alone and needs no new gem version. The toolkit upgrade and its migration will follow in a dedicated PR.

## Testing instructions

No behaviour change on a fresh checkout, which is the normal case: the local branch is already at the fetched commit, so the reset is a no-op. The argument-resolution logic was exercised across all combinations (argument / empty argument / environment variable / `release/*` branch / trunk / feature branch / unset), and the resulting script passes `shellcheck` in every repo. The next release build exercising this script is the real check.

### DIFF
--- a/.buildkite/commands/checkout-release-branch.sh
+++ b/.buildkite/commands/checkout-release-branch.sh
@@ -32,7 +32,7 @@ BRANCH_NAME="release/${RELEASE_VERSION}"
 git fetch origin "$BRANCH_NAME"
 git checkout "$BRANCH_NAME"
 # Buildkite can reuse a working copy where "$BRANCH_NAME" was left at a different commit by a previous job,
-# so force the local branch to the fetched commit. `reset --hard` rather than
-# `git pull`, to avoid merging if the two diverged; `FETCH_HEAD` rather than
-# `origin/$BRANCH_NAME`, which `git fetch <branch>` only updates opportunistically.
+# so force the local branch to the fetched commit.
+#  - `reset --hard` rather than `git pull`, to avoid merging if the two diverged
+#  - `FETCH_HEAD` rather than `origin/$BRANCH_NAME`, which `git fetch <branch>` only updates opportunistically.
 git reset --hard FETCH_HEAD
--- a/.buildkite/commands/checkout-release-branch.sh
+++ b/.buildkite/commands/checkout-release-branch.sh
@@ -4,8 +4,8 @@
 #
 # Usage: checkout-release-branch.sh [<RELEASE_VERSION>]
 #
-# The release version is taken from, in order of precedence:
-#  1. the first argument, if one is passed
+# The release version is taken from the first of these that is set and non-empty:
+#  1. the first argument
 #  2. the `RELEASE_VERSION` environment variable
 #  3. the `release/*` branch the build is running on, via `BUILDKITE_BRANCH`
 #
@@ -17,15 +17,13 @@
 
 echo "--- :git: Checkout Release Branch"
 
-if [[ $# -gt 0 ]]; then
-  # An argument was passed explicitly, so it must not be empty — that would mean the caller
-  # meant to forward a version but it resolved to nothing, which is a pipeline misconfiguration.
-  RELEASE_VERSION="${1:?release version argument was passed but is empty}"
-elif [[ -n "${RELEASE_VERSION:-}" ]]; then
-  : # Already provided through the pipeline environment
-elif [[ "${BUILDKITE_BRANCH:-}" =~ ^release/ ]]; then
+RELEASE_VERSION="${1:-${RELEASE_VERSION:-}}"
+
+if [[ -z "$RELEASE_VERSION" && "${BUILDKITE_BRANCH:-}" =~ ^release/ ]]; then
   RELEASE_VERSION="${BUILDKITE_BRANCH#release/}"
-else
+fi
+
+if [[ -z "$RELEASE_VERSION" ]]; then
   echo "Error: no release version. Pass it as \$1, set RELEASE_VERSION, or run on a release/* branch." >&2
   exit 1
 fi
@@ -33,7 +31,7 @@ fi
 BRANCH_NAME="release/${RELEASE_VERSION}"
 git fetch origin "$BRANCH_NAME"
 git checkout "$BRANCH_NAME"
-# Buildkite can reuse a working copy where "$BRANCH_NAME" was left at an older commit by a previous job,
+# Buildkite can reuse a working copy where "$BRANCH_NAME" was left at a different commit by a previous job,
 # so force the local branch to the fetched commit. `reset --hard` rather than
 # `git pull`, to avoid merging if the two diverged; `FETCH_HEAD` rather than
 # `origin/$BRANCH_NAME`, which `git fetch <branch>` only updates opportunistically.
--- a/.buildkite/commands/checkout-release-branch.sh
+++ b/.buildkite/commands/checkout-release-branch.sh
@@ -1,12 +1,35 @@
 #!/bin/bash -eu
 
-if [[ -z "${RELEASE_VERSION}" ]]; then
-    echo "RELEASE_VERSION is not set."
-    exit 1
+# Checks out the `release/*` branch for a given release version.
+#
+# Usage: checkout-release-branch.sh [<RELEASE_VERSION>]
+#
+# The release version is taken from, in order of precedence:
+#  1. the first argument, if one is passed
+#  2. the `RELEASE_VERSION` environment variable
+#  3. the `release/*` branch the build is running on, via `BUILDKITE_BRANCH`
+#
+# Buildkite, by default, checks out a specific commit, ending up in a detached HEAD state.
+# But some release steps need to be on the `release/*` branch instead, namely:
+#  - when a `release-pipelines/*.yml` needs to `git push` to the `release/*` branch (for version bumps)
+#  - when a job `pipeline upload`'d by such a pipeline builds from that branch, so that the build
+#    includes the commit that the version bump just added.
+
+echo "--- :git: Checkout Release Branch"
+
+if [[ $# -gt 0 ]]; then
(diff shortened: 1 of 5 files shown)


.buildkite/commands/checkout-release-branch.sh
```diff
@@ -10,3 +10,6 @@ fi
 BRANCH_NAME="release/${RELEASE_VERSION}"
 git fetch origin "$BRANCH_NAME"
 git checkout "$BRANCH_NAME"
+# Buildkite can reuse a working copy where "$BRANCH_NAME" was left at an older commit by a previous job,
+# so realign it on the remote. `reset --hard` rather than `git pull`, to avoid merging if the two diverged.
```
mokagio: Same nitpick as on [DayOne-Android#8243](https://github.com/bloom/DayOne-Android/pull/8243#discussion_r3687267956).

"so realign it on the remote" confused me. "on the remote" made me first think of an operation done _on_ the remote, but the meaning I guess is more like "_based on_ the remote."

Suggestion:

```suggestion
# so force the local branch to the fetched commit. `reset --hard` rather than
# `git pull`, to avoid merging if the two diverged.
```

AliSoftware: Good point, agreed — "on the remote" did read ambiguously. Applied in a7f2b3f5.

For consistency I applied the same wording to all 13 repos (including the two Tumblr PRs on `github.tumblr.net`), so the snippet stays byte-identical everywhere.

AliSoftware:
@mokagio heads-up — these PRs grew since your review, so they're worth another look rather than assuming they're still the three-line change you saw. Same comment on all 13 repos.

Two commits were added on top of the wording fix you suggested:

**1. Reset to `FETCH_HEAD` rather than `origin/$BRANCH_NAME`.** `git fetch origin <branch>` always writes `FETCH_HEAD`, but it only updates `refs/remotes/origin/<branch>` when the remote's configured fetch refspec covers that branch. With Buildkite's default `+refs/heads/*:refs/remotes/origin/*` the two are equivalent — but on a clone whose refspec was narrowed after `origin/<branch>` already existed, the fetch leaves that ref stale and the reset lands on the old commit: the very failure this script exists to prevent, reintroduced through the back door. Resetting to `FETCH_HEAD` drops the refspec dependency entirely.

**2. Standardized the script across all 13 repos.** Tangent to the original issue, bundled deliberately. The script had drifted into **five** different shapes — argument required via `${1?…}` or `${1:?…}`, argument plus a `BUILDKITE_BRANCH` fallback, argument with a hand-rolled usage check, the same under a different variable name, and (in simplenote-android) no argument at all, reading `RELEASE_VERSION` from the environment. Having rolled the same one-line fix out thirteen times this week, that divergence is pure friction.

All 13 now share one canonical script. The resolution order is a **superset** of what every repo did before — argument, then `RELEASE_VERSION` from the environment, then the `release/*` branch the build runs on — so **no call site needed changing**. The `BUILDKITE_BRANCH` fallback only fires on a branch matching `^release/`, and it derives the branch name back from that same value, so it cannot select a branch other than the one the build was already triggered on.

It also closes two latent bugs: under `bash -eu`, both `[[ -z "${RELEASE_VERSION}" ]]` on an unset variable and a bare `RELEASE_VERSION=$1` with no arguments abort with `unbound variable` before their intended usage message can print. And an argument that is *passed but empty* — what happens when a pipeline forwards an unset `$RELEASE_VERSION` — is now a hard error everywhere, instead of resolving to `release/` or silently falling through to the current branch.

The scripts are byte-identical across all 13 repos and pass `shellcheck` everywhere; the argument resolution was exercised across every combination (argument / empty argument / environment variable / `release/*` / trunk / feature branch / unset). Full rationale in the updated PR description.

Worth flagging that the [release-toolkit PR](https://github.com/wordpress-mobile/release-toolkit/pull/763) also picked up a real fix from your review — the tag-based lookup was preferring a leftover draft over the release that actually owns the tag.

AliSoftware:
[wordpress-mobile/release-toolkit#763](https://github.com/wordpress-mobile/release-toolkit/pull/763) has shipped in **14.11.2**, so the do-not-merge note is gone from the description — but unlike the other repos in this rollout, **this PR does not bump the gem**.

This repo is pinned to `~> 13.8`, so reaching 14.11.2 means crossing a **major version**. The 13.x → 14.0.0 migration has three breaking changes:

- `ios_update_metadata_source` and `an_update_metadata_source` are replaced by `gp_update_metadata_source`
- translator comments (`#.` lines) are dropped from regenerated PO files unless migrated into the `source_files` hash
- PO entry ordering changed

This repo does call those actions, so bumping it here would break the localization lanes and would need per-lane verification of the regenerated PO files. That does not belong in a PR about a CI checkout script.

So this PR ships the `checkout-release-branch.sh` fix on its own — it stands alone and needs no new gem version. **The `release-toolkit` upgrade and its migration will follow in a dedicated PR.**

For reference, the other eleven repos in this rollout are now on 14.11.2: nine were already on the 14.11.x line and took a patch bump, and the two Tumblr repos crossed the major cleanly because they call none of the affected actions.

AliSoftware:
📌 The deferred `release-toolkit` upgrade now has its own PR: **Automattic/simplenote-android#1858**, stacked directly on this branch.

It does the 13.8 → 14.11.2 bump along with the migration this repo needed to make that possible — `an_update_metadata_source` → `gp_update_metadata_source`, with all four translator comments moved into `source_files` as `{ path:, comment: }` so nothing is lost from GlotPress. It also regenerates `PlayStoreStrings.pot` so the effects are reviewable rather than surfacing mid-release, and drops the nokogiri pin whose own comment said to remove it once the toolkit reached >= 14.4.1.

**Merge order: this PR first**, then #1858 — GitHub will retarget it to `trunk` automatically once this one lands.

iangmaia (approved):

.buildkite/release-pipelines/complete-code-freeze.yml
```diff
@@ -14,7 +14,6 @@ steps:
       echo '--- :robot_face: Use bot for Git operations'
       source use-bot-for-git
 
-      echo '--- :git: Checkout release branch'
```
iangmaia, line 17: Oh nice touch to remove these log entries and have them in the script itself. I guess this was a leftover from our very first iterations on the release via CI setup.

PR #1858, opened 2026-08-03
Upgrade release-toolkit to 14.11.2 and migrate to gp_update_metadata_source

AliSoftware:
> [!NOTE]
> **Stacked on top of the [AINFRA-2725 checkout](https://github.com/Automattic/simplenote-android/pull/1856) fix.** This PR targets `ainfra-2725/reset-release-branch-checkout`, not the default branch, so the diff shows only the toolkit upgrade. Merge that PR first; this one will then retarget automatically.

## What does it do?

Completes AINFRA-2725 for this repo. The rollout bumped `release-toolkit` to 14.11.2 everywhere it was a patch bump, but had to leave this repo on `~> 13.8`: reaching 14.x crosses a major version whose breaking changes land squarely on the metadata source generation this repo uses.

### 1. Migrate to `gp_update_metadata_source`

`an_update_metadata_source` is deprecated in favour of `gp_update_metadata_source`. The important difference is that the new action **regenerates the `.pot` from scratch** rather than editing it in place, so translator comments are dropped unless declared alongside the source paths.

Every translator comment currently in `PlayStoreStrings.pot` is therefore carried into `source_files` using the `{ path:, comment: }` form. Nothing is lost from GlotPress.

### 2. Regenerate `PlayStoreStrings.pot`

Rather than leaving the change to surface on the next release, the lane was run here so the effects are reviewable. Three things change, all inherent to regenerating from scratch:

- **Entries are sorted alphabetically** by `msgctxt`, so the file is reordered.
- **The previous version's `release_note_N` entry is dropped.** The generator emits a release-note entry only for the version it is given. ⚠️ *This one is worth a localization reviewer's eye* — it isn't in the toolkit's `MIGRATION.md`, but WordPress-Android's `.po` shows the same single-release-note shape after its own migration on 14.7.0, so it's the accepted behaviour rather than a regression here.
- **The header** loses its two leading `#` comment lines and `Project-Id-Version`, and `X-Generator` now names the toolkit. This matches WordPress-Android's post-migration header field for field.

## Testing instructions

The `.pot` in this PR is the real output of the migrated lane, so the diff *is* the test. Worth verifying as a reviewer:

1. Every `#. translators:` comment that existed before still exists after — the count is unchanged.
2. The only entry that disappeared is the previous version's `release_note_N`.
3. `msgid` values are unchanged; only ordering, that one entry, and the header differ.


Simplenote/metadata/PlayStoreStrings.pot
```diff
@@ -1,37 +1,25 @@
-# Translation of Release Notes & Play Store Descriptions in English (US)
-# This file is distributed under the same license as the Release Notes & Play Store Descriptions package.
 msgid ""
 msgstr ""
-"PO-Revision-Date: 2019-06-19 11:49:24+0000\n"
+"PO-Revision-Date: 2026-08-03 20:00+0200\n"
```
iangmaia, line 3: 👍

iangmaia (approved):
👍

AliSoftware:
Connecting the dots: this PR happened to also close https://linear.app/a8c/issue/AINFRA-2787/update-release-toolkit-to-14x-simplenote-android

PR #1873, opened 2026-09-21
Bump targetSdk to 36 (Android 16)

Twinsen81:
Closes SIMPL-80. Google Play release blocker.

### Fix

Bumps `targetSdkVersion` to 36. Two Android 16 behaviour changes have no opt-out and needed code:

**Back navigation** — `onBackPressed()` is no longer called. Affected screens moved to `OnBackPressedDispatcher` callbacks. `PasscodeUnlockActivity` is a plain `Activity`, so it registers a platform `OnBackInvokedCallback` instead.

**Edge-to-edge** — now forced. Window insets applied to the passcode layouts.

Two bugs found while testing, fixed here:

- Passcode screens crashed on tablets: the `sw600dp` layout was missing the root id the inset code looks up.
- In landscape, a back gesture left the editor toolbar and Markdown tabs hidden. The gesture dismisses the keyboard without a key event, so `onKeyPreIme()` never cleared focus. Driven by the IME inset now.

### Test

Galaxy S24+ (16), Pixel 6 Pro (17 beta), Galaxy Tab A9 (16), Pixel Tablet emulator (35), phone emulators (36, 31, 28).

1. Drawer open → back closes it, app stays. Second back exits.
2. Editor → back returns to the list. With the keyboard up, back hides it and keeps you in the editor — in landscape the toolbar and tabs must come back.
3. Passcode lock → back locks and goes home, still locked on relaunch. Never unlocks.
4. Passcode settings on a tablet → opens without crashing.
5. Signup/login → back returns to the welcome screen.

### Review

`PasscodeUnlockActivity` is the security-relevant one: back must lock, never dismiss.

`onBackPressed()` is kept deliberately: the platform only consults the `OnBackInvokedCallback` where the dispatcher is enabled for the app, so `onBackPressed()` is still the live lock path elsewhere. Tested on API 28 and 31 for that reason — please don't drop it as redundant for API 33+.

### Release

No release notes needed — no intended user-facing change.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR is not assigned to a milestone.</td>
    </tr>
  </tbody>
</table>

</p>

obsantos:
Two things worth addressing before merge; everything else looks good.

### 1. Content focus is cleared on every keyboard dismissal, not just back

`NoteEditorFragment.releaseContentFocusWhenKeyboardHides()` clears focus whenever the IME inset goes away for any reason. That is much broader than the `KEYCODE_BACK` path in `SimplenoteEditText.onKeyPreIme()` it is compensating for.

Scenario: editing with the keyboard up, then open the Info/History bottom sheet, tap a link, or switch apps. The IME hides, the listener fires, and the content field loses focus. When the sheet is dismissed or the user returns, the field is unfocused, the keyboard does not come back, and the caret/selection handles are gone. Before this PR focus survived those interruptions.

Suggestion: scope the clear to the case that actually needs it (landscape editing inside `NoteEditorActivity`, which is the only place `hideToolbarForLandscapeEditing()` acts), or gate it on the window still having focus. Related: `mWasKeyboardVisible` is a fragment field but the listener is installed per view in `onCreateView()`, so a stale `true` survives rotation and the first inset pass on the new view clears the restored focus. Resetting it before installing the listener avoids that.

### 2. The `PasscodeUnlockActivity` comment misstates which devices still use `onBackPressed()`

The comment says `onBackPressed()` only covers devices older than Android 13. In practice `enableOnBackInvokedCallback` defaults to `false` on Android 13, 14 and 15 regardless of `targetSdk` (the default-true rule lives in the Android 16 package parser), so the platform `OnBackInvokedCallback` is ignored there and `onBackPressed()` is the live lock path on every device below Android 16. The PR description's "predictive back is only the default at API 35+" has the same off-by-one.

The code is correct today. The risk is a future cleanup trusting the comment and removing the override as redundant for API 33+, which would let back dismiss the lock screen without locking on Android 13-15. Rewording the comment to "covers every device below Android 16 unless the manifest opts in explicitly" closes that door.

obsantos:
Follow-up with the remaining, lower-priority findings from the same review.

### 3. `mWasKeyboardVisible` is not reset when the fragment view is recreated

`NoteEditorFragment.java` — the flag lives on the fragment, but the inset listener is installed per view in `onCreateView()`. A stale `true` survives rotation: the new view's first inset pass sees the IME hidden, treats it as a dismissal, and clears the focus that `PhoneWindow` just restored. Resetting the flag to `false` right before installing the listener avoids that.

### 4. Back-button `AppLog` entry dropped for the non-task-root path

`NoteEditorActivity.java` — `handleBackPressed()` logged "Tapped back button in navigation bar" on every back. The replacement callback only runs when the editor is the task root, so the common case (editor opened from the list) no longer logs the action. Either log from the toolbar click listener too, or accept the narrower logging deliberately.

### 5. Drawer-close callback loses its former top priority

`NotesActivity.java` — the old `onBackPressed()` override ran before the dispatcher, so an open drawer always won. `mCloseDrawerOnBack` is now registered in `configureNavigationDrawer()` before the fragments are added, and any callback registered later (including child `FragmentManager` back-stack callbacks of those fragments) takes precedence over it. With the drawer open, such a callback would handle back first while the drawer stays open. Registering the drawer callback after the fragments are attached, or in `onPostCreate()` alongside the state sync, restores the old ordering.

### 6. System bar icon colour is hardcoded to light regardless of background

`AbstractPasscodeKeyboardActivity.setUpEdgeToEdge()` forces light (white) status and navigation bar icons on every API level. The library's own default `passcodelock_background` is `#E5E5E5`; it only works here because the Simplenote module overrides that colour to blue. Any consumer without that override, or a future colour change, gets invisible icons on the lock screen. Deriving the appearance from the resolved background luminance (`ColorUtils.calculateLuminance`) or exposing it as a resource would make it self-consistent.

### 7. Insets ignore `displayCutout` on rotatable tablets

`AbstractPasscodeKeyboardActivity.setUpEdgeToEdge()` pads for `Type.systemBars()` only. Rotation is allowed on `sw600dp`, so on a tablet or foldable with a side cutout in landscape the prompt or keypad edge can sit under the cutout. `Type.systemBars() | Type.displayCutout()` covers it.

### 8. Public `onBackPressed()` on `FullScreenDialogFragment` has no callers

With back now routed through the `ComponentDialog` dispatcher, the fragment's own public `onBackPressed()` (around line 311) is dead code with no call sites in the repo. Removing it, or routing the dispatcher callback through it, leaves a single entry point.

Twinsen81:
Thanks — all eight looked at, seven addressed in 273e84f9. Verified each rather than taking them on trust; notes where that changed the answer.

**1. Focus cleared on every keyboard dismissal** — confirmed, and it was a real regression. A/B'd on an API 36 emulator: without the listener, HOME→return restored keyboard and focus; with it, both were lost.

Rather than scope the clear, I stopped touching focus at all. `hideToolbarForLandscapeEditing()` now tests `mIsKeyboardVisible` alongside `hasFocus()`, and the inset listener just calls it on a visibility transition. Keyboard visibility is what the chrome actually tracks, so the whole class of focus side effects goes away: after a back gesture the toolbar returns *and* the caret stays.

**2. `PasscodeUnlockActivity` comment** — reworded, and the same claim removed from the PR description.

I could not verify the exact version boundary: API 35 emits no `CoreBackPreview` lines at all, so absence of the tag proves nothing either way. So the comment no longer names a version — it states the dependency and adds an explicit "don't remove this as redundant" note, which closes the risk wherever the boundary sits.

**3. Stale `mWasKeyboardVisible`** — fixed, reset before the listener is installed. For the record the rotation case specifically can't occur: `NoteEditorActivity` declares `configChanges="orientation|screenSize|…"` so it isn't recreated on rotation. Correct for other view-recreation paths.

**4. Dropped `AppLog` entry** — restored on the toolbar path. The task-root callback now logs its own distinct action rather than repeating the same string.

**5. Drawer callback priority** — moved to `onPostCreate()`. Technically right about ordering; for the record it looks unreachable in practice, since the only back-stack entries in `NotesActivity`'s FragmentManager come from `FullScreenDialogFragment` and the WordPress publish dialog, both full-screen modals.

**6. Hardcoded icon colour** — now derived from `ColorUtils.calculateLuminance` of the resolved background. No visual change today (blue is below the threshold, so still white icons), but a consumer without the override no longer gets invisible icons.

**7. `displayCutout`** — added, insets are now `systemBars() | displayCutout()`.

**8. Dead `onBackPressed()`** — routed the dispatcher *through* it rather than deleting it: it carries an `isAdded()` guard the callback lacked, so this leaves one entry point and keeps the guard.

**Testing** — API 35 tablet emulator, API 36 phone emulator, API 28 phone emulator. Drawer back (closes drawer, app stays, second back exits) passes on all three. Editor: toolbar hides on focus and returns on back, gesture back keeps focus, HOME→return keeps focus and keyboard. Passcode screen renders correctly on tablet. No crashes on any device. `assembleDebug`, `lintDebug`, `testDebugUnitTest` green, 111 tests / 0 failures.

obsantos (approved):
Thanks for the bump and changes! LGTM 👍

